PR #50, opened 2020-10-15
Default cache doesn't handle family id correctly

### LINKED ISSUE #49: default_storage_manager_test is flaky
**Where is the issue?**
Tests

Test default_storage_manager_test.TestReadRefreshToken failed on go 1.14, passed on build previously

See if test is flaky and needs refactor

Passing test: https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/45/checks?check_run_id=1254682076
Failing test: https://github.com/AzureAD/microsoft-authentication-library-for-go/runs/1255127547?check_suite_focus=true

keegan-caruso:
Fixes #49 

Problem was both in the tests and the implementation.

This brings behavior more in line with MSAL4J's implementation: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/0b20b1430ab57cd1f331935b79aa8542c5be9c9c/src/main/java/com/microsoft/aad/msal4j/TokenCache.java#L543

Could use an ack from someone on an MSAL team on cache key behavior.

Might be worth revisiting some of this API

### DIFF
--- a/src/internal/tokencache/default_storage_manager.go
+++ b/src/internal/tokencache/default_storage_manager.go
@@ -96,23 +96,24 @@ func (m *defaultStorageManager) ReadRefreshToken(
 	defer lock.RUnlock()
 	if familyID != "" {
 		for _, rt := range m.refreshTokens {
-			if testClientIDRefreshToken(rt, homeAccountID, envAliases, clientID) {
+			if matchFamilyRefreshToken(rt, homeAccountID, envAliases) {
 				return rt
 			}
 		}
 		for _, rt := range m.refreshTokens {
-			if testFamilyRefreshToken(rt, homeAccountID, envAliases) {
+			if matchClientIDRefreshToken(rt, homeAccountID, envAliases, clientID) {
 				return rt
 			}
 		}
 	} else {
 		for _, rt := range m.refreshTokens {
-			if testFamilyRefreshToken(rt, homeAccountID, envAliases) {
+			if matchClientIDRefreshToken(rt, homeAccountID, envAliases, clientID) {
 				return rt
 			}
 		}
+
 		for _, rt := range m.refreshTokens {
-			if testClientIDRefreshToken(rt, homeAccountID, envAliases, clientID) {
+			if matchFamilyRefreshToken(rt, homeAccountID, envAliases) {
 				return rt
 			}
 		}
@@ -120,13 +121,13 @@ func (m *defaultStorageManager) ReadRefreshToken(
 	return nil
 }
 
-func testFamilyRefreshToken(rt *refreshTokenCacheItem, homeAccountID string, envAliases []string) bool {
+func matchFamilyRefreshToken(rt *refreshTokenCacheItem, homeAccountID string, envAliases []string) bool {
 	return msalbase.GetStringFromPointer(rt.HomeAccountID) == homeAccountID &&
 		checkAlias(msalbase.GetStringFromPointer(rt.Environment), envAliases) &&
 		msalbase.GetStringFromPointer(rt.FamilyID) != ""
 }
 
-func testClientIDRefreshToken(rt *refreshTokenCacheItem, homeAccountID string, envAliases []string, clientID string) bool {
+func matchClientIDRefreshToken(rt *refreshTokenCacheItem, homeAccountID string, envAliases []string, clientID string) bool {
 	return msalbase.GetStringFromPointer(rt.HomeAccountID) == homeAccountID &&
 		checkAlias(msalbase.GetStringFromPointer(rt.Environment), envAliases) &&
 		msalbase.GetStringFromPointer(rt.ClientID) == clientID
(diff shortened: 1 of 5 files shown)


jennyf19:
@bgavrilMS can you take a look?

src/internal/tokencache/default_storage_manager.go
```diff
@@ -96,37 +96,38 @@ func (m *defaultStorageManager) ReadRefreshToken(
 	defer lock.RUnlock()
 	if familyID != "" {
```
bgavrilMS, line 97: Interesting way of handling FOCI. I think it works. 

In .NET the algorithm we went with is: 

Simple algo: 

1. Try to find an FRT (i.e. ignore clientID filter)
2. If found, try to redeem the FRT
3. If AAD returns a token, hurray! Client is part of the family and we got a token.
3.1. If however AAD returns error code "client_mismatch" ( I think you need to ask AAD to add MSAL.GO on a special list for this to be returned) or "invalid_grant", then you know the client ID making the request is not part of the family

4. Assuming that FRT was either not found or it could not be redeemed, try to find the normal RT (filter by clientID)

bgavrilMS: More complex algo involves remembering in the token cache which ClientIDs are part of the family. MSAL doesn't know this from the start, so assume a client is part of the family if you don't know otherwise.

src/internal/tokencache/default_storage_manager.go
```diff
@@ -96,37 +96,38 @@ func (m *defaultStorageManager) ReadRefreshToken(
...
 		for _, rt := range m.refreshTokens {
-			if testClientIDRefreshToken(rt, homeAccountID, envAliases, clientID) {
+			if matchFamilyRefreshToken(rt, homeAccountID, envAliases) {
```
bgavrilMS, line 99: yes, fair enough, you should not filter by clientID when looking for an FRT.

bgavrilMS:
I assume the logic to call `ReadRefreshToken(familyID="xxx")` is based on AppMetadata ?

bgavrilMS:
By the way, we have 2 apps in the lab that are part of the family, could be useful for writing an integration test, see: 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/353be255d40c36bea67e2f816c4fca5ed902471a/tests/Microsoft.Identity.Test.Common/TestConstants.cs#L166

and here are the .NET integration tests: 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/353be255d40c36bea67e2f816c4fca5ed902471a/tests/Microsoft.Identity.Test.Integration.net45/SeleniumTests/FociTests.cs#L46

bgavrilMS (approved):
LGTM assuming end to end tests work (see comments for details)

keegan-caruso:
> I assume the logic to call `ReadRefreshToken(familyID="xxx")` is based on AppMetadata ?

Yes, that is here: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/src/internal/tokencache/default_cache_manager.go#L82

src/internal/tokencache/default_storage_manager_test.go
```diff
@@ -311,74 +312,203 @@ func TestWriteIDToken(t *testing.T) {
...
+				familyID:      "",
+				clientID:      "cid2",
+			},
+			want: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := tt.m.ReadRefreshToken(tt.args.homeAccountID, tt.args.envAliases, tt.args.familyID, tt.args.clientID); !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("defaultStorageManager.ReadRefreshToken() = %v, want %v", render.AsCode(got), render.AsCode(tt.want))
```
keegan-caruso, line 509: bringing in render as a test dependency so we get meaningful test failures. with just using %v on a struct with a pointer, you only get the pointer value (an address)

src/internal/tokencache/default_storage_manager.go
keegan-caruso: Is it worth writing an issue for?

bgavrilMS: Is there an issue? I see msal.go already does the app metadata trick. I was trying to give some context around FOCI, not suggesting there is a bug.

keegan-caruso: The implementation seems a little different, looks like this one doesn't handle client_mismatch / invalid_grant and I don't think failover from FRT to RT

bgavrilMS: That's probably fine. I do wonder what happens if an app which is part of the family is removed from the family? Will MSAL go into an infinite loop or will the user just have to re-login interactively ? (second variant is acceptable imo )

henrik-me: leaving the family is supported, as far as I recall

---
In reply to: [506577218](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/50#discussion_r506577218) [](ancestors = 506577218)

bgavrilMS: If you leave the family, then on refreshing the FRT you get "client_mismatch" or "invalid_grant" from EVO, which msal MUST interpret as a `UIRequiredException`.

After interactive login, the app metadata will reflect the new state of world, where the app is no longer in the family. 

Should be fairly easy to unit test.

PR #51, opened 2020-10-20
Add go fmt ci check

keegan-caruso:


.github/workflows/go.yml
```diff
@@ -29,6 +29,10 @@ jobs:
       - name: Get dependencies
         run: go get -v -t -d ./...
 
+        # designed to only run on linux
+      - name: Format Check
+        run: if [ $(gofmt -l -s . | wc -l) -ne 0 ]; then echo "fmt failed"; exit 1; fi
```
keegan-caruso, line 34: go fmt
  -l print file name
  -s include simplifications

wc
  -l only print out number of lines

Fail if number of entries is not 0

jennyf19 (approved):

jennyf19:
awesome @keegan-caruso ! :shipit:

Issue #62, opened 2020-10-27
Reduce use of unneeded pointer types within internal/


element-of-surprise:
There are types within internal/ that use pointers to basic types, such as:

```go
type Account struct {
	HomeAccountID     *string `json:"home_account_id,omitempty"`
	Environment       *string `json:"environment,omitempty"`
	Realm             *string `json:"realm,omitempty"`
	LocalAccountID    *string `json:"local_account_id,omitempty"`
	AuthorityType     *string `json:"authority_type,omitempty"`
	PreferredUsername *string `json:"username,omitempty"`
	GivenName         *string `json:"given_name,omitempty"`
	FamilyName        *string `json:"family_name,omitempty"`
	MiddleName        *string `json:"middle_name,omitempty"`
	Name              *string `json:"name,omitempty"`
	AlternativeID     *string `json:"alternative_account_id,omitempty"`
	RawClientInfo     *string `json:"client_info,omitempty"`
	additionalFields  map[string]interface{}
}
```
This shouldn't be necessary for json encode/decode.  This creates a lot of objects on the heap that cause GC slowdown.  

Not that this is never done, but usually only if you need to tell the difference between a zero value and set value.  Protocol Buffer version 2 does this.  However, that method has been deprecated in version 3 because of these types of issues.

This will take a non-insignificant amount of work to do and if it is agreed this should be worked on, we should avoid colliding with another PR in internal/.

PR #65, opened 2020-10-27
Fix comments and minor style issues in msal pkg

keegan-caruso:


jennyf19:
idk if the work on [this](https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/62) has started, so let's coordinate before we merge things. fyi: @element-of-surprise

PR #67, opened 2020-10-28
Remove unneeded pointers

element-of-surprise:
There wasn't a reason I could locate for using pointers to basic types within the structs.  I've removed them from packages in internal/

In addition, as this required touching some test code, I started some really basic refactoring there.

Many calls signatures I shortened using the multiple delcaration syntax, aka... "varname, varName1 string" vs "varName string varName1 string"

Got rid of most 80 character line limits for things I was calling.  These will get shorter in future changes.

Using pretty.Compare as it is superior to reflect.DeepEqual in every way.
Also DeepEqual was being used where basic type equality could be used, scratched most of that.

There is a side effect to this change, in that additionalFields actually holds all fields.  I'm not sure what caused this (I didn't investigate), but it doesn't matter.  The results for output are the same, just uses slightly more space.  This will be fixed when I change how we encode/decode json.


henrik-me:
no concerns as long as the additional fields are preserved.

And I like the code without the pointers much more!

henrik-me (approved):
:shipit:

henrik-me:
> go 1.14

nit, thought we had moved to 0.15

---
Refers to: go.mod:3 in 3aaeb26. [](commit_id = 3aaeb268a412e61eef944b94f963d2cb68727364, deletion_comment = False)

henrik-me:
there are lots of dependencies in this project, we should ensure to keep those to a minimum and we need to ensure those have the right licenses, security review etc. so we can actually use them

internal/msalbase/account_test.go
```diff
@@ -8,17 +8,19 @@ import (
...
 
-var accHID = "hid"
-var accEnv = "env"
-var accRealm = "realm"
-var authType = "MSSTS"
-var accLid = "lid"
-var accUser = "user"
+var (
+	accHID   = "hid"
+	accEnv   = "env"
```
jennyf19, line 13: nit: I would just write out `accHomeAccountId = 'homeAccountId"

internal/msalbase/account_test.go
```diff
@@ -8,17 +8,19 @@ import (
...
+	accRealm = "realm"
+	authType = "MSSTS"
+	accLid   = "lid"
```
jennyf19, line 16: >accLid [](start = 1, length = 6)

nit: is this local account id? same comment as for homeAccountId, to make it easier to read.

internal/msalbase/account_test.go
```diff
@@ -8,17 +8,19 @@ import (
...
-var accEnv = "env"
-var accRealm = "realm"
-var authType = "MSSTS"
-var accLid = "lid"
-var accUser = "user"
+var (
+	accHID   = "hid"
+	accEnv   = "env"
+	accRealm = "realm"
+	authType = "MSSTS"
```
jennyf19, line 15: >MSSTS [](start = 13, length = 5)

side note: @henrikm we should discuss what MSSTS means...should i open an issue?

internal/tokencache/access_token_cache_item.go
```diff
@@ -12,82 +12,65 @@ import (
...
-		Environment:                    &environment,
-		Realm:                          &realm,
-		CredentialType:                 &credentialType,
-		ClientID:                       &clientID,
-		Secret:                         &accessToken,
-		Scopes:                         &scopes,
-		CachedAt:                       &cachedAtString,
-		ExpiresOnUnixTimestamp:         &expiresOnString,
-		ExtendedExpiresOnUnixTimestamp: &extExpiresOnString,
+func createAccessTokenCacheItem(homeID, env, realm, clientID string, cachedAt, expiresOn, extendedExpiresOn int64, scopes, token string) *accessTokenCacheItem {
```
jennyf19, line 28: >string [](start = 61, length = 6)

do you need string here w/clientID?

internal/tokencache/access_token_cache_item_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"reflect"
 	"testing"
 	"time"
+
+	"github.com/kylelemons/godebug/pretty"
```
jennyf19, line 11: question - in go, do you usually leave a space before adding the external imports, or should the line above be removed?

internal/tokencache/access_token_cache_item_test.go
```diff
@@ -21,16 +23,16 @@ var (
...
-		Environment:                    &env,
-		CredentialType:                 &credential,
-		ClientID:                       &clientID,
-		Realm:                          &realm,
-		Scopes:                         &scopes,
-		Secret:                         &secret,
-		ExpiresOnUnixTimestamp:         &expiresOn,
-		ExtendedExpiresOnUnixTimestamp: &extExpiresOn,
-		CachedAt:                       &cachedAt,
+		HomeAccountID:                  testHID,
```
jennyf19, line 26: >HID [](start = 38, length = 3)

nit: same as earlier, i would write out homeAccountId, so it's easier to read quickly

internal/tokencache/access_token_cache_item_test.go
```diff
@@ -69,33 +71,34 @@ func TestAccessTokenPopulateFromJSONMap(t *testing.T) {
...
+		t.Errorf("TestAccessTokenPopulateFromJSONMap(access tokens): -want/+got:\n %s", diff)
 	}
-	actualHID := *actualAccessToken.HomeAccountID
-	if !reflect.DeepEqual(actualHID, testHID) {
-		t.Errorf("Expected home account ID %s differs from actual home account ID %s", actualHID, testHID)
+
+	gotExtra := got.additionalFields["extra"].(string)
+	if gotExtra != "this_is_extra" {
+		t.Errorf("TestAccessTokenPopulateFromJSONMap(extra field): got %s, want %s", gotExtra, "this_is_extra")
 	}
```
jennyf19, line 92: is this standard go unit test structure? i assume it's in the go docs...haven't seen the want/got before. pretty cool

internal/tokencache/app_metadata_cache_item.go
```diff
@@ -11,33 +11,31 @@ import (
...
+	Environment      string `json:"environment,omitempty"`
 	additionalFields map[string]interface{}
 }
 
-func createAppMetadata(familyID string, clientID string, environment string) *appMetadata {
-	metadata := &appMetadata{
-		FamilyID:    &familyID,
-		ClientID:    &clientID,
-		Environment: &environment,
+func createAppMetadata(familyID, clientID, environment string) *appMetadata {
```
jennyf19, line 20: >string [](start = 55, length = 6)

this is cool. we don't do this in C#.

internal/tokencache/default_storage_manager.go
```diff
@@ -4,12 +4,17 @@
...
 	"sync"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/internal/msalbase"
 )
 
+// TODO(jdoak): Investigate this lock. It is strange to have a global lock.
+// TODO(jdoak): Find out what our expected number of concurrent reads are and the expected number of tokens.
+// RWMutex is only more performant in high reads with low writes. Also, if the number of tokens is low,
+// we might be able to use atomic.Value with map copies to achieve lockless read/writes. If the token number
+// is really low, we could replace this with slices.
```
jennyf19, line 17: @jmprieur can probably help with this.

jennyf19 (approved):
:shipit:

jennyf19:
Agree w/@henrikm on the security reviews and 3rd party libs

---
In reply to: [718277691](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/67#issuecomment-718277691) [](ancestors = 718277691)

internal/tokencache/default_storage_manager.go
jennyf19: or @bgavrilMS thoughts?

keegan-caruso:
> > go 1.14
> 
> nit, thought we had moved to 1.15
> 
> Refers to: go.mod:3 in [3aaeb26](https://github.com/AzureAD/microsoft-authentication-library-for-go/commit/3aaeb268a412e61eef944b94f963d2cb68727364). [](commit_id = [3aaeb26](https://github.com/AzureAD/microsoft-authentication-library-for-go/commit/3aaeb268a412e61eef944b94f963d2cb68727364), deletion_comment = False)

Maybe you are thinking of our test matrix? Which is [1.13, 1.14, 1.15]

internal/tokencache/access_token_cache_item_test.go
```diff
@@ -48,8 +50,8 @@ func TestCreateAccessTokenCacheItem(t *testing.T) {
 		"user.read",
 		"access",
 	)
-	if !reflect.DeepEqual(extExpiresOn, *actualAt.ExtendedExpiresOnUnixTimestamp) {
-		t.Errorf("Actual ext expires on %s differs from expected ext expires on %s", *actualAt.ExtendedExpiresOnUnixTimestamp, extExpiresOn)
+	if !reflect.DeepEqual(extExpiresOn, actualAt.ExtendedExpiresOnUnixTimestamp) {
```
abhidnya13, line 53: I see on line 106, we moved from using `reflect.DeepEqual` to a simple comparison. 
Can we do this here as well ?

internal/tokencache/access_token_cache_item_test.go
abhidnya13: +1 to this, very cool

keegan-caruso (approved):

abhidnya13 (approved):
🚢

internal/msalbase/account_test.go
henrik-me: Yeah, it's just a test value. Has no meaning really AFAIK. MS sts...

internal/msalbase/account_test.go
element-of-surprise: I would table this for the future.  These names are unchanged from the current dev code.  The tests need a lot of help and it is very likely that none of these will survive at the end.  acc prefix here  isn't useful and in the end, we probably are going to change how this is done.

internal/msalbase/account_test.go
element-of-surprise: Same as above, this is unchanged from the original code and probably wont' survive long term.

internal/tokencache/access_token_cache_item_test.go
element-of-surprise: So there are several different ways that imports are done.  You can bundle everything together, no line breaks.  

My method is to divide in up to 5 sections divided by a line break:

- stdlib
- depot imports
- third party imports
- protocol buffer imports
- side effect imports

Rarely do you get above 3 outside of package main

internal/tokencache/access_token_cache_item_test.go
element-of-surprise: This is closer to standard go testing structure.  The got, want thing is a google internal standard, where got always proceeds want EXCEPT when it its a diff, which is -want/+got.  

Table driven tests are the standard in Go, though a lot of this is single value tests so they might not fit into it as well.  
Cheney has a good write up: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

internal/tokencache/access_token_cache_item_test.go
element-of-surprise: Removed use of reflect completely from this file.

element-of-surprise:
Talked to Jenny oob, going to table the MSST value change until some team discussion and the variable renames can wait a bit.

Issue #56, opened 2020-10-23
[Bug] APIs that perform I/O must take a `context.Context` as their first parameter


jhendrixMSFT:
E.g. `ConfidentialClientApplication.AcquireTokenByClientCredential()` should have a first param `ctx context.Context`.

This indicates to the caller that the API performs an I/O request and allows the caller to control the lifetime of the request.

devigned:
Passing in a context a `ctx` is a must have. My team would hesitate to take a dependency on a rpc library that didn't allow for request cancellation.

jhendrixMSFT:
This has been fixed in the `dev` branch.

jhendrixMSFT:
Found a few more APIs that need a context.

jhendrixMSFT:
Should be fully fixed now.

Issue #60, opened 2020-10-27
[Feature Request] Propose new API for Device Code flow + public client application


jennyf19:

jhendrixMSFT:
Fixed in the `dev` branch.

PR #68, opened 2020-10-29
Refactored public surface area for device code flow

jhendrixMSFT:
Renamed PublicClientApplication constructor function and updated its sig
to take an options struct.
Updated AcquireTokenSilent() and AcquireTokenByDeviceCode() APIs to take
a context.Context and options structs.
Removed "Get" prefix from getter-type methods.
Replaced some interface return types with their underlying type.

### DIFF
--- a/msal/confidential_client_application.go
+++ b/msal/confidential_client_application.go
@@ -19,13 +19,13 @@ type ConfidentialClientApplicationOptions struct {
 
 	// Client sets the transport for making HTTP requests.
 	// Leave this as nil to use the default HTTP transport.
-	Client HTTPClient
+	HTTPClient HTTPClient
 }
 
 // DefaultConfidentialClientApplicationOptions returns an instance of ConfidentialClientApplicationOptions initialized with default values.
 func DefaultConfidentialClientApplicationOptions() ConfidentialClientApplicationOptions {
 	return ConfidentialClientApplicationOptions{
-		Client: http.DefaultClient,
+		HTTPClient: http.DefaultClient,
 	}
 }
 
--- a/msal/confidential_client_application.go
+++ b/msal/confidential_client_application.go
@@ -17,6 +17,9 @@ type ConfidentialClientApplicationOptions struct {
 	// By default there is no cache persistence.
 	Accessor CacheAccessor
 
+	// The host of the Azure Active Directory authority. The default is https://login.microsoftonline.com/common.
+	Authority string
+
 	// Client sets the transport for making HTTP requests.
 	// Leave this as nil to use the default HTTP transport.
 	HTTPClient HTTPClient
@@ -25,6 +28,7 @@ type ConfidentialClientApplicationOptions struct {
 // DefaultConfidentialClientApplicationOptions returns an instance of ConfidentialClientApplicationOptions initialized with default values.
 func DefaultConfidentialClientApplicationOptions() ConfidentialClientApplicationOptions {
 	return ConfidentialClientApplicationOptions{
+		Authority:  authorityPublicCloud,
 		HTTPClient: http.DefaultClient,
 	}
 }
@@ -41,12 +45,16 @@ type ConfidentialClientApplication struct {
 // NewConfidentialClientApplication creates a ConfidentialClientApplication instance given a client ID, authority URL and client credential.
 // Pass nil for options to accept the default values; this is the same as passing the result
 // from a call to DefaultConfidentialClientApplicationOptions().
-func NewConfidentialClientApplication(clientID string, authority string, clientCredential ClientCredentialProvider, options *ConfidentialClientApplicationOptions) (*ConfidentialClientApplication, error) {
+func NewConfidentialClientApplication(clientID string, clientCredential ClientCredentialProvider, options *ConfidentialClientApplicationOptions) (*ConfidentialClientApplication, error) {
 	cred, err := createInternalClientCredential(clientCredential)
 	if err != nil {
 		return nil, err
 	}
-	clientApp := createClientApplication(clientID, authority)
+	if options == nil {
+		def := DefaultConfidentialClientApplicationOptions()
+		options = &def
+	}
+	clientApp := createClientApplication(clientID, options.Authority)
 	return &ConfidentialClientApplication{
 		clientApplication: clientApp,
 		clientCredential:  cred,
--- a/msal/confidential_client_application.go
+++ b/msal/confidential_client_application.go
@@ -4,9 +4,30 @@
 package msal
 
 import (
+	"context"
+	"net/http"
+
 	"github.com/AzureAD/microsoft-authentication-library-for-go/internal/msalbase"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/internal/requests"
 )
+
+// ConfidentialClientApplicationOptions configures the PublicClientApplication's behavior.
+type ConfidentialClientApplicationOptions struct {
+	// Accessor controls cache persistence.
+	// By default there is no cache persistence.
+	Accessor CacheAccessor
+
+	// Client sets the transport for making HTTP requests.
+	// Leave this as nil to use the default HTTP transport.
+	Client HTTPClient
+}
+
+// DefaultConfidentialClientApplicationOptions returns an instance of ConfidentialClientApplicationOptions initialized with default values.
+func DefaultConfidentialClientApplicationOptions() ConfidentialClientApplicationOptions {
+	return ConfidentialClientApplicationOptions{
+		Client: http.DefaultClient,
+	}
+}
 
 // ConfidentialClientApplication is a representation of confidential client applications.
 // These are apps that run on servers (web apps, web API apps, or even service/daemon apps),
@@ -17,10 +38,10 @@ type ConfidentialClientApplication struct {
 	clientCredential  *msalbase.ClientCredential
 }
 
-// CreateConfidentialClientApplication creates a ConfidentialClientApplication instance given a client ID, authority URL and client credential.
-func CreateConfidentialClientApplication(
-	clientID string, authority string, clientCredential ClientCredentialProvider,
-) (*ConfidentialClientApplication, error) {
+// NewConfidentialClientApplication creates a ConfidentialClientApplication instance given a client ID, authority URL and client credential.
+// Pass nil for options to accept the default values; this is the same as passing the result
+// from a call to DefaultConfidentialClientApplicationOptions().
+func NewConfidentialClientApplication(clientID string, authority string, clientCredential ClientCredentialProvider, options *ConfidentialClientApplicationOptions) (*ConfidentialClientApplication, error) {
 	cred, err := createInternalClientCredential(clientCredential)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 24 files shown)


msal/public_client_application.go
```diff
@@ -14,24 +18,32 @@ type PublicClientApplication struct {
...
 }
 
-//SetHTTPManager allows users to use their own implementation of HTTPManager.
-func (pca *PublicClientApplication) SetHTTPManager(httpManager HTTPManager) {
-	webRequestManager := createWebRequestManager(httpManager)
-	pca.clientApplication.webRequestManager = webRequestManager
+// DefaultPublicClientApplicationOptions returns an instance of PublicClientApplicationOptions initialized with default values.
+func DefaultPublicClientApplicationOptions() PublicClientApplicationOptions {
+	return PublicClientApplicationOptions{
+		Client: http.DefaultClient,
```
keegan-caruso, line 35: Would we need to tune DefaultMaxIdleConnsPerHost or MaxIdleConnsPerHost ?

msal/public_client_application.go
```diff
@@ -14,24 +18,32 @@ type PublicClientApplication struct {
...
-	return pca, nil
+// PublicClientApplicationOptions configures the PublicClientApplication's behavior.
+type PublicClientApplicationOptions struct {
+	// Accessor controls cache persistence.
+	// By default there is no cache persistence.
+	Accessor CacheAccessor
+
+	// Client sets the transport for making HTTP requests.
+	// Leave this as nil to use the default HTTP transport.
+	Client HTTPClient
```
keegan-caruso, line 29: Could we just use an http.Client? and users could substitute either a custom httpClient or just the Transport?

msal/public_client_application.go
jhendrixMSFT: I don't believe so but am open to the possibility.

FWIW, in the Azure SDK for Go we've been using the standard library's default HTTP client without any complaints (either it's not an issue or customers are configuring it to their requirements).

msal/public_client_application.go
jhendrixMSFT: We could, but it tightly couples us to this specific implementation.

In the track 2 Go SDK, we use a pipeline middleware that lets us (and SDK consumers) have more control over how HTTP requests/responses are handled.  If we keep this as a generic interface it would allow consumers to inject their own middleware (ideally I'd like to see us being able to use pipeline policies we have in `azcore` as we provide mechanisms like retries, logging, telemetry etc; see [here](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/azcore) for more info).

I wanted to start simple for now, and refine it later.

msal/confidential_client_application.go
```diff
@@ -45,52 +66,48 @@ func createInternalClientCredential(interfaceCred ClientCredentialProvider) (*ms
...
 
 // AcquireTokenByAuthCode is a request to acquire a security token from the authority, using an authorization code.
 // Users need to create an AcquireTokenAuthCodeParameters instance and pass it in.
-func (cca *ConfidentialClientApplication) AcquireTokenByAuthCode(
-	authCodeParams *AcquireTokenAuthCodeParameters) (AuthenticationResultProvider, error) {
+func (cca *ConfidentialClientApplication) AcquireTokenByAuthCode(ctx context.Context, scopes []string, redirectURI string, options *AcquireTokenByAuthCodeOptions) (*msalbase.AuthenticationResult, error) {
+	authCodeParams := createAcquireTokenAuthCodeParameters(scopes, redirectURI)
 	authCodeParams.requestType = requests.AuthCodeConfidential
 	authCodeParams.clientCredential = cca.clientCredential
+	if options != nil {
```
keegan-caruso, line 92: Should cases like this be considered an error where we know that we need something off of the options struct?

test/devapps/username_password_sample.go
```diff
@@ -4,48 +4,50 @@
...
-		log.Info("No valid account found")
-		tryUsernamePasswordFlow(publicClientApp)
-	} else {
-		silentParams := msal.CreateAcquireTokenSilentParametersWithAccount(config.Scopes, userAccount)
-		result, err := publicClientApp.AcquireTokenSilent(silentParams)
+	var result *msalbase.AuthenticationResult
+	var err error
+	if userAccount != nil {
+		// found a cached account, now see if an applicable token has been cached
+		// NOTE: this API conflates error states, i.e. err is non-nil if an applicable token isn't
```
keegan-caruso, line 34: Any suggestions on this?

jhendrixMSFT: I've been thinking about this one a bit.  It's somewhat similar to calling an [io.Reader](https://golang.org/pkg/io/#Reader) returning `io.EOF` at the end of a stream.

I'm not familiar enough with how customers use this functionality.  Does it seem valuable to return a distinct error, e.g. `msal.ErrNoToken` (something like that)?  Are there use-cases where distinguishing this specific case is valuable?

msal/public_client_application.go
henrik-me: not sure if this is the best way to pass in the HttpClient, but certainly a functionality we should support like we do for other libraries

---
In reply to: [514778150](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/68#discussion_r514778150) [](ancestors = 514778150)

henrik-me: Aside: why is the object not called HttpClient, client can be so many things

---
In reply to: [515451140](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/68#discussion_r515451140) [](ancestors = 515451140,514778150)

test/devapps/username_password_sample.go
henrik-me: MSAL errors codes are mostly well known. e.g. it may be a MsalClientException, an MsalServiceException. And there are specific error codes that lets a developer write the code in a specific way. we should do the same here. Java, Python, Node and .NET show cases this

---
In reply to: [515436829](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/68#discussion_r515436829) [](ancestors = 515436829)

msal/public_client_application.go
jhendrixMSFT: Renamed to `HTTPClient`

msal/public_client_application.go
```diff
@@ -70,13 +85,35 @@ func (pca *PublicClientApplication) AcquireTokenByDeviceCode(
...
 }
+
+// AcquireTokenSilentOptions contains the optional parameters to acquire a token silently (from cache).
+type AcquireTokenSilentOptions struct {
+	// Account specifies the account to use when acquiring a token from the cache.
+	Account *msalbase.Account
+}
+
+// AcquireTokenByDeviceCodeOptions contains the optional parameters used to acquire an access token using the device code flow.
+type AcquireTokenByDeviceCodeOptions struct {
```
jhendrixMSFT, line 111: We use this in our track 2 SDKs to reduce the chance of breaking changes.  It works well for generated code where there tends to be more churn.  For hand-written APIs where we can be tactical it might not make as much sense.  Any thoughts on this one?

henrik-me: I think it's fine to have an option object for all flows, though it there is nothing today then likelihood of anything in the future is relatively small. I would though actually have expected some option

---
In reply to: [516285138](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/68#discussion_r516285138) [](ancestors = 516285138)

msal/confidential_client_application.go
```diff
@@ -4,10 +4,31 @@
...
 import (
+	"context"
+	"net/http"
+
 	"github.com/AzureAD/microsoft-authentication-library-for-go/internal/msalbase"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/internal/requests"
 )
 
+// ConfidentialClientApplicationOptions configures the PublicClientApplication's behavior.
+type ConfidentialClientApplicationOptions struct {
```
element-of-surprise, line 15: A few things I'd note here:
- We really need to do something about these names (though probably not in this PR)
- We should not use the options struct. This is non-go standard.  I know the official SDK track 2 guidelines say to do this, but since we aren't in the SDK, we should take advantage of variardic options.  This struct would become private and modified by those args.  It handles better defaults.
- HTTPClient, which we are renaming somewhere above, should just be eliminated in the end.  If we really have a use case, we can add it back later (but I doubt we will).

jhendrixMSFT: The Go blog details both variadic config and optional args struct as cromulent ways for handling configuration.  There's a nice write-up as part of the module compatibility blog post.

https://blog.golang.org/module-compatibility

We (the Go SDK team) went round and round on this, and in the end we felt that the self-documenting properties of the optional params struct is the better customer experience.  Note that we haven't had any usability studies as of yet, so our position may change.

Having a configurable HTTP client allows the use of something like our pipeline middleware for controlling request/response lifetime.  Most common case I the ability to specify a retry policy for handling transient failures.  At any rate we can defer it for now.

jhendrixMSFT: I'm not saying it has to be optional args struct, just providing my perspective (whileI do think there's benefit of having consistency across SDKs, this isn't the hill I'm gonna die on).

henrik-me: We have bring your own HttpClient in all our libraries. Has been a wish so many times that we are providing it out of the box with new libraries we write.

henrik-me:
While there is more work to be done, I think think change makes the MSAL api more clear and is an improvement.

jhendrixMSFT:
Addresses https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/60

test/devapps/device_code_flow_sample.go
```diff
@@ -16,51 +17,45 @@ func deviceCodeCallback(deviceCodeResult msal.DeviceCodeResultProvider) {
...
 	}
-	if userAccount == nil {
-		log.Info("No valid account found")
-		tryDeviceCodeFlow(publicClientApp)
-	} else {
-		silentParams := msal.CreateAcquireTokenSilentParametersWithAccount(config.Scopes, userAccount)
-		result, err := publicClientApp.AcquireTokenSilent(silentParams)
+	var result *msalbase.AuthenticationResult
+	var err error
+	if userAccount != nil {
```
jmprieur, line 37: Do we need to test if the `userAccount` is not `nil`? couldn't just `AcquireTokenSilent` return an error if the `userAccount` is `nil`? (Other MSALs do that)

jmprieur (commented):
LGTM
I left a few comments and questions, mainly about the consistency with the other MSAL libraries

test/devapps/device_code_flow_sample.go
jhendrixMSFT: With the current implementation, a non-nil error is ambiguous; it means that either the user account wasn't found or that something else failed (e.g. perhaps there was a transient IO failure making the network request).

I think we need to decide on the semantics we want here, and also look at what kind of error(s) other implementations expose here.  I've opened https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/70 to track standardizing on error types.

test/devapps/device_code_flow_sample.go
```diff
@@ -16,51 +17,45 @@ func deviceCodeCallback(deviceCodeResult msal.DeviceCodeResultProvider) {
...
-	if err != nil {
-		log.Fatal(err)
-	}
-	publicClientApp.SetCacheAccessor(cacheAccessor)
-	var userAccount msal.AccountProvider
-	accounts := publicClientApp.GetAccounts()
+	// create a PublicClientApplication with a  custom cache accessor
+	options := msal.DefaultPublicClientApplicationOptions()
+	options.Accessor = cacheAccessor
+	publicClientApp := msal.NewPublicClientApplication(config.ClientID, config.Authority, &options)
```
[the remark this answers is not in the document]
jhendrixMSFT: I can certainly move it to the optional params struct.

In our `azidentity` implementation, both the authority and client ID are optional parameters (we use the developer sign-on client ID for the default).  Does that make sense for MSAL?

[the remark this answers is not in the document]
jhendrixMSFT: Moved the authority to default options.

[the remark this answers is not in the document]
jmprieur: In MSAL we would always pass-in the ClientId

msal/public_client_application.go
```diff
@@ -4,64 +4,79 @@
...
 type PublicClientApplication struct {
 	clientApplication *clientApplication
 }
 
-// CreatePublicClientApplication creates a PublicClientApplication instance given a client ID and authority URL.
-func CreatePublicClientApplication(clientID string, authority string) (*PublicClientApplication, error) {
+// NewPublicClientApplication creates a PublicClientApplication instance given a client ID and authority URL.
+// Pass nil for options to accept the default values; this is the same as passing the result
+// from a call to DefaultPublicClientApplicationOptions().
+func NewPublicClientApplication(clientID string, authority string, options *PublicClientApplicationOptions) *PublicClientApplication {
```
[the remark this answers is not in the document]
jhendrixMSFT: Fixed.

msal/public_client_application_test.go
```diff
@@ -51,26 +51,18 @@ func TestCreateAuthCodeURL(t *testing.T) {
...
-	authCodeParams := &AcquireTokenAuthCodeParameters{
-		commonParameters: tokenCommonParams,
-	}
 	wrm.On("GetTenantDiscoveryResponse",
 		"https://login.microsoftonline.com/v2.0/v2.0/.well-known/openid-configuration").Return(tdr, nil)
 	actualTokenResp := &msalbase.TokenResponse{}
 	wrm.On("GetAccessTokenFromAuthCode", mock.AnythingOfType("*msalbase.AuthParametersInternal"), "", "", make(map[string]string)).Return(actualTokenResp, nil)
 	cacheManager.On("CacheTokenResponse", mock.AnythingOfType("*msalbase.AuthParametersInternal"), actualTokenResp).Return(testAcc, nil)
-	_, err := testPCA.AcquireTokenByAuthCode(authCodeParams)
+	_, err := testPCA.AcquireTokenByAuthCode(context.Background(), []string{"openid"}, "", nil)
```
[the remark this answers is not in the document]
jhendrixMSFT: I've opened https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/72 to track this.

test/devapps/authorization_code_sample.go
```diff
@@ -41,10 +42,10 @@ func getToken(w http.ResponseWriter, r *http.Request) {
 	}
 	code := codes[0]
 	// Getting the access token using the authorization code
-	authCodeParams := msal.CreateAcquireTokenAuthCodeParameters(config.Scopes, config.RedirectURI)
-	authCodeParams.Code = code
-	authCodeParams.CodeChallenge = config.CodeChallenge
-	result, err := publicClientApp.AcquireTokenByAuthCode(authCodeParams)
+	result, err := publicClientApp.AcquireTokenByAuthCode(context.Background(), config.Scopes, config.RedirectURI, &msal.AcquireTokenByAuthCodeOptions{
```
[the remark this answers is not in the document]
jhendrixMSFT: Is it always the same URL?  What about for Azure Stack and/or sovereign clouds?

msal/confidential_client_application.go
```diff
@@ -45,52 +66,48 @@ func createInternalClientCredential(interfaceCred ClientCredentialProvider) (*ms
...
 	authParams := cca.clientApplication.clientApplicationParameters.createAuthenticationParameters()
+	clientCredParams := createAcquireTokenClientCredentialParameters(scopes)
 	clientCredParams.augmentAuthenticationParameters(authParams)
 	req := requests.CreateClientCredentialRequest(cca.clientApplication.webRequestManager, authParams, cca.clientCredential)
 	return cca.clientApplication.executeTokenRequestWithCacheWrite(req, authParams)
 }
 
-// GetAccounts gets all the accounts in the token cache.
-func (cca *ConfidentialClientApplication) GetAccounts() []AccountProvider {
+// Accounts gets all the accounts in the token cache.
```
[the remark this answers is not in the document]
jhendrixMSFT: Opened https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/73 to track this.

test/devapps/username_password_sample.go
jhendrixMSFT: I took a cursory look at the C# implementation and it doesn't appear to throw an exception if the account isn't found (it's not doc'ed that way, and I dug a bit into the implementation and didn't see anyplace where it obviously throws).

henrik-me: GetAccounts doesn't throw. My comment was arround the API in general (sorry for not being clear about that). My main comment on errors was that we should take that as a separate work item (already tracked elsewhere). We should simply follow the std. msal patterns for user flows and that will lead to specific expected errors that guide the developer and the user experience.

test/devapps/authorization_code_sample.go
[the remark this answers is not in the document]
jmprieur: No it's not always the same URL The URL depends on the cloud, but this could be part of the options? (most customers would use the public Azure cloud?)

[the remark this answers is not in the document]
jhendrixMSFT: I see.  So just to confirm, we always specify this same redirect URL but since we never follow it there's no need to make it configurable, is that correct?

keegan-caruso (approved):
LGTM. Definitely a step in the right direction.

test/devapps/authorization_code_sample.go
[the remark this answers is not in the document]
jhendrixMSFT: The `redirectURI` param has been removed.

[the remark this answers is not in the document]
jhendrixMSFT: @jmprieur I've removed the param, anything else or is this good to merge?

jhendrixMSFT:
Fixes https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/56

jennyf19 (approved):
:shipit:

PR #74, opened 2020-11-06
Remove go-render

keegan-caruso:
Use godebug and refactor test code to use
godebug compare functions instead of reflection


keegan-caruso:
duplication is not very useful in table driven tests. I modified the duplication exclusion pattern to include test files. Should show next run

keegan-caruso:
fixed in #79

PR #77, opened 2020-11-09
Adding json custom encoder/decoder

element-of-surprise:
I was unable to locate an encoder/decoder that did what I wanted here.  Third-party json packages are really focused around speed, mostly by going to generated calls or reducing allocations.  

Unfortunately, I think the best way to look at this is the way you'd look at third-party code. You'd use it to see if it did what you wanted, and if it did, you'd import it (kinda like what you do with encoding/json).  

This code isn't going to be for the feint of heart.  It must use reflection and uses a state machine design.  This just means it is hard to follow, just like anything that uses reflection.  In essence, this is a JSON parser and json.Encoder/Decoder act as the lexer.  

I've included the design doc in markdown in the directory.  

This might be adding more code than is being removed.  However, when there is an error in this code, this seems to be a straightforward error that is returned vs silent errors when keys just don't get recorded.  In the most general sense, if the calls that use it work, then it works vs. the current situation where the tests will likely be silent.  

I did not do the integrations here, just the package add.  I'm still working on the integrations, as this necessitated a bunch of type conversions from pointers (yea!), interface changes, fixing tests, etc....  About 90% done there.  Once this goes in I'll send the next up.

### DIFF
--- a/internal/json/design.md
+++ b/internal/json/design.md
@@ -45,7 +45,7 @@ They provide streaming processing (efficient) and return errors on bad JSON.
 Support for json.Marshaler/Unmarshaler allows for us to use non-basic types
 that must be specially encoded/decoded (like time.Time objects).
 
-We don't support types that can't custome unmarshal or have AdditionalFields
+We don't support types that can't customer unmarshal or have AdditionalFields
 in order to prevent future devs from forgetting that important field and
 generating bad return values.
 
--- a/internal/json/design.md
+++ b/internal/json/design.md
@@ -1,0 +1,140 @@
+# JSON Package Design
+Author: John Doak(jdoak@microsoft.com)
+
+## Why?
+
+This project needs a special type of marshal/unmarshal not directly supported
+by the encoding/json package. 
+
+The need revolves around a few key wants/needs:
+- unmarshal and marshal structs representing JSON messages
+- fields in the messgage not in the struct must be maintained when unmarshalled
+- those same fields must be marshalled back when encoded again
+
+The initial version used map[string]interface{} to put in the keys that
+were known and then any other keys were put into a field called AdditionalFields.
+
+This has a few negatives:
+- Dual marshaling/unmarshalling is required
+- Adding a struct field requires manually adding a key by name to be encoded/decoded from the map (which is a loosely coupled construct), which can lead to bugs that aren't detected or have bad side effects
+- Tests can become quickly disconnected if those keys aren't put
+in tests as well. So you think you have support working, but you
+don't. Existing tests were found that didn't test the marshalling output.
+- There is no enforcement that if AdditionalFields is required on one struct, it should be on all containers
+that don't have custom marshal/unmarshal.
+
+This package aims to support our needs by providing custom Marshal()/Unmarshal() functions.
+
+This prevents all the negatives in the initial solution listed above. However, it does add its own negative:
+- Custom encoding/decoding via reflection is messy (as can be  seen in encoding/json itself)
+
+Go proverb: Reflection is never clear
+Suggested reading: https://blog.golang.org/laws-of-reflection
+
+## Important design decisions
+
+- We don't want to understand all JSON decoding rules
+- We don't want to deal with all the quoting, commas, etc on decode
+- Need support for json.Marshaler/Unmarshaler, so we can support types like time.Time
+- If struct does not implement json.Unmarshaler, it must have AdditionalFields defined
+- We only support root level objects that are \*struct or struct
+
+To faciliate these goals, we will utilize the json.Encoder and json.Decoder.
+They provide streaming processing (efficient) and return errors on bad JSON.
+
+Support for json.Marshaler/Unmarshaler allows for us to use non-basic types
+that must be specially encoded/decoded (like time.Time objects).
+
+We don't support types that can't custome unmarshal or have AdditionalFields
+in order to prevent future devs from forgetting that important field and
+generating bad return values.
+
+Support for root level objects of \*struct or struct simply acknowledges the
+fact that this is designed only for the purposes listed in the Introduction.
+Outside that (like encoding a lone number) should be done with the
+regular json package (as it will not have additional fields).
+
+We don't support a few things on json supported reference types and structs:
+- \*map: no need for pointers to maps
+- \*slice: no need for pointers to slices
+- any further pointers on struct after \*struct
+
+There should never be a need for this in Go.
+
+## Design
+
+## State Machines
+
+This uses state machine designs that based upon the Rob Pike talk on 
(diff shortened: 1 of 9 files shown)


internal/json/struct.go
```diff
@@ -0,0 +1,288 @@
...
+
+	t, isPtr, err := fieldBaseType(d.value, goName)
+	if err != nil {
+		return nil, fmt.Errorf("type(%s) had field(%s) %w", d.value.Type().Name(), goName, err)
+	}
+
+	switch t.Kind() {
+	// We need to recursively call ourselves on any *struct or struct.
+	case reflect.Struct:
+		if isPtr {
```
keegan-caruso, line 134: I know that in general you shouldn't, but what about the odd cases of ptr to map or slice?  This will panic. Should we error instead?

internal/json/design.md
```diff
@@ -0,0 +1,140 @@
...
+- If struct does not implement json.Unmarshaler, it must have AdditionalFields defined
+- We only support root level objects that are \*struct or struct
+
+To faciliate these goals, we will utilize the json.Encoder and json.Decoder.
+They provide streaming processing (efficient) and return errors on bad JSON.
+
+Support for json.Marshaler/Unmarshaler allows for us to use non-basic types
+that must be specially encoded/decoded (like time.Time objects).
+
+We don't support types that can't custome unmarshal or have AdditionalFields
```
keegan-caruso, line 48: nit: "custome"

internal/json/struct.go
element-of-surprise: - Since we aren't a general purpose package, I just didn't think it was worth the support
- If we go down the road of needing to support these kinds of things (which we shouldn't be doing), we will end up needing to support things like ****map[string]**[]*map[string]interface{}.  I mean, its really just building in more recursion and tracking to support everything, but I'm not sure its worth the dollars in time.
- There just isn't a reason to let MSAL developers add these types.  This should be panicing before anything ships.  Even if we return an error, we are basically saying "the package is broken" for this op, forever.

internal/json/design.md
element-of-surprise: fixed.

internal/json/mapslice.go
```diff
@@ -0,0 +1,331 @@
...
+		return err
+	}
+
+	return nil
+}
+
+type sliceWalk struct {
+	dec           *json.Decoder
+	s             reflect.Value // *[]slice
+	skipOpenParen bool
```
keegan-caruso, line 193: this is unused

internal/json/struct.go
```diff
@@ -0,0 +1,288 @@
...
+
+	dec := newDecoder(jdec, v)
+	return dec.run()
+}
+
+type decoder struct {
+	dec        *json.Decoder
+	value      reflect.Value // This will be a reflect.Struct
+	translator translateFields
+	hasAdd     bool
```
keegan-caruso, line 42: this is unused

internal/json/json.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	return false
+}
+
+// hasMarshalJSON will determine if the value or a pointer to this value has
+// the MarshalJSON method.
+func hasMarshalJSON(v reflect.Value) bool {
+	if method := v.MethodByName(marshalJSON); method.Kind() != reflect.Invalid {
+		_, ok := v.Interface().(json.Marshaler)
+		if ok {
```
keegan-caruso, line 104: This pattern is used a few places in this file, could be simplified down to `return ok`

internal/json/json.go
```diff
@@ -0,0 +1,186 @@
...
+		return nil, err
+	}
+	return buff.Bytes(), nil
+}
+
+// Unmarshal unmarshals a []byte representing JSON into i, which must be a *struct. In addition, if the struct has
+// a field called AdditionalFields of type map[string]interface{}, JSON data representing fields not in the struct
+// will be written as key/value pairs to AdditionalFields.
+func Unmarshal(b []byte, i interface{}) error {
+	if b == nil || len(b) == 0 {
```
keegan-caruso, line 59: nit: this is more explicit but len is safe for nil slices

internal/json/mapslice.go
```diff
@@ -0,0 +1,331 @@
...
+	}
+	return nil, nil
+}
+
+func (s *sliceWalk) next() (stateFn, error) {
+	if s.dec.More() {
+		return s.storeValue, nil
+	}
+	// Nothing left in the slice, remove closing ]
+	s.dec.Token()
```
keegan-caruso, line 256: Token() potentially returns an error, is this safe to ignore here?

element-of-surprise: Yes, because this is going to pop up to some other value and the decoder will error immediately if there is a parse problem.  But, just in case of a refactor that nullifies that, I went ahead and captured it.  Also, I should have documented it if I don't use an error.

keegan-caruso (approved):
LGTM

henrik-me:
I'm a bit surprised this is needed. I read your explanation. Maintaining a json parser (or at least some levels of it) is something I would assume in most places is discouraged. Speed may be a factor for service operations btw. I'm actually surprised Google doesn't have a json parser we can use?

element-of-surprise:
encoding/json is the Go language lexer/parser.  What it doesn't handle is the "AdditionalFields" requirement that we have.  The generic way to handle this is map[string]interface{}, which has issues of keeping things synced.  That is Go's generic solution for unstructured data.

I can't speak to every API out there, but the ability to retain additional fields to push back up is new to me.  I've seen this done with a type of merge API call on the backend that takes care of that vs pushing to the client.

Google doesn't build their APIs (internally at least) to have this type of problem and they don't natively use JSON.  Google uses protocol buffers and when you see JSON, it has really been converted from a proto.  At Google, everything is structured protocol buffer.

There are quite a few third-party json libraries, but they concentrate on lower allocations and speed, usually with less features.

This uses the lexer that is part of encoding/json and should be faster than the map[string]interface{} code with less allocations for the general case.  Just the changes I'm making away from pointers will also give a decent speed increase because it will eliminate a lot of heap allocation and gc cleanup. The code, as it was when I started, was not written with speed in mind.

I don't expect any maintenance as part of this package add.  I've occasionally had to write similar code for other types of decodes with zero maintenace overhead.  I loathe writing this type of code, but I loathe the bugs in forgetting to add fields or forgetting to add "AdditionalFields" to a new struct, or not testing the field more.  To be honest, the moment I had to do this work I was regretting taking this on.

But as for the "risk", which is a valid point, it is about as risky as a third-party lib.  If you have a bug in those, you either must fix it yourself or wait for the authors to fix it.  Rarely do people go through the entire code to understand it.  In this case, you had a full review by someone else who understands the code.

John Doak
jdoak@microsoft.com

PR #79, opened 2020-11-12
JSON integration, pointer updates, ...

element-of-surprise:
This CL's main purpose was to update with JSON, but as part of that I needed to update some pointers to non-pointers.

This kinda cascaded into a bunch of changes.  Not everything that can be de-pointered was touched.

This also led to finding a few bugs and test re-writes.

Finally, there are a ton of comments in here so I an


internal/msalbase/token_response.go
```diff
@@ -22,76 +22,78 @@ type tokenResponseJSONPayload struct {
...
 	ExtExpiresOn   time.Time
 	rawClientInfo  string
-	ClientInfo     *ClientInfoJSONPayload
+	ClientInfo     ClientInfoJSONPayload
 }
 
-//HasAccessToken checks if the TokenResponse has an access token secret
-func (tr *TokenResponse) HasAccessToken() bool {
+// HasAccessToken checks if the TokenResponse has an access token secret.
+func (tr TokenResponse) HasAccessToken() bool {
```
jennyf19, line 51: nit: I would remove "secret" from this line and line 55, that doesn't make sense. not sure why it was there.

internal/msalbase/token_response.go
```diff
@@ -114,11 +116,12 @@ func CreateTokenResponse(authParameters *AuthParametersInternal, responseCode in
 
 	idToken, err := CreateIDToken(payload.IDToken)
 	if err != nil {
-		//ID tokens aren't always returned, so the error is just logged
+		// ID tokens aren't always returned, so the error is just logged
+		// TODO(jdoak): we should probably remove this. Either this is an error or isn't.
 		log.Errorf("ID Token error: %v", err)
```
jennyf19, line 121: please remove

internal/requests/username_password_request.go
```diff
@@ -5,55 +5,57 @@ package requests
...
-		if mexDoc, err := req.webRequestManager.GetMex(userRealm.FederationMetadataURL); err == nil {
-			wsTrustEndpoint := mexDoc.UsernamePasswordEndpoint
-			if wsTrustResponse, err := req.webRequestManager.GetWsTrustResponse(req.authParameters, userRealm.CloudAudienceURN, &wsTrustEndpoint); err == nil {
-				if samlGrant, err := wsTrustResponse.GetSAMLAssertion(&wsTrustEndpoint); err == nil {
-					return req.webRequestManager.GetAccessTokenFromSamlGrant(req.authParameters, samlGrant)
-				}
-			}
+		mexDoc, err := req.webRequestManager.GetMex(userRealm.FederationMetadataURL)
+		if err != nil {
+			// TODO(someone): check for ui interaction in api result...
```
jennyf19, line 43: >/ [](start = 3, length = 1)

I don't know what this means. Usually if there is an issue, MSALs through a UiRequiredException, which app developers catch and then issue an interactive request. With U/P request, however, there is no UI. So....I don't know what this comment means, do you know @henrik-me?

internal/requests/username_password_request.go
```diff
@@ -5,55 +5,57 @@ package requests
...
+			return msalbase.TokenResponse{}, err
+		}
+		wsTrustEndpoint := mexDoc.UsernamePasswordEndpoint
+		wsTrustResponse, err := req.webRequestManager.GetWsTrustResponse(req.authParameters, userRealm.CloudAudienceURN, wsTrustEndpoint)
+		if err != nil {
+			return msalbase.TokenResponse{}, err
+		}
+		samlGrant, err := wsTrustResponse.GetSAMLAssertion(wsTrustEndpoint)
+		if err != nil {
+			log.Println("here")
```
jennyf19, line 53: nit: why println here?

internal/msalbase/account_test.go
```diff
@@ -79,12 +93,18 @@ func TestAccountConvertToJSONMap(t *testing.T) {
...
 	if err != nil {
-		t.Errorf("Error should be nil, but it is %v", err)
+		panic(err)
 	}
-	if !reflect.DeepEqual(jsonMap, actualJSONMap) {
-		t.Errorf("JSON account %+v differs from expected JSON account %+v", jsonMap, actualJSONMap)
+
+	got := map[string]interface{}{}
+	if err := stdJSON.Unmarshal(b, &got); err != nil {
+		panic(err)
```
henrik-me, line 103: any reason these are preferred to be panics over asserts? Assuming because it's core functionality and nothing really with the MSAL functionality itself?

internal/msalbase/authentication_result.go
```diff
@@ -5,82 +5,67 @@ package msalbase
...
-		if err != nil {
-			return nil, errors.New("access token in cache expires at an invalid time")
-		}
-		grantedScopes = SplitScopes(storageTokenResponse.accessToken.GetScopes())
-	} else {
-		return nil, errors.New("no access token present in cache")
+	accessToken := storageTokenResponse.AccessToken.GetSecret()
+	expiresOn, err := ConvertStrUnixToUTCTime(storageTokenResponse.AccessToken.GetExpiresOn())
+	if err != nil {
+		return AuthenticationResult{}, errors.New("access token in cache expires at an invalid time")
```
henrik-me, line 32: >access token in cache expires at an invalid time [](start = 45, length = 48)

nit: interesting wording

internal/tokencache/default_storage_manager.go
```diff
@@ -84,46 +89,57 @@ func (m *defaultStorageManager) WriteAccessToken(accessToken *accessTokenCacheIt
...
+	}
 
+	// TODO(jdoak): So, I redid this code to make it clearer that we
+	// changed order based on if FamilyID was set. But I'm not sure
+	// why I have to loop twice over the tokens instead of just
+	// trying both matchers.  If someone could shed some light on this
+	// I could make a note for future maintainers. If order doesn't
+	// matter, then we could ditch byFamily and byClient for cleaner
+	// code.
+	for _, matcher := range matchers {
```
jennyf19, line 121: @bgavrilMS can you review this part? thanks.

internal/msalbase/authentication_result_test.go
```diff
@@ -7,85 +7,91 @@ import (
 	"reflect"
 	"testing"
 	"time"
+
+	"github.com/kylelemons/godebug/pretty"
```
henrik-me, line 11: >"github.com/kylelemons/godebug/pretty" [](start = 1, length = 38)

any reason for this? if so why not everywhere?

internal/wstrust/saml_token_info.go
```diff
@@ -5,6 +5,11 @@ package wstrust
 
 type SamlAssertionType int
 
+// TODO(msal reviewer): I love iota, but...  It is only safe if this value
+// is never sent to another service or written to disk for retrieval. Otherwise
+// the value should be statically assigned.  If someone can verify that.
+// Also, I couldn't figure out why not just use msalbase.SAMLV1Grant... here
+// instead of these values, instead we insert them in default_web_request_manager.go.
```
jennyf19, line 12: I would not use the iota, based on some quick reading. using msalbase.SAMLV1Grant makes sense

msal/client_application_parameters.go
```diff
@@ -5,26 +5,32 @@ package msal
...
-	return p
-}
-
-func (p *clientApplicationParameters) setAadAuthority(authorityURI string) {
-	p.commonParameters.setAadAuthority(authorityURI)
+func createClientApplicationParameters(clientID, authorityURI string) (*clientApplicationParameters, error) {
+	// NOTE: I moved setADDAuthority here.  It called a method that was only
+	// used on the output of this function, which is only called here.  It also
+	// ignored the error output. That seems buggy (anytime you ignore an error, must document why).
+	cp, err := createApplicationCommonParameters(clientID, authorityURI)
```
jennyf19, line 20: >createApplicationCommonParameters [](start = 12, length = 33)

do you mean "removed" and not "moved"? i think it was here because we have different authority types, like AAD, ADFS, or B2C, so eventually, we will need to handle the ADFS and B2C cases (but not in that order). I think what is here is fine, as we'll tackle that issue when we get to it.

test/devapps/authorization_code_sample.go
```diff
@@ -53,12 +53,19 @@ func getToken(w http.ResponseWriter, r *http.Request) {
 	fmt.Fprintf(w, "Access token is "+result.GetAccessToken())
 }
 
+// TODO(msal expert): This test doesn't do anything that I can tell. We call msal.NewPublicClientApplication(), but we
+// don't use it.  Someone needs to clarify to me how this is supposed to work.
```
jennyf19, line 57: yeah, this is on my list of TODOs. This sample didn't make sense to me, but I see someone has [tested it and checked it off](https://github.com/AzureAD/microsoft-authentication-library-for-go/projects/2#card-47369448), so idk. will try to take a look this weekend.

jennyf19 (approved):
:shipit:

jennyf19:
Thanks @element-of-surprise left a few comments and opened a few issues. idk if i'll have time to go back through it, i would like to spend more time on some parts, but time is a little fleeting atm.

---
In reply to: [726384926](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/79#issuecomment-726384926) [](ancestors = 726384926)

internal/json/json_test.go
```diff
@@ -0,0 +1,186 @@
...
+				}
+				`,
+			),
+			got: &StructA{},
+			want: &StructA{
+				Name: "John",
+				ID:   3,
+				Meta: &StructB{
+					Address: "291 Street",
+					AdditionalFields: map[string]interface{}{
```
bgavrilMS: Very elegant, in MSAL.NET we did this by extracting each known field from the JSON object and keeping whatever remains as "AdditionalFields". While it works, the code is much more verbose and the token caching code has more concerns that it needs.

internal/msalbase/Account.go
```diff
@@ -38,56 +38,21 @@ func NewAccount(homeAccountID, env, realm, localAccountID, authorityType, userna
...
 }
 
 //GetEnvironment returns the environment of an account
-func (acc *Account) GetEnvironment() string {
+func (acc Account) GetEnvironment() string {
 	return acc.Environment
 }
-
-//PopulateFromJSONMap populates an account object from a map (used for cache deserialization)
-func (acc *Account) PopulateFromJSONMap(j map[string]interface{}) error {
```
bgavrilMS, line 61: So cool to be able to get rid of this code.

bgavrilMS, line 61: Do you have an idea of how the perf changes ? I think that the original method does not use reflection.

internal/msalbase/authentication_result.go
```diff
@@ -5,82 +5,67 @@ package msalbase
...
 
-// CreateAuthenticationResult creates an AuthenticationResult
-func CreateAuthenticationResult(tokenResponse *TokenResponse, account *Account) (*AuthenticationResult, error) {
-	grantedScopes := tokenResponse.GrantedScopes
-	declinedScopes := tokenResponse.declinedScopes
-	//Checking if there are any scopes that were declined, this would lead to an invalid access token
-	if len(declinedScopes) > 0 {
-		return nil, errors.New("token response failed because declined scopes are present")
+// CreateAuthenticationResult creates an AuthenticationResult.
+// TODO(jdoak): Make this a method on TokenResponse() that takes only 1 arge, Account.
```
bgavrilMS, line 48: Please do the todo or delete it.

internal/msalbase/authentication_result.go
```diff
@@ -5,82 +5,67 @@ package msalbase
...
+func CreateAuthenticationResult(tokenResponse TokenResponse, account Account) (AuthenticationResult, error) {
+	if len(tokenResponse.declinedScopes) > 0 {
+		return AuthenticationResult{}, errors.New("token response failed because declined scopes are present")
```
bgavrilMS, line 51: Could we print the declined scopes in the error message. I know this is not your code...

internal/msalbase/authentication_result.go
```diff
@@ -5,82 +5,67 @@ package msalbase
...
-	expiresOn := tokenResponse.ExpiresOn
-	ar := &AuthenticationResult{account, idToken, accessToken, expiresOn, grantedScopes, declinedScopes}
-	return ar, nil
+	return AuthenticationResult{
+		Account:        account,
+		idToken:        tokenResponse.IDToken,
+		AccessToken:    tokenResponse.AccessToken,
+		ExpiresOn:      tokenResponse.ExpiresOn,
+		GrantedScopes:  tokenResponse.GrantedScopes,
+		DeclinedScopes: nil,
```
bgavrilMS, line 59: nit: not sure what's the point of exposing this field if it's always null.

bgavrilMS (approved):
LGTM

internal/msalbase/id_token.go
```diff
@@ -29,32 +31,45 @@ type IDToken struct {
...
 	idToken.RawToken = jwt
 	return idToken, nil
 }
 
-// GetLocalAccountID extracts an account's local account ID from an ID token
-func (idToken *IDToken) GetLocalAccountID() string {
-	if idToken.Oid != "" {
-		return idToken.Oid
+// IsZero indicates if the IDToken is the zero value.
+func (i IDToken) IsZero() bool {
```
bgavrilMS, line 59: Is this a language idiom to avoid using null? The IdToken is optional in some flows, indeed. Is this function supposed to capture this?

element-of-surprise: Some of the code relied on IDToken == nil, as you've indicated.  This allows you to detect the same condition with a non-pointer as you correctly guessed.  You'll see this naming for a method in a few places in stdlib and has enough use to be called an idiom.  In open source code I'd say that overuse of pointers for detection is probably more prevalent.  That is usually a mistake if you want performance, as  it might escape and get put on the heap.

This can be done easier in some cases, like:

```go
type myStruct struct {
  name string
}

var zeroMyStruct = myStruct{}

func (m myStruct) isZero() bool {
  if m == zeroMyStruct {
    return true
  }
  return false
```

But you can't do struct comparisons if you have slices or maps in the struct (sigh).

internal/requests/username_password_request.go
element-of-surprise: FYI, this comment was from the original developer, so I'm not sure what it meant.

internal/msalbase/account_test.go
element-of-surprise: I'm going to assume by assert you mean t.Fatalf().  So this is something that is more my style than a Go thing really.  

But it is as exactly as you say, setup code I panic in because that code is not part of the test.  At a quick glance it tells me the test is completely broken vs an unexpected outcome from the test.

internal/msalbase/authentication_result.go
element-of-surprise: I'm not sure what you mean from the comment.  FYI errors.New() is the original error (the + probably just means I moved it around).  

Do you want:

return AuthenticationResult{}, fmt.Errorf("access token validation error on field expiresOn: %w", err)

And then go into ConvertStrUnixToUTCTime(which I am going to get rid of) and have it return:

fmt.Errorf("time field was an unrecognizable string(%s)", str)

internal/msalbase/authentication_result_test.go
element-of-surprise: pretty can display differences in two types in a human readable way that reflect.DeepEqual() doesn't give and the other package you had(something I think was called go-render, which I removed) doesn't do well.  Kyle was a Go author, this is used all the time inside Google, so its a package I trust, except when dealing with protocol buffers (you have to use special compare package for those).  

It hasn't gone everywhere because I haven't fixed up every test.  As I have to touch tests, I replace as much as I can to make the output actually readable for errors.  

In the end, there will be no reflect.DeepEqual() and for comparisons of structs, I will always use pretty.

internal/msalbase/Account.go
element-of-surprise: It should be faster on encode/decode (though I haven't benchmarked it, but should be a slight edge).   So this was a double conversion, once reading everything into a type and then stuffing it into an interface{}, which did a heap allocation for all fields, regardless.  Assertions are reflections of a type underneath too.  

That speed will probably be irrelevant.  Bigger problems would be the creation of so many objects that have to get gc'd.  Lots of small objects on the gc is the Go performance killer.

I've actually got a few more ideas on how to reduce allocations from this with AdditionalFields, but I'll add that "upgrade" later.

I can't say for this service, but as far as raw speed goes, I found in Kusto that the network couldn't send fast enough to make a difference in speed on any of these conversion methods, it was the biggest limiter.  However, allocations are where you can affect overall performance.  

Here's a question for some future changes, does MSAL backends support gzip or deflate compression?  If so I should add that in for client sends (I'm going to add it for returns, but that gets negotiated where the send does not).  A fix I put in on Kusto's backend gave us 2x performance with compression on their apps and web tooling.

internal/msalbase/authentication_result.go
element-of-surprise: This is another example of github needing a better diff engine.  It's hiding what was really changed:
Original code:

```go
// CreateAuthenticationResult creates an AuthenticationResult
func CreateAuthenticationResult(tokenResponse *TokenResponse, account *Account) (*AuthenticationResult, error) {
	grantedScopes := tokenResponse.GrantedScopes
	declinedScopes := tokenResponse.declinedScopes
	//Checking if there are any scopes that were declined, this would lead to an invalid access token
	if len(declinedScopes) > 0 {
		return nil, errors.New("token response failed because declined scopes are present")
	}
	idToken := tokenResponse.IDToken
	accessToken := tokenResponse.AccessToken
	expiresOn := tokenResponse.ExpiresOn
	ar := &AuthenticationResult{account, idToken, accessToken, expiresOn, grantedScopes, declinedScopes}
	return ar, nil
}
```
This was returning the equivalent of a nil declinded scopes when it did its check, by returning a nil value.  You will never get that field filled in the original code.  

I didn't remove it, because I wasn't sure if that field can be set somewhere else.  If not, then I agree that it should not just unexposed, it should just be eliminated.  

Let me know what you want me to do here.

internal/msalbase/authentication_result.go
element-of-surprise: This is another example of github needing a better diff engine.  It's hiding what was really changed:
Original code:

```go
// CreateAuthenticationResult creates an AuthenticationResult
func CreateAuthenticationResult(tokenResponse *TokenResponse, account *Account) (*AuthenticationResult, error) {
	grantedScopes := tokenResponse.GrantedScopes
	declinedScopes := tokenResponse.declinedScopes
	//Checking if there are any scopes that were declined, this would lead to an invalid access token
	if len(declinedScopes) > 0 {
		return nil, errors.New("token response failed because declined scopes are present")
	}
	idToken := tokenResponse.IDToken
	accessToken := tokenResponse.AccessToken
	expiresOn := tokenResponse.ExpiresOn
	ar := &AuthenticationResult{account, idToken, accessToken, expiresOn, grantedScopes, declinedScopes}
	return ar, nil
}
```
This was returning the equivalent of a nil declinded scopes when it did its check, by returning a nil value.  You will never get that field filled in the original code.  

I didn't remove it, because I wasn't sure if that field can be set somewhere else.  If not, then I agree that it should not just unexposed, it should just be eliminated.  

Let me know what you want me to do here.

internal/msalbase/auth_parameters_internal.go
```diff
@@ -35,8 +35,10 @@ type AuthParametersInternal struct {
 }
 
 //CreateAuthParametersInternal creates an authorization parameters object
```
keegan-caruso, line 37: space before, period after + potentially rename to NewAuthParametersInternal

internal/requests/auth_code_request.go
```diff
@@ -19,33 +21,30 @@ const (
...
 	RequestType       AuthCodeRequestType
 }
 
 // CreateAuthCodeRequest creates an instance of AuthCodeRequest
-func CreateAuthCodeRequest(
-	webRequestManager WebRequestManager,
-	authParameters *msalbase.AuthParametersInternal,
-	reqType AuthCodeRequestType) *AuthCodeRequest {
-	req := &AuthCodeRequest{
+func CreateAuthCodeRequest(webRequestManager WebRequestManager, authParameters msalbase.AuthParametersInternal, reqType AuthCodeRequestType) *AuthCodeRequest {
```
keegan-caruso, line 32: Did this one need to stay a pointer?

internal/requests/auth_code_request_test.go
```diff
@@ -9,62 +9,92 @@ import (
...
+
+func createACRTestParams() msalbase.AuthParametersInternal {
+	tdr := createTDR()
+	params := msalbase.CreateAuthParametersInternal("clientID", createTestAuthorityInfo())
+	// TODO(msal expert): This is the only change here that is an actual change of the test. This is
+	// also changed in other mock tests. Found that calls to methods like AuthCodeRequest.Execute()
+	// would set these Endpoints. But for some reason, none of the mocks expected it.
+	// I'm not sure why this worked (another reason for not using mocks), but I am
+	// pretty sure the mock call to GetAccessTokenFromAuthCode should have seen these.
+	// So I fixed it here, but someone should verify I'm not doing something bad.
```
keegan-caruso, line 36: It looks like the Mocks where only used to set return values, and not actually verify all the calls. I'm not seeing any calls to `AssertExpectations`

internal/msalbase/auth_parameters_internal.go
keegan-caruso: ... Probably a lot of renames if we move away from Create*, this isn't the place for that.

internal/requests/device_code_response.go
```diff
@@ -4,40 +4,46 @@
...
 	"github.com/AzureAD/microsoft-authentication-library-for-go/internal/msalbase"
 )
 
 //DeviceCodeResponse represents the HTTP response received from the device code endpoint
 type DeviceCodeResponse struct {
-	BaseResponse    *msalbase.OAuthResponseBase
+	// TODO(jdoak): Ask someone about why BaseResponse doesn't have a tag.
+	// Either it should be encoded and we should tag it or we should tag it
+	// to be omitted on export or private.
+	BaseResponse    msalbase.OAuthResponseBase
```
keegan-caruso, line 18: Not sure, but this almost looks like it was suppose to be used as an embedded type...

internal/requests/device_code_response.go
```diff
@@ -4,40 +4,46 @@
...
+	if err := json.Unmarshal([]byte(responseData), &dcResponse); err != nil {
+		return dcResponse, err
 	}
 	dcResponse.BaseResponse = baseResponse
 	return dcResponse, nil
 }
 
 //ToDeviceCodeResult converts the DeviceCodeResponse to a DeviceCodeResult
-func (dcr *DeviceCodeResponse) ToDeviceCodeResult(clientID string, scopes []string) *msalbase.DeviceCodeResult {
+func (dcr *DeviceCodeResponse) ToDeviceCodeResult(clientID string, scopes []string) msalbase.DeviceCodeResult {
```
keegan-caruso, line 46: Does this still need a pointer receiver?

internal/tokencache/cache_serialization_contract_test.go
```diff
@@ -38,155 +39,161 @@ func TestCacheSerializationContractUnmarshalJSON(t *testing.T) {
 	jsonFile, err := os.Open(testFile)
```
keegan-caruso, line 39: These errors should be checked and errored/panicked on

msal/acquire_token_silent_parameters.go
```diff
@@ -9,34 +9,33 @@ import (
...
-	clientCredential *msalbase.ClientCredential
+	clientCredential msalbase.ClientCredential
 }
 
 // CreateAcquireTokenSilentParameters creates an AcquireTokenSilentParameters instance with an empty account.
 // This can be used in the case where tokens are acquired as the application instelf.
-func createAcquireTokenSilentParameters(scopes []string) *acquireTokenSilentParameters {
-	p := &acquireTokenSilentParameters{
+func CreateAcquireTokenSilentParameters(scopes []string) AcquireTokenSilentParameters {
+	p := AcquireTokenSilentParameters{
```
keegan-caruso, line 22: nit: couple other places simplified these style of returns, could do the same here

keegan-caruso (approved):
Overall, looks  better. Couple questions, not sure on all of your questions either.

internal/msalbase/token_response.go
element-of-surprise: I'm unclear on the instruction.  So the original code will return a TokenResponse{} regardless of getting an IDToken.  
Which should I do:
- Change CreateIDToken() to never error?
- Ignore the error here and put in a comment that says why (which I'm not sure, so you'd have to fill in)
- return this error, stop processing

internal/requests/username_password_request.go
element-of-surprise: ping @henrik-me

internal/requests/username_password_request.go
element-of-surprise: Left over debug log, removed.

internal/msalbase/authentication_result.go
element-of-surprise: @henrik-me ping

msal/client_application_parameters.go
element-of-surprise: What I meant (sorry, wasn't clear), was I moved the functionality of setADDAuthority into this func, as it was only used inside this function and was like a single call.

internal/msalbase/Account.go
bgavrilMS: Thanks @element-of-surprise . CCing @pmaytak who's looking at perf in MSAL .NET

We may want to invest into a similar approach for the other MSAL libraries. I fear however this is pretty complex, so I'm trying to better understand the return on investment here (outside maintainability). 

I do not know if gzip / deflate is supported by AAD services.

internal/msalbase/token_response.go
bgavrilMS: IdTokens aren't always returned, for example in service 2 service / client_credential grant where there is no user (and so no idToken). This is not an error. `log.Info` or `log.Verbose` is more suitable.

internal/msalbase/account_test.go
bgavrilMS: Makes sense, in MSAL.NET we used to do `Assert.Inconclusive("Test setup error ...")` but reverted to `Assert.Fail("Test setup error")` because the Inconclusive assert was too obscure. 

As long as the CI system interprets this as a failure and a developer looking at the test failures / logs easily understand what the problem is, it's good to go.

internal/msalbase/authentication_result.go
bgavrilMS: I think @henrik-me is saying that the error message is not actionable enough for a developer using MSAL.go
I do think you should necessarily try to improve error messages, because you may not have enough "MSAL context" to improve them. 

In this instance, it's possible (although highly unlikely) that a response has a "expires_in" key / pair that is invalid (missing or not a positive int). So maybe smth more in the lines of `The token response from the server is invalid because expires_in is missing / set to X`

internal/msalbase/authentication_result.go
bgavrilMS: This is fine, DeclinedScopes is part of the public API so we should not remove it. As long as developers understand which scopes were declined, it's fine by me.

For context - "Declined scopes" is a feature where MSAL libraries haven't fully aligned. Many MSAL libs **ignore** declined scopes. This is pretty bad, as it causes cache misses, but it's hard to change a default without incrementing major version lib, which is hard for a miriad of reasons (docs, samples, partners etc.). 
New libraries SHOULD fail when there are declined scopes, because the access token is not usable. The developer must be informed of which scopes were not issued.

internal/msalbase/token_response.go
bgavrilMS: You do not need the IdToken to create an `AuthenticationResult` but I think as long as there is a ClientInfo, you can rely on there being an IdToken as well. 

In fact, you should be able to create AuthenticationResult from **either** Access Token **or** Id Token, i.e. one of them can be missing. Here's the algorithm from MSAL.NET https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/01203edda53f4850b6a02f3569127f68086f5b8c/src/client/Microsoft.Identity.Client/AuthenticationResult.cs#L117

internal/tokencache/default_storage_manager.go
element-of-surprise: ping @bgavrilMS

internal/requests/username_password_request.go
bgavrilMS: Please delete the comment. 

Context: MSAL libraries have a fairly poor understanding of Kerberos / IntegratedWindowsAuth flow which is actually happening here. When there will be a better understanding, we will update all libs. AFAIK there is no indication that UI is required from this API.

internal/msalbase/authentication_result.go
element-of-surprise: I'm going to hold off on this for a big.  Many of these todos are notes for future edits.  This todo will happen in a future refactor.

internal/requests/username_password_request_test.go
```diff
@@ -8,57 +8,85 @@ import (
...
+		"https://login.microsoftonline.com/v2.0",
+		"login.microsoftonline.com",
+	)
+
+	usernamePassRequest = &UsernamePasswordRequest{
+		authParameters: uprTestAuthParams,
+	}
 }
 
+// TODO(msal expert): This test SEEMS borked. .Execute()
```
bgavrilMS, line 55: This flow should work, we can find a "federated" test user to try with. Maybe log an issue for this test to be fixed? @keegan-caruso - can you help with this?

internal/requests/auth_code_request.go
element-of-surprise: Yes, or at least for the moment.

The Execute() here modifies its internal state.  Most *Requests seem to do that.  We could probably still avoid a heap allocations by using pointer methods and not pointer objects.  But that only works if you don't have to use the object in an interface{} and the object isn't passed around in parents by pointer.  

I haven't analyzed the code that deep in order to see if it can be removed.

internal/requests/auth_code_request_test.go
element-of-surprise: That isn't the case with all mocks, but certainly the ones here.  The mocks should be replaced and the code should test the outcomes.   

However, this CL doesn't cover those types of changes, just changes enough to keep everything working.

internal/requests/device_code_response.go
element-of-surprise: Nope, changed.

internal/tokencache/cache_serialization_contract_test.go
element-of-surprise: Fixed

msal/acquire_token_silent_parameters.go
element-of-surprise: done

internal/msalbase/token_response.go
bgavrilMS: "secret" is how we store the actual token in the cache, I understood the meaning here.

internal/msalbase/token_response.go
bgavrilMS: Not saying you have to improve this area of MSAL.go - just giving some context. I'm aware that this PR is too large already, my goal is to bring MSAL context and help close on remaining issues.

msal/client_application_parameters.go
element-of-surprise: I did mean moved, as I moved the setADDAuthority code inside here.  Sorry for the confusion.

internal/msalbase/authentication_result.go
element-of-surprise: done.

internal/msalbase/auth_parameters_internal.go
element-of-surprise: Updated.  Going to hold renames for a future change.

internal/requests/username_password_request.go
element-of-surprise: Done.

internal/msalbase/authentication_result.go
element-of-surprise: Error updated.

internal/tokencache/default_storage_manager.go
bgavrilMS: The algorithm is a bit more complex than this, I assume there are some tests covering everything. I would keep the order as is. 

The algorithm is 

1. If application is NOT part of the family, search by client_ID
2. If app is part of the family or if we DO NOT KNOW if it's part of the family, search by family ID, then by client_id (we will know if an app is part of the family after the first token response). 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/311fe8b16e7c293462806f397e189a6aa1159769/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs#L95

internal/msalbase/authentication_result.go
element-of-surprise: @bgavrilMS 

The explanation makes sense.  

If DeclindedScopes will never be set for this type of result, it should be removed.  Its just cruft and I don't think consistency where two APIs work differently due to language, etc... is a good reason to keep it.  

I would like to remove this, objections?

internal/tokencache/default_storage_manager.go
element-of-surprise: @bgavrilMS 

There is a test, please verify it looks correct (as it currently passes) and advise if we should add another test:
internal/tokencache/default_storage_manager_test.go: TestDefaultStorageManagerReadRefreshToken

bgavrilMS: Wow, such a verbose test. I think this is fine. Closing.

element-of-surprise: Opened a bug for you to identify a test that we should add that will fail due to current method.  Will update the code after adding said test.

internal/requests/username_password_request_test.go
element-of-surprise: Going to let this finish offline, there are bugs to review all tests.  TODO will remain for the moment.

PR #84, opened 2020-11-17
Remove HTTPManager and dependent interfaces

jhendrixMSFT:
The indirection is unnecessary and can introduce future versioning
issues.  Instead, create http.Request where needed and send them using
the configured HTTPClient.
Flow context parameter to HTTP request creation.
Use http.Header and url.Values types instead of map[string]string.
Use query parameter encoding algorithm from the standard library.
Misc other bits of refactoring.


jhendrixMSFT:
Fixes https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/56

jhendrixMSFT:
One test failure that needs to be resolved.

msal/mock_http_manager.go
```diff
@@ -3,20 +3,20 @@
...
-func (mock *mockHTTPManager) Get(url string, requestHeaders map[string]string) (HTTPManagerResponse, error) {
-	args := mock.Called(url, requestHeaders)
-	return args.Get(0).(HTTPManagerResponse), args.Error(1)
-}
-
-//Post mocks the Post method of a HTTPManager
-func (mock *mockHTTPManager) Post(url string, body string, requestHeaders map[string]string) (HTTPManagerResponse, error) {
-	args := mock.Called(url, body, requestHeaders)
-	return args.Get(0).(HTTPManagerResponse), args.Error(1)
+func (m *mockHTTPManager) Do(req *http.Request) (*http.Response, error) {
```
sangonzal, line 16: Can this file be removed given that `http_manager` is being removed?

sangonzal (approved):
Looks good to me! I think the new http code looks a lot cleaner. There are some future features (like telemetry and logging) where having a single point that all http requests and responses go through (formerly msal_http_manager.go) is useful, but that can be done later on if it's needed.

msal/mock_http_manager.go
jhendrixMSFT: Not at present as this is still used as part of the mock testing.

internal/requests/device_code_response.go
```diff
@@ -28,14 +30,19 @@ type DeviceCodeResponse struct {
 
 // CreateDeviceCodeResponse creates a deviceCodeResponse instance from HTTP response.
 // TODO(jdoak): look at responseData and see if we can get it originally from []byte instead.
```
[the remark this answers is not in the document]
jhendrixMSFT: Good catch, fixed.

keegan-caruso (approved):
LGTM

internal/msalbase/oauth_response_base_test.go
```diff
@@ -9,20 +9,20 @@ import (
...
 func TestCreateOAuthResponseBase(t *testing.T) {
-	_, err := CreateOAuthResponseBase(404, oauthResponse)
+	_, err := CreateOAuthResponseBase(404, []byte(oauthResponse))
 	actualError := errors.New("HTTP 404")
 	if err.Error() != actualError.Error() {
 		t.Errorf("Actual error %v differs from expected error %v", err, actualError)
 	}
-	expectedErr := errors.New("invalid request")
-	_, err = CreateOAuthResponseBase(300, oauthResponseWithError)
+	expectedErr := errors.New("invalid request: missing payload content")
```
element-of-surprise, line 20: These expectedErr are brittle.  Any change to the text of the error can cause the tests to fail.

I know that wasn't your test, but can we change this so we just test the given inputs against outputs?  The output can be simply tested to see if it has an error, not a specific error.  This is sufficient 99.9% of the time.

internal/requests/auth_code_request.go
```diff
@@ -38,28 +40,28 @@ func CreateAuthCodeRequest(webRequestManager WebRequestManager, authParameters m
...
 	resolutionManager := CreateAuthorityEndpointResolutionManager(req.webRequestManager)
-	endpoints, err := resolutionManager.ResolveEndpoints(req.authParameters.AuthorityInfo, "")
+	endpoints, err := resolutionManager.ResolveEndpoints(ctx, req.authParameters.AuthorityInfo, "")
 	if err != nil {
 		return msalbase.TokenResponse{}, fmt.Errorf("unable to resolve endpoints: %w", err)
 	}
 
 	req.authParameters.Endpoints = endpoints
-	params := make(map[string]string)
+	params := url.Values{}
```
element-of-surprise, line 51: Thank you, this has been tickling my ocd.

internal/wstrust/wstrust_mex_document.go
```diff
@@ -49,9 +51,14 @@ func updateEndpoint(cached *Endpoint, found Endpoint) bool {
 
 // TODO(jdoak): Refactor into smaller bits
 
-func CreateWsTrustMexDocument(responseData string) (MexDocument, error) {
+func CreateWsTrustMexDocument(resp *http.Response) (MexDocument, error) {
 	definitions := &definitions{}
```
element-of-surprise, line 55: Nit: I'd move definitions down to right above line 61 where it is first used.

element-of-surprise (changes requested):

internal/msalbase/oauth_response_base_test.go
jhendrixMSFT: Agreed, fixed.

internal/wstrust/wstrust_mex_document.go
jhendrixMSFT: Fixed.

PR #86, opened 2020-11-20
Adding additional fields in struct client info

abhidnya13:
When adding e2e test automation, found out that the changes introduced  in [PR 77](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/77) broke some flows. 

Unmarshalling of ClientInfo gave an error as per [this](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/internal/json/design.md) doc in here : https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/internal/json/struct.go#L27

Adding fix for this.

### DIFF
--- a/internal/msalbase/token_response.go
+++ b/internal/msalbase/token_response.go
@@ -30,6 +30,8 @@ type tokenResponseJSONPayload struct {
 type ClientInfoJSONPayload struct {
 	UID  string `json:"uid"`
 	Utid string `json:"utid"`
+
+	AdditionalFields map[string]interface{}
 }
 
 // TokenResponse is the information that is returned from a token endpoint during a token acquisition flow.


element-of-surprise (approved):

internal/msalbase/token_response.go
```diff
@@ -30,6 +30,8 @@ type tokenResponseJSONPayload struct {
 type ClientInfoJSONPayload struct {
 	UID  string `json:"uid"`
 	Utid string `json:"utid"`
+
+	AdditionalFields map[string]interface{}
```
keegan-caruso, line 34: Can we get a test?

sangonzal (approved):
:shipit:

internal/msalbase/token_response.go
element-of-surprise: @keegan-caruso 

I was thinking about eventually removing the existing tests serialization tests.  I left the existing ones in to validate that the encoder/decoder was doing what it should outside its internal tests when I switched it out.

What I wanted to have before that was an etoe that proved that these worked.  The main bug here looks like that this was never used in any tests and of course the first time it was used, it croaked because the new library requires all structs to have AdditionalFields.  

Do you think we need these serialization/deserialization once we are actually testing the flows?  It provides extra safety, but I've always felt these types of tests were excessive and are usually a sign that the etoe and unit tests are weak.

keegan-caruso: It sounds wrong to have code here that is not exercised at all in UT. Agreed that e2e should catch it, but I'd prefer to have quicker feedback loops. e2e usually doesn't get run often enough.

That might be part of a larger discussion scope than these changes though.

keegan-caruso (approved):

internal/msalbase/token_response.go
element-of-surprise: I agree that there should be a unit test of whatever is using this message, which is not happening here.  But I want to separate the capability unit test that doesn't exist (something that uses this message) from a test for just encoding/decoding.  

Specifically,  do we need encoding/decoding tests vs just usage tests and etoe coverage?  We can talk this talk offline.

PR #87, opened 2020-11-20
Add race detector to build and test 

keegan-caruso:
See: https://golang.org/doc/articles/race_detector.html

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -34,10 +34,10 @@ jobs:
         run: if [ $(gofmt -l -s . | wc -l) -ne 0 ]; then echo "fmt failed"; exit 1; fi
 
       - name: Build
-        run: go build -v ./msal
+        run: go build -race -v ./msal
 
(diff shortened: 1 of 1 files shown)


sangonzal (approved):
:shipit:

.github/workflows/go.yml
```diff
@@ -34,10 +34,10 @@ jobs:
         run: if [ $(gofmt -l -s . | wc -l) -ne 0 ]; then echo "fmt failed"; exit 1; fi
 
       - name: Build
-        run: go build -v ./msal
+        run: go build -race -v ./msal
```
element-of-surprise, line 37: All the updates look good.  But I'm wondering why we are running with -v?  In the case of test failure it dumps info, but -v just dumps pages of info we don't care about.  Both in builds and tests.

keegan-caruso: -v was from the initial commit of the pipeline file, I'm fine dropping it

PR #88, opened 2020-11-20
add golangci-lint action

keegan-caruso:


keegan-caruso:
Optional linters: https://golangci-lint.run/usage/linters

I'm looking for feedback on others to enable

element-of-surprise:
https://github.com/securego/gosec has some interesting checks that might be worth adding.
gofmt - but I'm pretty sure sonar is doing that

I didn't find any of the others compelling enough to add.

sangonzal (approved):

PR #92, opened 2020-12-01
Adds storage package, deletes tokencache

element-of-surprise:
This PR accomplishes the following:

- Combine StorageManager and CacheManager
- Modify tests for new paradigm
- Remove mock storage manager usage for real storage manager
- Fix tests that only worked because of bad mocks
- Reduce number of files in package to more reasonable amount
- Move to storage package instead of TokenCache


internal/storage/storage.go
```diff
@@ -0,0 +1,427 @@
...
+	"sync"
+	"time"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/internal/json"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/internal/msalbase"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/internal/requests"
+)
+
+// TODO(someone): This thing does not expire tokens and DeleteCachedRefreshToken
+// is not implemented (or used anywhere).  Seems bad.
```
sangonzal, line 24: It does not look like RemoveAccounts has been implemented. The token cache should have at least two public APIs: 
- For seeing what accounts and associated tokens are in the cache (`GetAllAccounts() (Account, err)`) 
- For removing accounts and associated tokens (`RemoveAccount(Account account)`) 

The implementation for RemoveAccount seems to be DeleteAccount, but I think this needs to be updated to:
- Use RemoveAccount to be consistent with other libraries
- Update paramaters to get rid of envAliases[]. User should not have to pass this in. Might also consider if we should use an Account object instead of homeAccountId. 
- RemoveAccount should remove the account and any associated accessTokens, refreshTokens, and IdTokens. In it's current implementation, I think it's only removing accounts.
- Expose RemoveAccount on clientApplication. 

I opened up an issue to track this work. In the meantime, I think DeleteCacheRefreshToken can probably be removed. #93

sangonzal (approved):
Looks good to me, just need licenses on files.

keegan-caruso (approved):

PR #96, opened 2020-12-06
Support for setting redirect-url in token-fetch portion of confidential-app authentication code flow

harikb:
Changes
1. Correcting some broken links in README
2. Support for adding RedirectURI in token-fetch portion of confidential-app authentication code flow. 

See also this comment elsewhere for context https://github.com/Azure/azure-sdk-for-go/pull/13565#issuecomment-739461185

The change retains the ability to have a default RedirectURI while still allowing the caller to set one if available.
@jhendrixMSFT

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ The MSAL library for Go is part of the [Microsoft identity platform for develope
 
 Quick links:
 
-| [Getting Started](https://docs.microsoft.com/azure/active-directory/develop/#quickstarts) | [GoDoc](https://godoc.org/github.com/AzureAD/microsoft-authentication-library-for-go/msal) | [Wiki](https://github.com/AzureAD/microsoft-authentication-library-for-go/wiki) | [Samples](https://github.com/AzureAD/microsoft-authentication-library-for-go/tree/dev/examples) | [Support](README.md#community-help-and-support) |
+| [Getting Started](https://docs.microsoft.com/azure/active-directory/develop/#quickstarts) | [GoDoc](https://godoc.org/github.com/AzureAD/microsoft-authentication-library-for-go/msal) | [Wiki](https://github.com/AzureAD/microsoft-authentication-library-for-go/wiki) | [Samples](https://github.com/AzureAD/microsoft-authentication-library-for-go/tree/test/devapps) | [Support](README.md#community-help-and-support) |
 | ------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------------ | 
 
 ## Build Status
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Quick links:
 To install Go, visit [this link](https://golang.org/dl/).
 
 ### Installing MSAL Go
-`go get -u github.com/AzureAD/microsoft-authentication-library-for-go/`
+`go get -u github.com/AzureAD/microsoft-authentication-library-for-go/msal`
 
 ## Usage
 Before using MSAL Go, you will need to [register your application with the Microsoft identity platform](https://docs.microsoft.com/azure/active-directory/develop/quickstart-v2-register-an-app).
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Quick links:
 To install Go, visit [this link](https://golang.org/dl/).
 
 ### Installing MSAL Go
-`go get -u github.com/AzureAD/microsoft-authentication-library-for-go/msal`
+`go get github.com/AzureAD/microsoft-authentication-library-for-go/msal`
 
 ## Usage
 Before using MSAL Go, you will need to [register your application with the Microsoft identity platform](https://docs.microsoft.com/azure/active-directory/develop/quickstart-v2-register-an-app).
--- a/msal/acquire_token_auth_code_parameters.go
+++ b/msal/acquire_token_auth_code_parameters.go
@@ -18,6 +18,7 @@ type acquireTokenAuthCodeParameters struct {
 	CodeChallenge    string
 	clientCredential msalbase.ClientCredential
 	requestType      requests.AuthCodeRequestType
+	RedirectURI      string
 }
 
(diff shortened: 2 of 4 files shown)


jennyf19 (approved):

msal/public_client_application.go
```diff
@@ -129,4 +129,5 @@ type AcquireTokenByDeviceCodeOptions struct {
 type AcquireTokenByAuthCodeOptions struct {
 	Code          string
 	CodeChallenge string
+	RedirectURI   string
```
jhendrixMSFT, line 132: @jmprieur can you please comment on this?  I explicitly removed this in https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/68 per your request.

README.md
```diff
@@ -19,7 +19,7 @@ Quick links:
 To install Go, visit [this link](https://golang.org/dl/).
 
 ### Installing MSAL Go
-`go get -u github.com/AzureAD/microsoft-authentication-library-for-go/`
+`go get -u github.com/AzureAD/microsoft-authentication-library-for-go/msal`
```
jhendrixMSFT, line 22: Since you're updating this line can you please remove the `-u`?  This will trigger upgrading your dependencies in `go.mod` which is likely wrong (we had a customer complain about this in our Event Hubs repo).

jhendrixMSFT (changes requested):
Get confirmation that adding the `redirectURI` back is ok.

msal/public_client_application.go
harikb: I am not the original requester of that change, although our handles look similar. In my case, I was trying to use a token-fetch as an explicit call back to MS instead of implicit-token flow. As per the documentation, implicit token on auth-code flow is less secure and they recommend webapps fetch it back from MS. When we call back, this redirectURI is required, and matching original auth request.

README.md
harikb: Yes I will make that change

msal/public_client_application.go
jhendrixMSFT: The recommendation is for apps to use the `https://login.microsoftonline.com/common/oauth2/nativeclient` reply URL, this is what I did when testing the change and it worked for me.  See the conversation [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/68#discussion_r516579950).  I'm far from an expert here though so somebody from the MSAL team needs to comment further.

MSAL for other languages don't expose a redirect URI parameter.

harikb: @jmprieur I don't quite understand this statement that you made in #68 
> but MSAL stops the redirect

My use-case is a webapp that needs to redirect a user to MS auth servers and then back to the site. It would be what you consider a confidential-app and I am specifically choosing not to use 'implicit-token' checkbox in the App settings since that is not recommended.

After the user accepts the consent, the browser needs to be redirected back to the site, so I do need to supply a real redirect-url in addition to make it match with one of the configured urls on the app settings.  If you are saying MS will redirect back to my real site even if I specify `login.microsoftonline.com`, then I am mistaken, I have not tried that.

Now the specific fix is about making it configurable, so that I can pass the same value back when I fetch the token.
My app is not opensource, but it is very similar to your sample

First place I need to use the redirect-url
https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/test/devapps/confidential_auth_code_sample.go#L29

Second place I need to use the redirect-url (this fix)
https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/test/devapps/confidential_auth_code_sample.go#L61

jhendrixMSFT:
@harikb I've asked an MSAL expert to review this change.

jhendrixMSFT (approved):

msal/public_client_application.go
jhendrixMSFT: After discussion we do need to add this back.  However we shouldn't place this in the options struct as it makes it appear optional when it should be user-specified (falling back to a default value is incorrect).  Can you please move this out of the options struct and make it a parameter so it's clear the caller must supply a value?

jhendrixMSFT (changes requested):
Make RedirectURI a param

sangonzal (commented):
Overall looks good, thanks for contributing. I've changed the location of  `AcquireTokenByAuthCodeOptions` in #114 so it's clear that its shared in between public clients and confidential clients. Once the change when the RedirectUri == "" is made, it should be ready to be merged.

msal/acquire_token_auth_code_parameters.go
```diff
@@ -30,6 +31,11 @@ func createAcquireTokenAuthCodeParameters(scopes []string) *acquireTokenAuthCode
 
 func (p *acquireTokenAuthCodeParameters) augmentAuthenticationParameters(authParams *msalbase.AuthParametersInternal) {
 	p.commonParameters.augmentAuthenticationParameters(authParams)
-	authParams.Redirecturi = "https://login.microsoftonline.com/common/oauth2/nativeclient"
+	if p.RedirectURI != "" {
+		authParams.Redirecturi = p.RedirectURI
+	}
+	if authParams.Redirecturi == "" { // set it to default if it is still not set
+		authParams.Redirecturi = "https://login.microsoftonline.com/common/oauth2/nativeclient"
```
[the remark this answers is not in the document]
jhendrixMSFT: I propose making this a parameter so it's clear the caller must supply a value.  What do you think?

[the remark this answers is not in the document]
sangonzal: Oh yes good point. So instead of going in the options struct, it should be in the parameters for both [public client auth code](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/msal/public_client_application.go#L101) and [confidential client auth code](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/msal/confidential_client_application.go#L100)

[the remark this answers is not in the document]
harikb: If I make RedirectURI required/argument, I need to figure out what to in tests like these https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/msal/public_client_application_test.go#L80

I will work on it, but it will take a bit of time. I don't yet know how those mocks work. Thanks

jhendrixMSFT:
Superseded by https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/188

Issue #48, opened 2020-10-14
Update client credential grant to make do a cache lookup by default


sangonzal:
Client credential grant will only look up tokens in the cache if used with acquire token silent. It should instead do a default cache lookup, with the option to skip.

bgavrilMS:
I know MSAL Py has this approach while MSAL .NET does not. So this is not a strong requirement for GA. It can also be added later.

PR #97, opened 2020-12-07
Integration Tests

abhidnya13:
The integration tests are failing on the ADFS v4 test case. All other federated scenarios seem to be working fine, investigating on what is the issue.

Update: After investigation, this seems a bug, opened an issue to track this [here ](https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/98)

PR details: This adds e2e tests executing the Username password flow. It involves getting the user and application configuration from LabAPI and using this to run the e2e flow. Pointers to Lab API for reference : https://docs.msidlab.com/labapi/intro.html

The build fails because this includes a test which fails because of the bug mentioned above.


test/integrationtests/lab_interface.go
```diff
@@ -0,0 +1,108 @@
...
+	responseBody, err := sendRequestToLab("https://msidlab.com/api/user", query, l.getLabAccessToken())
+	if err != nil {
+		log.Fatal(err)
+	}
+	var users []user
+	err = json.Unmarshal(responseBody, &users)
+	if err != nil {
+		log.Fatal(err)
+	}
+	user := users[0]
```
sangonzal, line 84: nit: could maybe check that users[0] is not nil or empty, and if not return an useful error message incase that it is either.

test/integrationtests/lab_interface.go
```diff
@@ -0,0 +1,108 @@
...
+			log.Fatal(err)
+		}
+
+	}
+	scopes := []string{"https://msidlab.com/.default"}
+	result, err := l.labApplication.AcquireTokenByClientCredential(context.Background(), scopes)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return result.GetAccessToken()
```
sangonzal, line 71: This section gets called for every test, which means we make a network call each time to get a new token, since #48 has not been implemented. We should consider using acquire token silent or caching the token to avoid making that call every time.

test/integrationtests/http_client_helper.go
```diff
@@ -0,0 +1,32 @@
+package integrationtests
```
sangonzal, line 1: License missing on a couple of files

test/integrationtests/username_password_test.go
```diff
@@ -0,0 +1,56 @@
...
+	)
+	if err != nil {
+		t.Error(err)
+	}
+	if result.AccessToken == "" {
+		t.Error("No access token found")
+	}
+	if result.Account.GetUsername() != user.Upn {
+		t.Errorf("Incorrect user account")
+	}
```
sangonzal, line 36: Do we want to add checks for others objects in authentication result, like id token?

sangonzal (commented):
Overall looks good to me.

test/integrationtests/username_password_test.go
```diff
@@ -0,0 +1,56 @@
...
+}
+func TestAcquireTokenByUsernamePasswordManaged(t *testing.T) {
```
element-of-surprise, line 38: These 4 tests should be one table driven test.  If you haven't seen a table driven tests, if you do a google search for "golang table driven test" several articles should come up

test/integrationtests/test_constants.go
```diff
@@ -0,0 +1,18 @@
+package integrationtests
+
+// Scopes.
+const (
+	MsIDlabDefaultScope = "https://msidlab.com/.default"
```
element-of-surprise, line 5: There doesn't look to be a good reason to have a file with just constants in it.  I'd simply create a "integrations.go" file and put it in there along with all your tests from user_password_test.go until the file gets somewhere between 500 and 1000 lines before I'd break it up.

Also, may of these constants are exported(public), they should probably be private.

test/integrationtests/username_password_test.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/msal"
+)
+
+var labClientInstance = &labClient{}
+
+func AcquireTokenByUsernamePasswordCommonAAD(t *testing.T, user user, authority string, scope []string, appID string) {
```
element-of-surprise, line 15: See my table driven comment below, but this should probably just be in a table driven test as the test code in the test loop instead of a separate function.

test/integrationtests/username_password_test.go
```diff
@@ -0,0 +1,56 @@
...
+package integrationtests
+
+import (
+	"context"
+	"testing"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/msal"
+)
+
+var labClientInstance = &labClient{}
```
element-of-surprise, line 13: Not a fan of the globals for tests.  They usually end up with weird side effects later.  I'm recommending a single test below that is table drive, this can move inside that test.

element-of-surprise (changes requested):
Overall a pretty good start.  I like the breakdown of the constants into logical groups and the helper function.  

This seems to borrow the idea of breaking the package into a lot of files.  At this point, I'd probably make this a single file until we reach somewhere between 500 and 1000 lines.   

Top load the helper and the constants and the tests below that.  Other comments in line.  Any questions, feel free to write out to me on teams at any time.

test/integrationtests/http_client_helper.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"io/ioutil"
+	"log"
+	"net/http"
+)
+
+var defaultClient = &http.Client{}
```
keegan-caruso, line 9: do we need this or can we use the http.DefaultClient ?

test/integrationtests/http_client_helper.go
```diff
@@ -0,0 +1,32 @@
...
+
+func sendRequestToLab(url string, query map[string]string, accessToken string) ([]byte, error) {
```
keegan-caruso, line 11: this name isn't quite right. The implementation isn't about sending a request to the lab. It is creating an arbitrary request and returning the response body.

abhidnya13:
@element-of-surprise  Tried arranging the tests in a table driven tests format ! Let me know your feedback if anything can be improved there !

test/integrationtests/lab_interface.go
abhidnya13: I have made the change to do an acquireTokenSilent before the acquireToken method, but looks like some stuff there is broken, so acquire token silent really does not return back a token even when there is one in the cache, I'll create an issue to track this.

test/integrationtests/username_password_test.go
abhidnya13: Thats a good suggestion ! I see though that the idToken field is not exposed in the authentication result, will create an issue there as well to track this, once that is resolved, ill add a check here as well.

test/integrationtests/http_client_helper.go
abhidnya13: You are right ! We dont need it, removed it [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/97/commits/4b66c329c1a307d02c490ed9a6539e96b94e49e5)

test/integrationtests/http_client_helper.go
abhidnya13: Tried changing the name to reflect what it does, let me know your thoughts and suggestions if this doesn't seem appropriate

sangonzal (approved):
:shipit:

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+				tc.user.Password,
+			)
+			if err != nil {
+				t.Error(err)
+			}
+			if result.AccessToken == "" {
+				t.Error("No access token found")
+			}
+			if result.Account.GetUsername() != tc.user.Upn {
+				t.Errorf("Incorrect user account")
```
keegan-caruso, line 204: nit: don't need to use Errorf if there is no format string

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	HomeDomain       string `json:"homeDomain"`
+	HomeUPN          string `json:"homeUPN"`
+	B2cProvider      string `json:"b2cProvider"`
+	LabName          string `json:"labName"`
+	LastUpdatedBy    string `json:"lastUpdatedBy"`
+	LastUpdatedDate  string `json:"lastUpdatedDate"`
+	Password         string
+}
+type secret struct {
+	Secret string `json:"value"`
```
keegan-caruso, line 68: this "stutters" on use e.g. secret.Secret. Maybe call the field value? so use looks like `secret.Value`. a struct that represents a secret that holds the value

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		panic("Failed to get input user" + err.Error())
+	}
+	return testUser
+}
+
+func TestUsernamePassword(t *testing.T) {
+	labClientInstance, err := newLabClient()
+	if err != nil {
+		panic("Failed to get a lab client" + err.Error())
```
keegan-caruso, line 173: nit: here and 165, spacing is going to make it hard to read the error. e.g. "Failed to get a lab client. " + err.Error() would be an improvement.

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return body, nil
+}
+
+// Lab Interface
+
+type labClient struct {
+	labApplication *msal.ConfidentialClientApplication
```
keegan-caruso, line 45: shorten to just "App" or "Application"

test/integrationtests/lab_interface.go
abhidnya13: Tracked issue : https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/106

test/integrationtests/username_password_test.go
abhidnya13: Issue to track this : https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/107

keegan-caruso (approved):

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	"errors"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/msal"
+)
+
+// HTTP Helper
```
element-of-surprise, line 18: This is not a valid Go comment.  Also doesn't help, just remove

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	defer response.Body.Close()
+	body, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		return nil, err
+	}
+	return body, nil
+}
+
+// Lab Interface
```
element-of-surprise, line 42: same

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	Mfa              string `json:"mfa"`
+	ProtectionPolicy string `json:"protectionPolicy"`
+	HomeDomain       string `json:"homeDomain"`
+	HomeUPN          string `json:"homeUPN"`
+	B2cProvider      string `json:"b2cProvider"`
+	LabName          string `json:"labName"`
+	LastUpdatedBy    string `json:"lastUpdatedBy"`
+	LastUpdatedDate  string `json:"lastUpdatedDate"`
+	Password         string
+}
```
element-of-surprise, line 66: nit: add blank line between

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	err = json.Unmarshal(responseBody, &secret)
+	if err != nil {
+		return "", err
+	}
+	return secret.Value, nil
+}
+
+// TODO: Add getApp() when needed
+
+// TestConstants
```
element-of-surprise, line 142: Same and all below.  These don't provide value.

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+// Scopes.
+const (
```
element-of-surprise, line 145: Move all const to the top of the file

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	return testUser
+}
+
+func TestUsernamePassword(t *testing.T) {
+	labClientInstance, err := newLabClient()
+	if err != nil {
+		panic("Failed to get a lab client. " + err.Error())
+	}
+	tests := map[string]struct {
```
element-of-surprise, line 175: This is must improved. But I'm not a fan of the map[string]struct{}, mainly because I like things to execute in a consistent order over time.  This will execute in random order.  Using []struct is more standard.

```go
tests := []struct{
  desc string
  user user
}{
  {"Managed", getTestUser(labClientInstance, map[string]string{"usertype": "cloud"})},
  ...
}
```

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+		user user
+	}{
+		"Managed": {user: getTestUser(labClientInstance, map[string]string{"usertype": "cloud"})},
+		"ADFSv2":  {user: getTestUser(labClientInstance, map[string]string{"usertype": "federated", "federationProvider": "ADFSv2"})},
+		"ADFSv3":  {user: getTestUser(labClientInstance, map[string]string{"usertype": "federated", "federationProvider": "ADFSv3"})},
+		//"ADFSv4":  {user: getTestUser(labClientInstance, map[string]string{"usertype": "federated", "federationProvider": "ADFSv2"})},
+	}
+	for name, tc := range tests {
+		t.Run(name, func(t *testing.T) {
```
element-of-surprise, line 184: You can do a sub-test like this, though I don't know that it buys you anything.  I would just remove the sub-test and just build the "desc" field above into my error messages.

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+				panic(err)
+			}
+			result, err := publicClientApp.AcquireTokenByUsernamePassword(
+				context.Background(),
+				[]string{graphDefaultScope},
+				tc.user.Upn,
+				tc.user.Password,
+			)
+			if err != nil {
+				t.Error(err)
```
element-of-surprise, line 198: This should be t.Fatalf(), because you don't want to proceed after this.  

This is where I usually do my desc field instead of doing sub tests:

```go
t.Fatalf("TestUsernamePassword(%s): on AcquireTokenByUsernamePassword(): got err == %s, want err == nil", test.desc, err)
```

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+			}
+			if result.AccessToken == "" {
+				t.Error("No access token found")
```
element-of-surprise, line 201: In the same vein, this is what I'd do here.
```go
t.Fatalf("TestUsernamePassword(%s): got AccessToken == '', want AccessToken == non-empty string", test.desc)
```

test/integration/integration_test.go
```diff
@@ -0,0 +1,208 @@
...
+	"io/ioutil"
+	"net/http"
+	"os"
+	"testing"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/msal"
+)
+
+// HTTP Helper
+func httpRequest(url string, query map[string]string, accessToken string) ([]byte, error) {
```
element-of-surprise, line 19: The URL package includes a value used for queries: https://golang.org/pkg/net/url/#Values

If you use it instead, you can simply do:
request.URL.RawQuery = query.Encode()

element-of-surprise (changes requested):

test/integration/integration_test.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		result, err = l.app.AcquireTokenByClientCredential(context.Background(), scopes)
+		if err != nil {
+			return "", err
+		}
+	}
+	return result.GetAccessToken(), nil
+}
+
+func (l *labClient) getUser(query map[string]string) (user, error) {
```
element-of-surprise, line 106: Make this a url.Values too.

test/integration/integration_test.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	user := users[0]
+	user.Password, err = l.getSecret(map[string]string{"Secret": user.LabName})
+	if err != nil {
+		return user, err
+	}
+	return user, nil
+}
+
+func (l *labClient) getSecret(query map[string]string) (string, error) {
```
element-of-surprise, line 135: Same

test/integration/integration_test.go
```diff
@@ -0,0 +1,203 @@
...
+		desc string
+		user user
+	}{
+		{"Managed", getTestUser(labClientInstance, map[string]string{"usertype": "cloud"})},
+		{"ADFSv2", getTestUser(labClientInstance, map[string]string{"usertype": "federated", "federationProvider": "ADFSv2"})},
+		{"ADFSv3", getTestUser(labClientInstance, map[string]string{"usertype": "federated", "federationProvider": "ADFSv3"})},
+		{"ADFSv4", getTestUser(labClientInstance, map[string]string{"usertype": "federated", "federationProvider": "ADFSv4"})},
+	}
+	for _, test := range tests {
+		options := msal.DefaultPublicClientApplicationOptions()
```
element-of-surprise, line 181: Sorry, just now noticed this:

From line 181 to line 186: You can move this outside the test loop.  This is the same client and option with no internal state changes, so we only need to create it once.

element-of-surprise (approved):
Make these changes and you are home free.

test/integration/integration_test.go
abhidnya13: Not all of 181-186 could be put outside because the appId used to initialize the public client depends on the testcase, but I have moved the common part out

PR #102, opened 2020-12-08
Move CacheAccessor out of msal and fix up

element-of-surprise:
- Move CacheAccessor from msal/ to msal/cache
- msal.CacheAccessor -> cache.Token
- CacheAccssor.Before/After to IntoCache/ExportCache()
- storage.Manager.Serialize/Deserialize() to Marshal/Unmarshal()

This does a few things to help us out.  It moves this external interface into its own package that defines its scope. We move to more common naming, Marshal/Unmarshal().  The CacheAccessor(now cache.Token) has simplified methods that are clear using interfaces that give a single method call.  We also don't waste doing string/[]byte conversions that were not needed.  As a side effect, also shortens some test names.


msal/client_application.go
```diff
@@ -11,16 +11,17 @@ import (
...
 
 type noopCacheAccessor struct{}
 
-func (n noopCacheAccessor) BeforeCacheAccess(manager *storage.Manager) {}
-func (n noopCacheAccessor) AfterCacheAccess(manager *storage.Manager)  {}
+func (n noopCacheAccessor) IntoCache(cache cache.Unmarshaler) {}
+func (n noopCacheAccessor) ExportCache(cache cache.Marshaler) {}
 
-// cache provides an internal cache. It is defined to allow faking the cache in tests.
+// managet provides an internal cache. It is defined to allow faking the cache in tests.
```
keegan-caruso, line 22: nit: typo "manager"

internal/storage/storage.go
```diff
@@ -26,7 +26,7 @@ import (
 // TODO(someone): This thing does not expire tokens.
 
 // Manager is an in-memory cache of access tokens, accounts and meta data. This data is
-// updated on read/write calls. Deserialize() replaces all data stored here with whatever
+// updated on read/write calls. Unmarshal() replaces all data stored here with whatever
 // was given to it on each call.
 type Manager struct {
 	// TODO(jdoak): Going to refactor this (next PR).
```
keegan-caruso, line 32: is this up to date?

internal/storage/storage.go
```diff
@@ -384,23 +384,23 @@ func (m *Manager) update(cache *Contract) {
 
 // TODO(jdoak): Change this to return []byte, not string.
```
keegan-caruso, line 385: this looks resolved

internal/storage/storage.go
```diff
@@ -384,23 +384,23 @@ func (m *Manager) update(cache *Contract) {
...
 
-func (m *Manager) Serialize() (string, error) {
-	cache := m.Contract()
-
-	serializedCache, err := json.Marshal(cache)
+// Marshal implements cache.Marshaler.
+func (m *Manager) Marshal() ([]byte, error) {
+	b, err := json.Marshal(m.Contract())
```
jhendrixMSFT, line 389: Can we not simply `return json.Marshal(m.Contract())` here?

msal/client_application.go
```diff
@@ -56,12 +57,12 @@ func (client *clientApplication) acquireTokenSilent(ctx context.Context, silent
 	silent.augmentAuthenticationParameters(&authParams)
 
 	// TODO(jdoak): Think about removing this after refactor.
-	if sm, ok := client.cache.(*storage.Manager); ok {
-		client.cacheAccessor.BeforeCacheAccess(sm)
-		defer client.cacheAccessor.AfterCacheAccess(sm)
+	if sm, ok := client.manager.(*storage.Manager); ok {
```
keegan-caruso, line 60: Is there a reason why we don't want to call these on the fake?

msal/cache/cache.go
```diff
@@ -0,0 +1,32 @@
...
+// Unmarshaler unmarshals data from a storage medium into the internal cache, overwriting it.
+type Unmarshaler interface {
+	Unmarshal([]byte) error
+}
+
+// Token is used to provide external storage of token data. The data being passed is considered
+// opaque.
+type Token interface {
+	// IntoCache reads the cache in external storage into the internal cache, replacing it.
+	IntoCache(cache Unmarshaler)
```
keegan-caruso, line 28: I find this name to be confusing

jhendrixMSFT: Agreed.  Also see usage in `confidential_client_application.go`

internal/storage/storage.go
```diff
@@ -4,7 +4,7 @@
 // Package storage holds all cached token information for MSAL. This storage can be
 // augmented with third-party extensions to provide persistent storage. In that case,
 // reads and writes in upper packages will call Serialize() to take the entire in-memory
```
keegan-caruso, line 6: "Serialize()" => "Marshall()"

test/devapps/sample_cache_accessor.go
```diff
@@ -26,18 +26,18 @@ func (accessor *SampleCacheAccessor) BeforeCacheAccess(cache *storage.Manager) {
...
 }
 
-func (accessor *SampleCacheAccessor) AfterCacheAccess(cache *storage.Manager) {
-	data, err := cache.Serialize()
+func (accessor *SampleCacheAccessor) ExportCache(cache cache.Marshaler) {
+	data, err := cache.Marshal()
 	if err != nil {
 		log.Fatal(err)
 	}
-	err = ioutil.WriteFile(accessor.file, []byte(data), 0644)
```
keegan-caruso, line 40: we should fix this or suppress it

internal/storage/storage.go
element-of-surprise: It is, comment removed.

internal/storage/storage.go
element-of-surprise: We can since we did the conversion from string returns to bytes, good catch.

msal/cache/cache.go
element-of-surprise: Roger: would you have suggestions that you think are appropriate?  My mind is a little blank at the moment (I'll think about alternatives in the morning).

test/devapps/sample_cache_accessor.go
element-of-surprise: I'm assuming you mean the log.Fatal().  I changed these to log.Error.  In reality, there is not much to be done when these happen other than ignore them or log them.  Really up to the developer on what they want to do.

msal/client_application.go
element-of-surprise: Skipping the first poinrt and straight to your second point, now that we have Unmarshaller/Marshaller, we could make this more grandular.  Introduced a Serializer interface, it now tests for that and if it is implemented it uses it.

test/devapps/sample_cache_accessor.go
keegan-caruso: Sorry, missing context. gosec linter is failing. If you want to suppress it golangci-lint suppresses with a syntax of

// nolint:gosec

directly above the offending line.

otherwise permissions need to be at 600 or less

element-of-surprise:
Its old code, but it flagged it because I removed the string() conversion.  I changed it to 600, no reason not too I can see.

John Doak
jdoak@microsoft.com

msal/cache/cache.go
```diff
@@ -0,0 +1,38 @@
...
+
+// Serializer can seralize the cache to binary or from binary into the cache.
+type Serializer interface {
+	Marshaler
+	Unmarshaler
+}
+
+// Token is used to provide external storage of token data. The data being passed is considered
+// opaque.
+type Token interface {
```
sangonzal, line 32: Token also doesn't seem quite right here. Token, in the context of MSAL has a clear meaning - A JWT with claims used for authentication/authorization. It also doesn't have a "er" suffix, which I I think is Go idiom and which is used elsewhere. 

This type is something completely unrelated to the above definition - is allows a user to set callbacks for marshalling/unmarshalling of the in-memory token cache. I can't think of a better name myself, but if I was to choose the previous name "accessor" seems more appropriate

msal/cache/cache.go
sangonzal: Some ideas: 
- ImportCache, ExportCache
- ReadCache, WriteCache

msal/cache/cache.go
element-of-surprise: As to the Go idiom: "er" is only added to interface types that have a single method defined.  So instead of having Read.Read(), you have Reader.Read().  This does not apply on multi-method interfaces.

Do you feel the same way knowing that it is always:  cache.Token?   
We could just go all the way:  cache.ReadReplace

keegan-caruso (approved):

PR #103, opened 2020-12-09
Combine msalbase files

element-of-surprise:
The msalbase is in too many pieces.  This combines this into basically 3 files:
internal.go - Types that aren't exposed
msalbase.go - Types that are exposed
mock* - Mock implementation

From here I'm going to do renames and see about what can be made private.


internal/msalbase/internal.go
```diff
@@ -0,0 +1,307 @@
+package msalbase
```
sangonzal, line 1: New files need license headers.

sangonzal (approved):

internal/msalbase/internal.go
keegan-caruso, line 1: all the new files need the copyright header

jhendrixMSFT (approved):

PR #105, opened 2020-12-09
Squash msal files

element-of-surprise:
msal/ has too many files to look through and figure out what is what.  This reduces the files to just a handful grouped by function.

- There is only one structural change, which is removing a mock that wasn't actually used.
- There is one internal/ change that simple moved a type around.

This allows a better view to accomplish goals like:

- Removing *Providers, which was an outstanding item
- Looking at the structure to change our API surface


msal/apps.go
```diff
@@ -0,0 +1,442 @@
...
+	}
+
+	storageTokenResponse, err := client.manager.Read(ctx, authParams, client.webRequestManager)
+	if err != nil {
+		return msalbase.AuthenticationResult{}, err
+	}
+
+	result, err := msalbase.CreateAuthenticationResultFromStorageTokenResponse(storageTokenResponse)
+	if err != nil {
+		if reflect.ValueOf(storageTokenResponse.RefreshToken).IsNil() {
```
keegan-caruso, line 74: we missed this in a previous refactor, this doesn't make much sense anymore and should be an IsZero check. can we get a todo or a issue?

keegan-caruso (approved):
The few linting failures would be good to fix, otherwise LGTM

sangonzal (approved):
:shipit:

abhidnya13 (approved):
🚢

element-of-surprise:
Fixed up 2 lints, but 1 was: Error: SA5011: possible nil pointer dereference (staticcheck).  It didn't give any information on where this was or what file.

PR #110, opened 2020-12-15
Squashing requests package

element-of-surprise:
Prequel to a refactor


internal/requests/requests.go
```diff
@@ -0,0 +1,497 @@
+package requests
```
keegan-caruso, line 1: copyright headers on both new files

keegan-caruso (approved):

PR #112, opened 2020-12-17
Fixes on OauthTokenBase

element-of-surprise:
OAuthResponseBase is used weirdly.  It should be used with composition and instead we do double unmarshalling and then use it as an attribute.

I've removed this so we only have to JSON decode once.  

Also, changed the integration test so that if --short is used we won't do them.  This allows people like myself to skip integration tests when we don't have the setup.


keegan-caruso (approved):
LGTM

PR #113, opened 2020-12-18
New comms package

element-of-surprise:
Comms package, part of a new ops/ package, provides http calls using XML/REST/Soap. 

This divides our transport layer from our business logic.  

Business logic will come following this.

### DIFF
--- a/internal/ops/internal/comm/comm.go
+++ b/internal/ops/internal/comm/comm.go
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
 // Package comm provides helpers for communicating with HTTP backends.
 package comm
 
@@ -157,11 +160,7 @@ func (c *Client) xmlCall(ctx context.Context, u *url.URL, headers http.Header, b
 		return err
 	}
 
-	if err := xml.Unmarshal(data, resp); err != nil {
-		return err
-	}
-
-	return nil
+	return xml.Unmarshal(data, resp)
 }
 
 // URLFormCall is used to make a call where we need to send application/x-www-form-urlencoded data
@@ -198,7 +197,7 @@ func (c *Client) URLFormCall(ctx context.Context, endpoint string, qv url.Values
 	}
 
 	var unmarshal = json.Unmarshal
-	if _, ok := v.Type().FieldByName("AdditionalFields"); ok {
+	if _, ok := v.Elem().Type().FieldByName("AdditionalFields"); ok {
 		unmarshal = customJSON.Unmarshal
 	}
 	if resp != nil {
--- a/internal/ops/internal/comm/comm.go
+++ b/internal/ops/internal/comm/comm.go
@@ -93,7 +93,7 @@ func (c *Client) JSONCall(ctx context.Context, endpoint string, headers http.Hea
 	return nil
 }
 
-// XMLCall sends XML encoded in body and decodes the XML response into resp. This is used when
+// XMLCall connects to an endpoint and decodes the XML response into resp. This is used when
 // sending application/xml . If sending XML via SOAP, use SOAPCall().
 func (c *Client) XMLCall(ctx context.Context, endpoint string, headers http.Header, qv url.Values, resp interface{}) error {
 	if err := c.checkResp(reflect.ValueOf(resp)); err != nil {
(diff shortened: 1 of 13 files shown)


internal/ops/internal/comm/comm.go
```diff
@@ -0,0 +1,285 @@
+// Package comm provides helpers for communicating with HTTP backends.
```
keegan-caruso, line 1: copyright headers on new files

internal/ops/internal/comm/comm.go
```diff
@@ -0,0 +1,285 @@
...
+		qv = url.Values{}
+	}
+
+	u, err := url.Parse(endpoint)
+	if err != nil {
+		return fmt.Errorf("could not parse path URL(%s): %w", endpoint, err)
+	}
+	u.RawQuery = qv.Encode()
+
+	headers.Set("Content-Type", "application/xml; charset=utf-8") // This was not set in he original GetMex(), but...
```
keegan-caruso, line 113: at the minimum lets track this with a TODO and and owner, a GH issue would be better

internal/ops/internal/comm/comm.go
```diff
@@ -0,0 +1,285 @@
...
+	if len(body) > 0 {
+		req.Body = ioutil.NopCloser(strings.NewReader(body))
+	}
+
+	data, err := c.do(ctx, req)
+	if err != nil {
+		return err
+	}
+
+	if err := xml.Unmarshal(data, resp); err != nil {
```
keegan-caruso, line 160: I think you can return directly here

internal/ops/internal/comm/comm.go
```diff
@@ -0,0 +1,285 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// URLFormCall is used to make a call where we need to send application/x-www-form-urlencoded data
+// to the backend and receive JSON back. qv will be encoded into the request body.
+func (c *Client) URLFormCall(ctx context.Context, endpoint string, qv url.Values, resp interface{}) error {
```
keegan-caruso, line 169: It would be great to get some test coverage over this function

internal/ops/internal/comm/comm.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+// readBody reads the body out of an *http.Response. It supports gzip encoded responses.
+func (c *Client) readBody(resp *http.Response) ([]byte, error) {
+	var reader io.Reader = resp.Body
+	switch resp.Header.Get("Content-Encoding") {
+	case "":
+		// Do nothing
+	case "gzip":
+		reader = gzipDecompress(resp.Body)
```
keegan-caruso, line 262: similar comment, some sort of testing over gzip here would be good. but not sure if you are adding that at another level

msal/default_web_request_manager.go
```diff
@@ -246,11 +246,11 @@ const (
 )
 
 func addAADHeaders(headers http.Header, authParameters msalbase.AuthParametersInternal) {
```
keegan-caruso, line 248: guessing this gets removed with the business logic changes?

keegan-caruso (approved):
Couple small changes that would be good to see, otherwise LGTM

internal/ops/internal/comm/comm.go
element-of-surprise: I'm not sure what this comment means.  

The comment here is basically a note, as we don't have good end to end tests so I don't know if this is going to affect this call.  

But it seems like it should have this header

internal/ops/internal/comm/comm.go
element-of-surprise: Was going to let the end to end tests cover this.  At this point I'm not even sure any of the backends support gzip. 

Because Microsoft has no standard libraries, gzip support is very spotty in our services.  My guess is we don't support gzip at all, but it would require real life testing to find out.

msal/default_web_request_manager.go
element-of-surprise: These won't get removed until the integration changes.  The business logic will be separate, but I don't want to integrate at the same time because that will probably have massive impacts.

internal/ops/internal/comm/comm.go
element-of-surprise: This was me overlooking adding the test.  Added.

PR #115, opened 2020-12-23
New REST business logic

element-of-surprise:
Creates the ops package.  This package is simply a convenience wrapper for creating sub-clients that are grouped into categories based upon what the calls do. These categories are:

- accesstokens
- authority
- wstrust

This complete the business logic separation, but not the integrations (we are still using the old code).

This also refactors the saml definitions so that they used named structs.  This allows us to write tests using those structs instead of having to recreate the struct, which is virtually impossible with the size of these structs.


internal/ops/accesstokens/accesstokens_test.go
```diff
@@ -0,0 +1,686 @@
+package accesstokens
```
keegan-caruso, line 1: copyright header

internal/ops/accesstokens/accesstokens.go
```diff
@@ -0,0 +1,188 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/*
+Package accesstokens exposes a REST client for querying backend systems to get various types of
+access tokens (oauth) for use in authentication.
+
+These calls are of type "application/x-www-form-urlencoded".  This means we use url.Values to
+represent arguements and then encode them into the POST body message.  We receive JSON in
```
keegan-caruso, line 9: arguements => arguments

internal/ops/accesstokens/accesstokens.go
```diff
@@ -0,0 +1,188 @@
...
+type createTokenResp func(authParameters msalbase.AuthParametersInternal, payload msalbase.TokenResponseJSONPayload) (msalbase.TokenResponse, error)
+
+// Client represents the REST calls to get tokens from token generator backends.
+type Client struct {
+	// Comm provides the HTTP transport client.
+	Comm          urlFormCaller
+	TokenRespFunc createTokenResp
+}
+
+// GetAccessTokenFromUsernamePassword uses a username and password to get an OIDC access token.
```
keegan-caruso, line 49: not sure it is right to call it an OIDC access token, I'd just call it an access token. here and at 62

internal/ops/wstrust/wstrust_test.go
```diff
@@ -0,0 +1,308 @@
...
+		return fmt.Errorf("got endpoint == %s, want endpoint == %s", f.gotEndpoint, endpoint)
+	}
+
+	// Removes a uuid that will change every time.
+	// example: `urn:uuid:373ea2fa-d586-4cad-8bb8-10392ddbb5c6</wsa:messageID>``
+	bodyStr := replaceURNRE.ReplaceAllString(body.(string), `</was:messageID>`)
+	gotBodyStr := replaceURNRE.ReplaceAllString(body.(string), `</was:messageID>`)
+
+	if diff := diff.Diff(
+		strings.ReplaceAll(bodyStr, ">", ">\n"),    // So we can do a line by line comparision
```
keegan-caruso, line 130: nit: comparision => comparison

internal/ops/wstrust/wstrust_test.go
```diff
@@ -0,0 +1,308 @@
...
+				},
+			},
+		},
+	}
+
+	for _, test := range tests {
+		fake := &fakeXMLCaller{err: test.commErr, giveResp: test.giveResp}
+		client := Client{Comm: fake}
+
+		authParams.AuthorizationType = test.authorizationType
```
keegan-caruso, line 287: Can we get a test for AuthorizationType == AuthorizationTypeUsernamePassword here as well?

msal/default_web_request_manager.go
```diff
@@ -109,13 +111,19 @@ func (wrm *defaultWebRequestManager) GetWsTrustResponse(ctx context.Context, aut
 		soapAction = SoapActionDefault
 	}
 
+	if endpoint.URL == "" {
```
keegan-caruso, line 114: It know the default web request manager is leaving, but we should at least take the panic out

internal/ops/wstrust/wstrust_test.go
```diff
@@ -0,0 +1,308 @@
...
+					RequestSecurityTokenResponseCollection: wstrust.RequestSecurityTokenResponseCollection{
+						RequestSecurityTokenResponse: []wstrust.RequestSecurityTokenResponse{
+							{
+								RequestedSecurityToken: wstrust.RequestedSecurityToken{
+									Assertion: wstrust.Assertion{
+										Text: "hello",
+										XMLName: xml.Name{
+											Local: "Assertion",
+										},
+										Saml: samlv1Assertion,
```
keegan-caruso, line 272: minor, but a samlv2Assertion would gives us another scenario

PR #116, opened 2020-12-24
Remove Logging from main code and logrus

element-of-surprise:
Logrus is a third-party package for logging.  However, it is not the standard library logger and many of our users will not want logrus.

There is not standard interface that defines logging for Go and this causes real problems (https://github.com/golang/go/issues/13182).

While we could define one for users to implement, there is very little reason at this point to log.  All of our required output required can be captured by outputting errors to the user.  They can choose how they log.  

The only useful logging that was done here was debugging for a dev via the "Trace".  This is a logrus thing.  If we were going to do tracing, we'd want to do a distributed tracer via the context object and have it in more places.  So this was removed.

This leaves logging only in the sample apps using the go standard library logger.


internal/wstrust/wstrust_mex_document.go
```diff
@@ -129,23 +120,17 @@ func CreateWsTrustMexDocumentFromDef(definitions Definitions) (MexDocument, erro
 				return MexDocument{}, fmt.Errorf("cannot create MexDocument: %w", err)
 			}
 
-			log.Tracef("Associated port '%v' with binding, url '%v'", bindingName, url)
 			switch binding.EndpointType {
 			case wsEndpointTypeUsernamePassword:
-				if updateEndpoint(&usernamePasswordEndpoint, endpoint) {
```
keegan-caruso, line 135: looks like updateEndpoint can be simplified a bit if it doesn't need to return a bool

keegan-caruso (approved):
Would be good to remove the bool from updateEndpoint, otherwise LGTM

PR #117, opened 2021-01-04
Changes everything as a result of integrating the ops package

element-of-surprise:
Please see the design.md file for more information, but here is an overall list:

- Integrates ops/ into the codebase
- Moves tree from msal/ to apps/
- Deletes msal/msal.go in favor of apps/public apps/confidential
- Deletes msalbase, a single type is left and it was renamed to shared
- We no longer have public API types come from internal/. We do have a type alias to an internal type.
- Reorganization of the internals and a lot of function and type renames
- Deletion of a lot of constructors that weren't necessary

Those are just the highlights.


apps/confidential/confidential.go
```diff
@@ -0,0 +1,364 @@
...
+
+// tokener has a shared requests.Token object. I (jdoak) am not a fan. But at this point, that
+// object is internal/ and I don't want to pull it out. A confidential.Client is mean to be made
+// per user, so we don't want to be creating a bunch of requests.Token objects.
+var tokener = requests.NewToken()
+
+// New is the constructor for Client. userID is the unique identifier of the user this client
+// will store credentials for (a Client is per user). clientID is the Azure clientID and cred is
+// the type of credential to use.
+func New(userID, clientID string, cred Credential, options ...Option) (Client, error) {
```
jhendrixMSFT, line 225: While I get why you did this, I can't help but feel that the naming is now less intuitive/descriptive.

```go
// wordy but obvious
client := msal.NewConfidentialClientApplication()

// succinct but less clear
client := confidential.New()
```

Don't get me wrong, a lot of the type names before this change we just too verbose.  I just feel like we've swung the pendulum too far in the other direction.

I don't have a suggestion at present, and I'm fine with making such changes post-merge of this PR.

apps/design/design.md
```diff
@@ -0,0 +1,159 @@
...
+Before I began working on MSAL I was re-writing the Keyvault Go API.  https://github.com/element-of-surprise/keyvault . It does the right things to extract cers for TLS now. 
+I was still working on the Cert() API and hadn't exposed the public surface when I stopped.
+
+Since we have representation from the Go SDK team, we might have them go bridge this problem in
+the current implementation using some of that code so its possible for our users to store the
+cert in Keyvault.
+
+## Notes:
+
+Do we need: AcquireTokenSilent()??  Seems like we could just bake this into other acquire calls automatically????
```
abhidnya13, line 159: @sangonzal  recently had a [proposal](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/1948?_a=files&path=%2FConfidentialClient%2FClientCredentialCacheBehavior.md) to move confidential client flows to actually do the acquire token silent calls within the library instead of asking the developers to exclusively do it. So we really do not need it in the confidential client application acquire token calls where there is no user account associated. However, for public clients, acquire token silent call is still needed as acquire token silent would go in and find an access token for an account in the token cache that is associated with a user account. Developers would be the ones choosing which user accounts they want to do the cache look up for. That would look something like this : https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/sample/device_flow_sample.py#L49-L57

apps/design/design.md
```diff
@@ -0,0 +1,159 @@
...
+
+This is the general way to add a new feature to MSAL:
+
+- Add the REST calls to ops.REST
+- Add the higher level manipulations to requests.Token
+- Add your logic to the app/\<client\> and access the services via your requests.Token
+
+## Notable Differences To Other Clients
+
+### Confidential applications have a user ID
```
abhidnya13, line 100: This is a good addition so that users are aware of exactly how to use our library. If we bake in acquire token silent call in all the acquire_token calls for confidential clients, will this still be needed ?

abhidnya13 (commented):
Providing some MSAL perspective after going through the design doc.

apps/confidential/confidential.go
jhendrixMSFT: What is `userID`?  This param wasn't in the previous iteration.

.gitignore
```diff
@@ -4,6 +4,7 @@
 *.dll
 *.so
 *.dylib
+*.golangci.yml
```
keegan-caruso, line 7: I'd rather changes to this get reviewed and committed.

apps/confidential/confidential.go
keegan-caruso: see: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/9dbe73b396b302c2daa32f24847745f7b63bd94b/apps/design/design.md#confidential-applications-have-a-user-id

element-of-surprise: Naming is probably the hardest thing to get right, especially in Go where we believe in small names and using package structure as part of names.  And its a philosophical argument that will be happening long after I'm dead.  .Net and Java seem to think that StringObjectThatRepresentsANameFactory is good naming.  I of course dislike this terribly.  

We also tend to base these things on feelings of what we like or not.  This is because are brains are hardwired from what we have done before and to conserve energy our brains like to follow this path.  

So I'd like to take these discussion down a more evidence based path of discussion.  Let's take a kinda first principals approach, break down what are we trying to accomplish with our naming.  Then what this naming is doing and see if it meets our goals.  If it doesn't meet them, then let's talk about what would meet them.  

Here is what I see as the goals with our naming:

- Clear with what something does with a name
- Clear across multiple language implementations
- Easy to find the starting point

The first is kind of a two prong problem.  ConfidentialClientApplication() has no real meaning.  It is a confusing term.  app and server would have been better labels.  But that is already done.  So really it is now about: is the name clear if you know the terminology of Go and MSAL.

The second is, if you have used MSAL before, is the name clear enough within tight package contents, to be not have to spend a lot of cognitive energy making the translation.

To accomplish these we use a combination of:
- directory naming
- package naming
- type names and a type system
- small packages with few types (to make searches fast)

With this said, let's pretend I know Go and I've done MSAL from Python.  

My first step is to get my application client.  Let's say its the "confidential one".  Where would that be?

apps/ (no other places to be)

Okay, there is no apps package and I need confidential, is it clear where to find it from this:

apps/
  confidential/
  public/

So I see confidential, now, how do I get the client.  In Go, this is usually done with New().  If I look at the package, is it clear how to do that? 

To answer that, we have 2 of the 3 things I think we need:
- New() (no extra words) is generally used to indicate the main type of a package
- Is the godoc going to be short so there is very little to look through
- Is there a godoc example

I believe we are a two out of three here.  We use New(), we have 2 types and 3 constructors total.  The Godoc will be about a page in length.  

What we fail at is a godoc example.  A good one you copy and paste what is there for your own use.  We don't have one yet, but we will in the future

With this rational, do you believe that the naming is faulty, confusing or unclear to an engineer who has Go experience and MSAL experience?  If so is that lack of clarity for a moment (like 1 second) or is just so confusing that they are lost?  Would our devs have problems figuring it out?

apps/design/design.md
element-of-surprise: Can you write an item that proposes the changes we need to do here?

apps/design/design.md
element-of-surprise: So saying we bake it in, are you saying that we will not need a client per user?  
If you are not saying that, what is then the signaling mechanism to the dev?

.gitignore
element-of-surprise: I'm not sure what that means.  Do you mean separately?  Does this not count as a review, do you mean in our meeting?

keegan-caruso: I'd rather it not be in the .gitignore is all. If we need to change golangci.yml, then it should be changed.

apps/design/design.md
abhidnya13: We are already tracking it here : https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/48

apps/design/design.md
abhidnya13: Just saw a comment mentioning this. https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/117/files#diff-b55fa119dfdc3c44c3fb9c7b1cbb42bef63ebfdcca4ab41d2cac4c6d8ffdd948R170. If this is not used anywhere and is just there to make the developer aware, I think it would not matter even if we bake acquire token silent in the call. Resolving this

apps/public/public.go
```diff
@@ -0,0 +1,251 @@
...
+	authParams := pca.AuthParams
+	authParams.Scopes = scopes
+	authParams.AuthorizationType = authority.AuthorizationTypeDeviceCode
+
+	dc, err := pca.Token.DeviceCode(ctx, authParams)
+	if err != nil {
+		return DeviceCode{}, err
+	}
+
+	return DeviceCode{Result: dc.Result, ctx: ctx, authParams: authParams, client: pca, dc: dc}, nil
```
jhendrixMSFT, line 193: Not sure how I feel about the fact that we smuggle the provided context in the `DeviceCode` type.  It makes it unclear that `DeviceCode.AuthenticationResult()` makes IO requests.

apps/confidential/confidential.go
jhendrixMSFT: I agree that framed within the context of the module as a whole the naming makes sense.  It's just when I look at `confidential.New()` it seems very vague; perhaps that's an unfair comparison though (side note: while I have an appreciation for Go's preference for terse names [Java/C# are just too wordy], I have some bias for more descriptive names based on years of C/C++ programming where I sometimes had to debug in production environments without IDEs and other niceties; doesn't really translate here).

Ultimately, with proper documentation and examples I think customers will use what we provide and move on.

apps/public/public.go
element-of-surprise: I don't think that a context object indicates IO, because it has a lot of uses beyond that.  But it did get me thinking....

I usually do this type of embedding when internally we will use the object as part of a single flow that has an overall timeout bounded by a context.  That is a good exception to the "no context as an attribute" rule.  

In this case, this probably qualifies as two operations that don't share a combined time.  I'll go back and look at his code and may remove the smuggling.  Update this in a bit.

apps/confidential/confidential.go
```diff
@@ -0,0 +1,364 @@
...
+// NewCredFromCert creates a Credential from an x509.Certificate and a PKCS8 DER encoded private key.
+// CertFromPEM() can be used to get these values from a PEM file storing a PKCS8 private key.
+func NewCredFromCert(cert *x509.Certificate, key crypto.PrivateKey) Credential {
+	return Credential{cert: cert, key: key}
+}
+
+// Client is a representation of authentication client for confidential applications as defined in the
+// package doc. A new Client should be created PER SERVICE USER.
+// For more information, visit https://docs.microsoft.com/azure/active-directory/develop/msal-client-applications
+type Client struct {
```
jhendrixMSFT, line 165: How do I specify a custom HTTP client?  Customers will need this so they can inject things like retry logic and logging.

apps/confidential/confidential.go
```diff
@@ -0,0 +1,364 @@
...
+		return fmt.Errorf("the Authority(%s) does not appear to use https", o.Authority)
+	}
+	return nil
+}
+
+// Option is an optional argument to New().
+type Option func(o *Options)
+
+// Authority allows you to provide a custom authority for use in the client.
+func Authority(authority string) Option {
```
jhendrixMSFT, line 203: A common pattern I've seen is to prefix the variadic config funcs with a "With" prefix.  It's a nice visual cue that this is for configuration, plus all the options are sorted together in Intellisense.

apps/internal/client/client.go
```diff
@@ -0,0 +1,309 @@
...
+
+//GetAccessToken returns the access token of the authentication result
+func (ar AuthenticationResult) GetAccessToken() string {
+	return ar.AccessToken
+}
+
+// GetAccount returns the account of the authentication result
+func (ar AuthenticationResult) GetAccount() shared.Account {
+	return ar.Account
+}
```
catalinaperalta, line 129: Are these methods necessary?

apps/confidential/confidential.go
henrik-me: ignoring all the goodness of go and conciseness... I prefer the more wordy version as suggested above:
client := msal.NewConfidentialClientApplication()
it's consistent with the naming across libraries.

apps/public/public.go
element-of-surprise: I've gone back and built a more traditional flow that should alleviate this concern.

apps/confidential/confidential.go
henrik-me: we discussed this before. we had differences in opinion on this. We should ensure we can inject the  HttpClient, it's what we promise for MSAL's

apps/confidential/confidential.go
```diff
@@ -0,0 +1,364 @@
...
+type Account = shared.Account
+
+// CertFromPEM converts a PEM file (.pem or .key) for use with NewCredFromCert(). The file
+// must have the public certificate and the private key encoded. The private key must be encoded
+// in PKCS8 (not PKCS1). This is usally denoted by the section "PRIVATE KEY" (instead of PKCS1's
+// "RSA PRIVATE KEY"). If a PEM block is encoded and password is not an empty string, it attempts
+// to decrypt the PEM blocks using the password. This will return multiple x509 certificates,
+// though this use case should have a single cert. Multiple certs are due to certificate
+// chaining for use cases like TLS that sign from root to leaf.
+func CertFromPEM(pemData []byte, password string) ([]*x509.Certificate, crypto.PrivateKey, error) {
```
jhendrixMSFT, line 65: Is returning the certs from the intermediate authorities actually valuable?  How do I know which cert in the slice is the one I use for authentication (you can't authenticate with a cert from an intermediate authority, we had this bug in our current authentication library).

apps/confidential/confidential.go
element-of-surprise: One of the things discussed at the meeting was an elimination of custom http logic until a customer comes with a use case that can't be implemented in other ways (such as passing a Retrier via a context object).

I've eliminated that ability for the time being.  However it should be trivial to add in the future as a constructor option.

The previous interface I saw was not up to snuff (the one that had like Get() and Post()).  Any interface should mimic *http.Client methods exactly.  Aka:
``` go
// HTTP provides an HTTP client that will be used for all HTTP calls. This interface is not covered by any
// compatibility promise. Additions to the stdlib http.Client will be added here and may break compatibility without
// a semantic version change.
type HTTP interface {
    CloseIdleConnections()
    Do(req *Request) (*http.Response, error)
    Get(url string) (*http.Response, error)
    Head(url string) (*http.Response, error)
    Post(url, contentType string, body io.Reader) (*http.Response,error)
    PostForm(url string, data url.Values) (*http.Response,error)
}
```

abhidnya13: Tracking this here: #119

apps/confidential/confidential.go
element-of-surprise: I'm not sure it is common enough pattern to add clarity, but it also isn't obtrusive.  Maybe it will become common enough.

I've added With to all options in confidential and public.

apps/internal/client/client.go
keegan-caruso: Likely not, these two fields should be fine to get off of the AuthenticationResult

apps/confidential/confidential.go
element-of-surprise: Well, in this context I think there is only the one certificate (however, I can't be SURE, so I am handling them all).  But.....  in a TLS world, you need all of them or you can't validate the chain and it won't work.

Case in point, when I had to write an internal keyvault client to do TLS for Go, I had to figure this all out.  And to be able to serve TLS, you needed the entire chain that if I remember correctly goes into some type of pool.  I'd have to look back and the code for details.

Also Microsoft issued certs will not natively work with Go's PKCS12 package, I had to do a fork to get it to work.

apps/internal/client/client.go
element-of-surprise: Removed

apps/confidential/confidential.go
element-of-surprise: I'll leave further convesation into that bug.

apps/internal/client/client.go
element-of-surprise: They are removed.

apps/confidential/confidential.go
jhendrixMSFT: I think it would be helpful if there was a list of type names that _must_ be uniform across languages (maybe not 100% identical, but similar enough that it's obvious).

apps/confidential/confidential.go
jhendrixMSFT: There can definitely be more than one certificate.  See https://github.com/Azure/azure-sdk-for-go/issues/5412 for details on the problems as well as our solution.

My problem here is that this func can return multiple certificates however I don't know which one to pass to `confidential.NewCredFromCert()`.  If we have a legitimate use-case for returning all the certs I'm fine with that however we must document which cert is the one to use to authenticate (ideally the first in the slice).

henrik-me:
> */

the cache callback should contain a SuggestedCacheKey which developers would use to separate cache elements.
For desktop scenarios the cache extensions library is a good answer to caching, however for services: [https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/token-cache-serialization#token-cache-for-a-web-app-confidential-client-application](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/token-cache-serialization#token-cache-for-a-web-app-confidential-client-application) holds some good answers on how to handle the cache serialization without running into performance issues. For .net the Identity.Web library does this magic for asp net core and asp. net mvc apps/services. Also the functionality exposed for asp/.net mvc is the preferred for .net without a framework.

See also: [https://github.com/Azure-Samples/active-directory-dotnet-v1-to-v2/blob/master/ConfidentialClientTokenCache/Program.cs](https://github.com/Azure-Samples/active-directory-dotnet-v1-to-v2/blob/master/ConfidentialClientTokenCache/Program.cs) for a simple .net application leveraging this.


---
Refers to: apps/cache/cache.go:11 in 7eb5b45. [](commit_id = 7eb5b4519a229198822a8c808f9d0d2fe9ce85ac, deletion_comment = False)

apps/confidential/confidential.go
```diff
@@ -0,0 +1,364 @@
...
+shared options (fail).  That divides like 2 options from all others which makes the user look
+through more docs.  We can have all clients in one package, but I think separate packages
+here makes for better naming (public.Client vs client.PublicClient).  So I chose a little
+duplication.
+
+.Net People, Take note on X509:
+This uses x509.Certificates and private keys. x509 does not store private keys. .Net
+has some x509.Certificate2 thing that has private keys, but that is just some bullcrap that .Net
+added, it doesn't exist in real life.  Seriously, "x509.Certificate2", bahahahaha.  As such I've
+put a PEM decoder into here.
```
henrik-me, line 46: :smile:

apps/confidential/confidential.go
```diff
@@ -0,0 +1,364 @@
...
+				b, err := x509.DecryptPEMBlock(block, []byte(password))
+				if err != nil {
+					return nil, nil, fmt.Errorf("could not decrypt encrypted PEM block: %w", err)
+				}
+				block, _ = pem.Decode(b)
+				if block == nil {
+					return nil, nil, fmt.Errorf("encounter encrypted PEM block that did not decode")
+				}
+			} else {
+				return nil, nil, fmt.Errorf("encountered encrypted PEM block, but password was not passed")
```
henrik-me, line 85: >return nil, nil, fmt.Errorf("encountered encrypted PEM block, but password was not passed") [](start = 4, length = 91)

in Python and Java I believe we recently had a situation where the password was blank and that is a valid scenario. not sure if that is the same case here but it seems related.

apps/confidential/confidential.go
```diff
@@ -0,0 +1,364 @@
...
+Duplicate Calls shared between public.Client and this package:
+There is some duplicate call options provided here that are the same as in public.Client . This
+is a design choices. Go proverb(https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s):
+"a little copying is better than a little dependency". Yes, we could have another package with
+shared options (fail).  That divides like 2 options from all others which makes the user look
+through more docs.  We can have all clients in one package, but I think separate packages
+here makes for better naming (public.Client vs client.PublicClient).  So I chose a little
+duplication.
+
+.Net People, Take note on X509:
```
henrik-me, line 42: >X509: [](start = 26, length = 5)

Would be good to have this part reviewed by Petro/Ray/Bogdan

henrik-me:
@rayluo @bgavrilMS @SomkaPe : please review when you have time. we will not block this PR waiting for you but pieces here needs some consistency checks with other libraries.

apps/confidential/examples_test.go
```diff
@@ -0,0 +1,34 @@
+package confidential
```
henrik-me, line 1: love the samples and utests are so close to the code. Looks great!

apps/design/design.md
```diff
@@ -0,0 +1,159 @@
...
+- Santiago Gonzalez(Santiago.Gonzalez@microsoft.com)
+
+## History
+
+The original code submitted for Go MSAL was a translation of either Java or .Net code.  This was done as a best effort by an intern who was attempting their first crack at Go.  It had a
+very interesting structure that didn't fit into Go style and made it difficult to understand or
+change. It used global locks, global variables, base type classes (mimicing inheritence), ...
+
+This probably should have be re-written from scratch, but we decided to try and do it in pieces.
+The lesson to be learned from this is that this type of refactor leads to re-writing the code 7 or 8 times instead of once.
```
henrik-me, line 16: anything that you feel is still not go idiomatic which you would have done differently had you started with an empty repo?

MSAL has certain api and quirks, where consistency is valued higher than best possible across libraries.

apps/design/design.md
```diff
@@ -0,0 +1,159 @@
...
+
+This is documented here: https://golang.org/doc/go1.4#internalpackages
+
+For example, a package .../a/b/c/internal/d/e/f can be imported only by code in the directory tree rooted at .../a/b/c. It cannot be imported by code in .../a/b/g or in any other repository.
+
+We use this featurs quite liberally to make clear what is using an internal package.  For example:
+
+```
+apps/internal/base - Only can be used by packages defined at apps/
+apps/internal/base/internal/storage - Only can be use by package client
```
henrik-me, line 59: is this correct? how is this restricted to client? I think I'm misunderstanding something here.

apps/design/design.md
```diff
@@ -0,0 +1,159 @@
...
+
+This is the general way to add a new feature to MSAL:
+
+- Add the REST calls to ops.REST
+- Add the higher level manipulations to requests.Token
+- Add your logic to the app/\<client\> and access the services via your requests.Token
+
+## Notable Differences To Other Clients
+
+### Confidential applications have a user ID
```
henrik-me, line 100: >have a user ID [](start = 30, length = 14)

this should be removed, can be done in a new refactoring. the suggested cache key in the cache callback should handle the scenario on how to partition the cache. A confidential client outside of the obo flow doesn't really work with user id's and generally the account would be null.

apps/design/design.md
```diff
@@ -0,0 +1,159 @@
...
+
+The MSAL caching design is rather simple. These design decisions and the fact that multiple applications in different languages can share a cache mean it cannot be easily changed.
+
+The entire cache contents of a confidential.Client is read and written on 
+almost any action to and from an external cache. 
+
+It is not clear to a user that a confidential client should be per user to prevent scaling
```
henrik-me, line 107: >per user  [](start = 63, length = 9)

See above and other comments I made regarding the cache elsewhere. we can take a follow up item to make changes to this. @patil.abhidnya I believe you are already tracking this, let's ensure this gets  captured in an appropriate work item

apps/design/design.md
```diff
@@ -0,0 +1,159 @@
...
+We could create a new caching mechanism for Confidential that doesn't have this problem.  Treat
+it differently than Public. This would be a larger migration path across all SDKs, but it is
+probably the right long term move.
+
+### Use of x509.Certificate and CertFromPEM() function
+
+The original version of this package used an thumbprint and a private key to do authorizations
+based on a certificate. But there wasn't a real way to get a thumbprint.
+
+A thumbprint is defined in the Oauth spec, which we had to track down. It is an SHA-1 hash
```
henrik-me, line 134: > It is an SHA-1 hash [](start = 70, length = 20)

can be sha-2 afaik

henrik-me, line 134: >It is an SHA-1 hash [](start = 71, length = 19)

I believe other hashing is available in the updated spec, but we have to support sha-1 for backwards compatibility.

apps/internal/base/base.go
```diff
@@ -0,0 +1,299 @@
...
+		return "", err
+	}
+
+	v := url.Values{}
+	v.Add("client_id", clientID)
+	v.Add("response_type", "code")
+	v.Add("redirect_uri", redirectURI)
+	v.Add("scope", strings.Join(scopes, " "))
+
+	// There were left over from an implementation that didn't use any of these.  We may
```
henrik-me, line 165: some of these are needed for proper implementation of the auth code flow, some are needed for parameters that a developer would pass in for special behaviors. we should have work items to address those.

apps/internal/base/base_test.go
```diff
@@ -0,0 +1,130 @@
...
+	}
+
+	for _, test := range tests {
+		got, err := CreateAuthenticationResultFromStorageTokenResponse(test.storeToken)
+		switch {
+		case err == nil && test.err:
+			t.Errorf("TestCreateAuthenticationResultFromStorageTokenResponse(%s): got err == nil, want == != nil", test.desc)
+			continue
+		case err != nil && !test.err:
+			t.Errorf("TestCreateAuthenticationResultFromStorageTokenResponse(%s): got err == %s, want == == nil", test.desc, err)
```
henrik-me, line 120: >== [](start = 96, length = 2)

nit: ?

apps/internal/base/internal/storage/items_test.go
```diff
@@ -461,11 +462,11 @@ var (
 	hid          = "HID"
 	rtEnv        = "env"
 	rtClientID   = "clientID"
-	rtCredential = "RefreshToken"
+	rtCredential = "accesstokens.RefreshToken"
 	refSecret    = "secret"
 )
 
-var rt = &RefreshToken{
+var rt = &accesstokens.RefreshToken{
```
henrik-me, line 469: >accesstokens.Refres [](start = 10, length = 19)

at.rt contruct looks a bit odd. it's rather a cacheElement.Refreshtoken ?

henrik-me:
> Author: John Doak(jdoak@microsoft.com)

I'm not able to grog everything here, however for our e2e tests we have ways to ensure this works as needed. Validating compatibility with Python, Node, Java and .NET. This work is on the backlog as well. 

---
Refers to: apps/internal/json/design.md:2 in 7eb5b45. [](commit_id = 7eb5b4519a229198822a8c808f9d0d2fe9ce85ac, deletion_comment = False)

henrik-me:
> // Licensed under the MIT license.

can be part of the full implementation of the auth code flow with PKCE

---
Refers to: apps/tests/devapps/authorization_code_sample.go:2 in 7eb5b45. [](commit_id = 7eb5b4519a229198822a8c808f9d0d2fe9ce85ac, deletion_comment = False)

apps/public/public.go
```diff
@@ -0,0 +1,250 @@
...
+// AcquireTokenByUsernamePassword acquires a security token from the authority, via Username/Password Authentication.
+// NOTE: this flow is NOT recommended.
+func (pca Client) AcquireTokenByUsernamePassword(ctx context.Context, scopes []string, username string, password string) (AuthenticationResult, error) {
+	authParams := pca.AuthParams
+	authParams.Scopes = scopes
+	authParams.AuthorizationType = authority.AuthorizationTypeUsernamePassword
+	authParams.Username = username
+	authParams.Password = password
+
+	token, err := pca.Client.Token.UsernamePassword(ctx, authParams)
```
henrik-me, line 148: >Token [](start = 26, length = 5)

that the requests/flows internally are "hanging" on Token (applies everywhere), seems odd to me.

apps/design/design.md
element-of-surprise: Internal naming, some unmarshaling for things like time, etc....

Externally, it is now pretty close to idiomatic

apps/design/design.md
element-of-surprise: See the document on line 51.

apps/design/design.md
element-of-surprise: We use sha-1 and this is an internal detail only.

apps/internal/base/base_test.go
element-of-surprise: Fixed

apps/internal/base/internal/storage/items_test.go
element-of-surprise: I don't understand this question.  I mean the type is right and there is no such thing as a cacheElement.  RefreshToken represents a REST RefreshToken as sent by the server.  

REST access is organized under ops.REST, with the accesstokens coming from a sub-client.

apps/confidential/confidential.go
element-of-surprise: So, I don't think that fix is valid, if you are referring to:
```go 
  if _, ok := block.Headers["localKeyId"]; ok {
              cert, err = x509.ParseCertificate(block.Bytes)
              if err !=  nil {
                  fmt.Printf("Error parsing certificate from pem block:%v\n", err)
                  continue
              }
            }
```
I can find no option header standard for PEM that includes "locakKeyId" to tell you that something is the leaf certificate.  

When I look at this fix, it seems that the origin is from PKCS12.  PKCS12 is a different standard.  When a specific library does a PKCS12 to PEM conversion, it pops this in as an optional header.  

This *looks* to be coming from an asn1 id: asn1.ObjectIdentifier([]int{1, 2, 840, 113549, 1, 9, 21}) stored in PKCS12. (source; https://gitlab.cncf.ci/kubernetes/kubernetes/blob/a610ffd71ac8318c1cb2ddc4a234350e61162575/vendor/golang.org/x/crypto/pkcs12/pkcs12.go)

In standard PEM, this is not available unless a conversion was made from this specific pkcs12 package from what I see.

I don't think that fix is a valid fix for this or the library you linked.

So I think your question is actually still open (for both packages), how do you identify the leaf cert in a PEM encoding for all cases.  At this point I don't know that answer.

apps/confidential/confidential.go
element-of-surprise: Since we think that is a valid scenario, I fixed this to try with the password if the string is the zero value.

apps/design/design.md
element-of-surprise: @abhidnya13 confirm you have this issue (enforce confidential client per user) and I'll remove this.

apps/internal/base/base.go
element-of-surprise: @abhidnya13 can you confirm an entry to track this?

apps/public/public.go
element-of-surprise: Not sure I understand (unless it is the name?)  keegan and I have been talking about reversing the name (this is an old package name) from requests.Token to token.Requestor.  

This package is the upper layer logic that may make 1 or more REST calls in order to get a token from various sources.

Does that help, am I misunderstanding?

apps/design/design.md
element-of-surprise: Also, to be fair, I'm still not "in love".  There is still a certain amount of clunkyness.  I'm not sure if what I dislike is how MSAL does things, that I want more clients per auth service layers, or internal weirdness...

The most important thing is that now, with the exception of like two type aliases, our internal and external are separated.  If we like the external surface, then we now can really change internals completely without harming the external surface.

apps/design/design.md
element-of-surprise: I think someone will need to walk me through what you want changed in the cache.  This is mostly a simplification of what was provided in the original code.  Someone will need to walk me through the changes you want.

As for the integration tests, those are still with @abhidnya13 .

apps/confidential/confidential.go
jhendrixMSFT: The fix was to find the certificate with the same public key as that of the provided private key, see https://github.com/Azure/go-autorest/blob/master/autorest/adal/persist.go#L111.

apps/design/design.md
abhidnya13: Tracking this here: https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/120

apps/confidential/confidential.go
element-of-surprise: Interesting, though that is with a PKCS1 key instead of a PKCS8 key (which is what I added support for).  I'll have to look back, I though the standard for PKCS1 keys was to use RSA PRIVATE KEY and not PRIVATE KEY.  

Though, PEM seems to be very loose. 

Maybe I really need to add support for PKCS1 and PKCS8 and try to decode both types regardless of what the text is.  

I can't imagine how many iterations of features causes certificate/key standards to be such a mess.  ASN1.DER/CER, sometimes base64, sometimes not, inside PKCS1, PKCS12, PKCS8, ............

Text headers with options not listed......

apps/design/design.md
abhidnya13: Tracking this here : https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/117#discussion_r553581029

apps/internal/base/base.go
abhidnya13: Tracking this here: https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/121

apps/confidential/confidential.go
jhendrixMSFT: Yes agreed, it's quite the quagmire.

What if for now we drop support for certs with intermediate authorities and add it later?  As it stands now, this is pretty much not usable unless we assume we get one cert back (i.e. always use the first cert in the slice) which seems brittle.

apps/public/public.go
keegan-caruso: I think it is the name, this isn't obvious from the name that "Token" is a token sub-client/token requestor.

apps/confidential/confidential.go
henrik-me: Alternative is keeping this (perhaps with a few additions) and doing testing with common scenarios where this is needed. We can fail with some meaningful message of we detect something is not working as expected.

apps/design/design.md
element-of-surprise: Removed, new notes added, abhidnya13 opened bug.

apps/design/design.md
element-of-surprise: @henrik-me  does this need to go in this CL or can go in another with a new item tracker?  Cache behavior hasn't changed.

apps/internal/base/base.go
element-of-surprise: Being tracked by abhidnya13

apps/public/public.go
henrik-me: it's the name ... token is not the thing that does the flow, it's rather an artifact. this logic seems inversed to me. the request object is used to get tokens back in the response (among other things)

henrik-me (approved):
:shipit:

apps/confidential/confidential.go
element-of-surprise: I'm going to defer this change to the next CL.  This will get the big changes out. 

The follow-up for this is on my next change.

element-of-surprise: https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/122

keegan-caruso (approved):

PR #125, opened 2021-01-08
Rename requests.Token to oauth.Client

element-of-surprise:
To be more in line with the package's function.


keegan-caruso (approved):
LGTM

PR #126, opened 2021-01-08
Renames

element-of-surprise:
Renames a a bunch of internal types/methods and one external  to keep clarity while providing brevity.


keegan-caruso (approved):
LGTM

PR #127, opened 2021-01-08
Add copyright headers in a few files, add stringer generated go files

element-of-surprise:


keegan-caruso:
Copyright headers on base_test.go, accesstokens.tokens.go, grant.go then should be good

keegan-caruso (approved):
thanks!

PR #129, opened 2021-01-09
Todos: Combine two different enums into a single enum

element-of-surprise:
Two types both representing what type of client is being used.

These were combined into AppType.
Removed stringers for old files and generated new stringer file.


apps/internal/oauth/oauth.go
```diff
@@ -25,7 +25,7 @@ type accessTokens interface {
 	DeviceCodeResult(ctx context.Context, authParameters authority.AuthParams) (accesstokens.DeviceCodeResult, error)
 	FromUsernamePassword(ctx context.Context, authParameters authority.AuthParams) (accesstokens.TokenResponse, error)
 	FromAuthCode(ctx context.Context, req accesstokens.AuthCodeRequest) (accesstokens.TokenResponse, error)
-	FromRefreshToken(ctx context.Context, rtType accesstokens.RefreshTokenReqType, authParams authority.AuthParams, cc *accesstokens.Credential, refreshToken string) (accesstokens.TokenResponse, error)
+	FromRefreshToken(ctx context.Context, rtType accesstokens.AppType, authParams authority.AuthParams, cc *accesstokens.Credential, refreshToken string) (accesstokens.TokenResponse, error)
```
keegan-caruso, line 28: rtType doesn't track well with the rename. same with a couple of the other names. "reqType", "rt", etc

keegan-caruso (approved):
LGTM, thanks!

PR #137, opened 2021-01-13
Remove Get* from method names and fix test names/desc that were not in line with previous name changes

element-of-surprise:
All in subject line


apps/internal/base/internal/storage/storage.go
```diff
@@ -26,10 +26,10 @@ import (
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/shared"
 )
 
-// getAadinstanceDiscoveryResponser is provider to allow for faking in tests.
+// getAADInstanceDiscoveryResponser is provider to allow for faking in tests.
 // It is always implemented in production by ops/authority.Client
-type getAadinstanceDiscoveryResponser interface {
-	GetAadinstanceDiscoveryResponse(ctx context.Context, authorityInfo authority.Info) (authority.InstanceDiscoveryResponse, error)
+type getAADInstanceDiscoveryResponser interface {
```
keegan-caruso, line 31: "get" in a type name is a bit non-standard. comment could also be reworked

apps/internal/base/internal/storage/storage.go
```diff
@@ -26,10 +26,10 @@ import (
...
+	AADInstanceDiscoveryResponse(ctx context.Context, authorityInfo authority.Info) (authority.InstanceDiscoveryResponse, error)
```
keegan-caruso, line 32: can we drop "Response" off of the method name?

keegan-caruso (approved):
LGTM

PR #135, opened 2021-01-13
Fixed cacheAccessor option

allfro:
Refactored `base.go` and `public.go` to accept the `cacheAccessor` parameters properly.

PR #140, opened 2021-01-15
Fixing cache accessor

abhidnya13:
#135  was opened to fix the usage of `noopCacheAccessor`. The PR breaks the integration tests due to a null pointer exception for confidential client with client secret scenario. 
This PR is opened to fix the bug in `base.go` instead of making the same changes in both `public.go` and `confidential.go`

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -123,11 +123,13 @@ func New(clientID string, authorityURI string, cacheAccessor cache.ExportReplace
 		return Client{}, err
 	}
 	authParams := authority.NewAuthParams(clientID, authInfo)
-
+	if cacheAccessor == nil {
+		cacheAccessor = noopCacheAccessor{}
+	}
 	return Client{ // Note: Hey, don't even THINK about making Base into *Base. See "design notes" in public.go and confidential.go
 		Token:         token,
 		AuthParams:    authParams,
-		cacheAccessor: noopCacheAccessor{},
+		cacheAccessor: cacheAccessor,
 		manager:       storage.New(token),
 	}, nil
 }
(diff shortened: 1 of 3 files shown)


apps/internal/base/base.go
```diff
@@ -123,11 +123,13 @@ func New(clientID string, authorityURI string, cacheAccessor cache.ExportReplace
 		return Client{}, err
 	}
 	authParams := authority.NewAuthParams(clientID, authInfo)
-
+	if cacheAccessor == nil {
```
keegan-caruso, line 126: might be worth documenting this in the comment for the method

keegan-caruso (approved):

apps/internal/base/base.go
element-of-surprise: The other option is to make this a Go style option:

```go
func Option func(c *Client)

type WithCacheAccessor(ca cache.ExportReplace) Option {
  return func(c *Client) {
    if c == nil {
      return
    }
    c.cacheAccessor = ca
}

func New(clientID, authorityURI, token *oauth.Client, options ...Option) (Client, error) {
  ...
  client := Client {
    ...
   cacheAccessor: noopCacheAccessor{}
  }
  for _, o := range options {
    o(&client)
  }
  return client, nil
```

Then its a true option (which I prefer over accepting nils).

abhidnya13: @element-of-surprise , Tried changing this to accept it as an option in the latest commit.  
What do you think ?

keegan-caruso: Looks better, thanks

abhidnya13: Merging this in.

PR #141, opened 2021-01-15
Error design document

abhidnya13:


apps/errors/error_design.md
```diff
@@ -0,0 +1,47 @@
+# MSAL Error Design
```
henrik-me, line 1: it should be easy for the consumer of the library to know what they need to do based on the error returned (exception or not). The client exceptions have one purpose, service exceptions has another purpose. also we need to support classifications to ensure we can follow standard MSAL patterns.

element-of-surprise: @henrik-me 

This should do all of that.  You can diff on if the error is a service error with the built in errors.Is().  All errors contain error text on what the issue is.  Service errors come with not only the error but a Verbose(err) to yank out the http request and response object.

If we need to add specific sub-errors that can be switched on, we can do that with the same error typing we do with service errors and they can be detected with errors.Is() as well.  

I would say that the difficulty for the consumer will be in interpreting what the server side is saying, as those errors are not always clear.

apps/errors/error_design.md
```diff
@@ -0,0 +1,47 @@
...
+
+MSAL throws MsalClientException for things that go wrong inside the library (e.g. bad configuration)
+
+All the equivalents of MSALClientExceptions are solved by throwing an on-the go error like this :
+
+``` Go
+return "", errors.New("authority does not have two segments")
+```
+
+## Service side errors
```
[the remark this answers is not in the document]
element-of-surprise: It looks like part of the markdown got caught in the ``` - ``` section, specifrically ##Future Decisions.  That needs to get fixed.

henrik-me (approved):
:shipit:

abhidnya13:
Thanks @element-of-surprise for the review, the document looks very polished ! Updated with your suggestions in the latest commit !

apps/errors/error_design.md
```diff
@@ -0,0 +1,119 @@
...
+    e.Resp.Request = nil // This brings in a bunch of TLS crap we don't need
+    e.Resp.TLS = nil     // Same
+    return fmt.Sprintf("%s:\nRequest:\n%s\nResponse:\n%s", e.Err, prettyConf.Sprint(e.Req), prettyConf.Sprint(e.Resp))
+}
+```
+
+A user will always receive the most concise error we provide.  They can tell if it is a server side error using the Go error package:  
+
+```go
+if errors.Is(err, CallErr{}) {
```
jhendrixMSFT, line 81: This is incorrect.  `errors.Is()` is for checking for a sentinel error value like `io.EOF`.

https://blog.golang.org/go1.13-errors

abhidnya13: Not a Go expert, but I think you are right. 
From what the blog says,  we can still use errors.As() for type checking to see if the errors is of type CallError which would mean it can be used to differentiate between client side and server side errors, right ? @jhendrixMSFT

jhendrixMSFT: Your assessment is correct.

There's an interesting [blog](https://dave.cheney.net/2016/04/27/dont-just-check-errors-handle-them-gracefully) from Dave Cheney where he asserts "Assert errors for behaviour, not type".  Granted this was written before `errors.As()` and `errors.Is()` were available so perhaps the thinking has somewhat changed, although the advent of those functions don't eliminate the problems he cites.  @element-of-surprise thoughts on this?

element-of-surprise: That is probably my mistake, this looks like an example I gave.  I rarely use errors.Is() or As().  Your assessment seems correct: https://play.golang.org/p/GB8epf1jIov

For our users, they can also simply do it old school:

```go
if ce, ok := err.(CallError); ok { // CallError is always the outer error
  // Do something
}
```

My thoughts tend to go in a similar vein to Dave.  There are two types of errors.
- Ones you can do something about
- Ones you can't

Error values and error types(like CallError) are about errors you can do something about.  fmt.Errorf("") is about ones you can't.  

In my packages I use error values/types when I need to distinguish types that I might be able to do things like retry the call or I need to extract something very specific from the error that I can give the user (I need fields and not just a string).

After that it is strictly about error clarity for the user to manually correct the problem.

abhidnya13: Thanks for this discussion, updated this part in the latest commit !

apps/errors/error_design.md
```diff
@@ -0,0 +1,110 @@
...
+    e.Resp.Request = nil // This brings in a bunch of TLS crap we don't need
+    e.Resp.TLS = nil     // Same
+    return fmt.Sprintf("%s:\nRequest:\n%s\nResponse:\n%s", e.Err, prettyConf.Sprint(e.Req), prettyConf.Sprint(e.Resp))
+}
+```
+
+A user will always receive the most concise error we provide.  They can tell if it is a server side error by checking the type of the error like this:
+
+```go
+if ce, ok := err.(CallError); ok { // CallError is always the outer error
```
jhendrixMSFT, line 81: Why did this change to a type assertion?

apps/errors/error_design.md
```diff
@@ -0,0 +1,110 @@
...
+  // Do something
+}
+```
+
+We provide a Verbose() function that can retrieve the most verbose message from any error we provide:
+
+```go
+fmt.Println(errors.Verbose(err))
```
jhendrixMSFT, line 89: Does this mean we'll be providing an `errors` package with this method?  The name would collide with `errors` from the standard library.

element-of-surprise: Most of this is already designed.  So there already is an errors package.  I included a wrapper around errors.New() in the package already, so name collisions would be less of a problem.

The way pike deals with this is to just duplicate everything in the standard errors package (as it is tiny) into his custom errors package.  We could do the same thing here to avoid any issues.  I think that would be including like 2 more functions.

apps/errors/error_design.md
element-of-surprise: I like the type assertion way (old school), but Joel is write, you should do errors.As().

element-of-surprise (approved):
Just the errors.As() that Joel commented on and it looks good to me.

apps/errors/error_design.md
```diff
@@ -0,0 +1,110 @@
...
+Service side errors occur when an external RPC responds either with an HTTP error code or returns a message that includes an error.
+
+These errors can be transitory (please slow down) or permanent (HTTP 404).  To provide our diagnostic goals, we require the ability to differentiate these errors from other errors.
+
+The current implementation includes a specialized type that captures any error from the server:
+
+```go
+// CallErr represents an HTTP call error. Has a Verbose() method that allows getting the
+// http.Request and Response objects. Implements error.
+type CallErr struct {
```
jhendrixMSFT, line 59: Since this type wraps an error it should also have an `Unwrap() error` method to return the wrapped error (maybe this is already there, unclear if this doc is documenting the entire surface area of this type).

element-of-surprise: We currently are only wrapping errors that contain strings that we generate based on the response content.

Unwrap() error seems like something you'd put here because of the enclosed error.  The logic goes, "hey I see an error containing an error", I need Unwrap().  I'm actually guilty of doing this a few times.

But that isn't actually the case.  Its one more reason I think the additions to errors just made a mess (like go.mod).  If you look at https://blog.golang.org/go1.13-errors , one of the things they talk about is a section called: Whether to Wrap.  

At this time (which could always change, but I don't think so for this one), Unwrap() would have no value and isn't required.

jhendrixMSFT: Thanks for reminding me of this article, I read it a while ago and promptly forgot about it.  Agreed with your assessment.

jhendrixMSFT (approved):

abhidnya13:
Thanks for reviewing all, merging this in now.

PR #145, opened 2021-01-19
Fixing device code sample

abhidnya13:
Was trying the device code flow sample and it was broken because the context getting cancelled immediately. 
Suggesting this change to fix it.


keegan-caruso (approved):
LGTM

PR #149, opened 2021-01-26
Add support for interactive auth flow

jhendrixMSFT:
Implementation for interactive authentication with PKCE.
Launches the default browser with select_account prompt.


jhendrixMSFT:
Need to add tests but wanted to get the review going.

jhendrixMSFT:
Fixes

apps/internal/base/internal/local/server.go
```diff
@@ -0,0 +1,109 @@
...
+		done: make(chan struct{}),
+		s:    &http.Server{},
+	}
+	return rs
+}
+
+// Start starts the local HTTP server on a separate go routine.
+// The return value is the full URL plus port number.
+func (s *Server) Start(reqState string) string {
+	port := rand.Intn(600) + 8400
```
element-of-surprise, line 59: This method of port selection isn't robust.  You are randomly choosing a port >= 8400.  However, there is no guarantees that port will be open.  So the ListenAndServe() may fail immediately.

This also seems very insecure, your redirect URL will be all machines addresses.  That means a scanner or something can trip your redirect on a public IP.  I think you want to use an IPv4 or IPv6 localhost.

It is also missing return error codes and messages for the browser, so it won't be clear what failed in the http message, just in your internal error.

Also, no tests.

Here is an updated version that you can integrate (works a little differently) + tests.
```go
package local

import (
	"context"
	"fmt"
	"net"
	"net/http"
)

var okPage = []byte(`
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8" />
    <title>Authentication Complete</title>
</head>
<body>
    <p>Authentication complete. You can return to the application. Feel free to close this browser tab.</p>
</body>
</html>
`)

const failPage = `
<!DOCTYPE html>
<html>
<head>
    <meta charset="utf-8" />
    <title>Authentication Failed</title>
</head>
<body>
	<p>Authentication failed. You can return to the application. Feel free to close this browser tab.</p>
	<p>Error details: error %s error_description: %s</p>
</body>
</html>
`

// Result is the result from the redirect.
type Result struct {
	// Code is the code sent by the authority server.
	Code string
	// Err is set if there was an error.
	Err error
}

// Server is an HTTP server.
type Server struct {
	// Addr is the address the server is listening on.
	Addr     string
	ctx      context.Context
	resultCh chan Result
	s        *http.Server
	reqState string
	code     string
	err      error
}

// New creates a local HTTP server and starts it.
func New(reqState string) (*Server, error) {
	var l net.Listener
	var err error
	for i := 0; i < 10; i++ {
		l, err = net.Listen("tcp", "localhost:0")
		if err != nil {
			continue
		}
		break
	}
	if err != nil {
		return nil, err
	}

	serv := &Server{
		Addr:     "http://" + l.Addr().String(),
		s:        &http.Server{Addr: "localhost:0"},
		reqState: reqState,
		resultCh: make(chan Result, 1),
	}
	serv.s.Handler = http.HandlerFunc(serv.handler)

	if err := serv.start(l); err != nil {
		return nil, err
	}

	return serv, nil
}

func (s *Server) start(l net.Listener) error {
	go func() {
		err := s.s.Serve(l)
		if err != nil {
			select {
			case s.resultCh <- Result{Err: err}:
			default:
			}
		}
	}()

	return nil
}

// Result gets the result of the redirect operation. Once a single result is returned, the server
// is shutdown. ctx deadline will be honored.
func (s *Server) Result(ctx context.Context) Result {
	select {
	case <-ctx.Done():
		return Result{Err: ctx.Err()}
	case r := <-s.resultCh:
		return r
	}
}

// Shutdown shuts down the server.
func (s *Server) Shutdown() {
	// Note: You might get clever and thing you can do this in handler() as a defer, you can't.
	s.s.Shutdown(context.Background())
}

func (s *Server) putResult(r Result) {
	select {
	case s.resultCh <- r:
	default:
	}
}

func (s *Server) handler(w http.ResponseWriter, r *http.Request) {
	q := r.URL.Query()

	headerErr := q.Get("error")
	if headerErr != "" {
		desc := q.Get("error_description")
		// Note: It is a little weird we handle some errors by not going to the failPage. If they all should,
		// change this to s.error() and make s.error() write the failPage instead of an error code.
		w.Write([]byte(fmt.Sprintf(failPage, headerErr, desc)))
		s.putResult(Result{Err: fmt.Errorf(desc)})
		return
	}

	respState := q.Get("state")
	switch respState {
	case s.reqState:
	case "":
		s.error(w, http.StatusInternalServerError, "server didn't send OAuth state")
		return
	default:
		s.error(w, http.StatusInternalServerError, "mismatched OAuth state, req(%s), resp(%s)", s.reqState, respState)
		return
	}

	code := q.Get("code")
	if code == "" {
		s.error(w, http.StatusInternalServerError, "authorization code missing in query string")
		return
	}

	w.Write(okPage)
	s.putResult(Result{Code: code})
}

func (s *Server) error(w http.ResponseWriter, code int, str string, i ...interface{}) {
	err := fmt.Errorf(str, i...)
	http.Error(w, err.Error(), code)
	s.putResult(Result{Err: err})
}
```

Here's the test:

```go
package local

import (
	"context"
	"io/ioutil"
	"net/http"
	"net/url"
	"strings"
	"testing"
	"time"

	"github.com/kylelemons/godebug/pretty"
)

func TestServer(t *testing.T) {
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()

	tests := []struct {
		desc       string
		reqState   string
		q          url.Values
		failPage   bool
		statusCode int
	}{
		{
			desc:       "Error: Query Values has 'error' key",
			reqState:   "state",
			q:          url.Values{"state": []string{"state"}, "error": []string{"error"}},
			statusCode: 200,
			failPage:   true,
		},
		{
			desc:       "Error: Query Values missing 'state' key",
			reqState:   "state",
			q:          url.Values{"code": []string{"code"}},
			statusCode: http.StatusInternalServerError,
		},
		{
			desc:       "Error: Query Values missing had 'state' key value that was different that requested",
			reqState:   "state",
			q:          url.Values{"state": []string{"etats"}, "code": []string{"code"}},
			statusCode: http.StatusInternalServerError,
		},
		{
			desc:       "Error: Query Values missing 'code' key",
			reqState:   "state",
			q:          url.Values{"state": []string{"state"}},
			statusCode: http.StatusInternalServerError,
		},
		{
			desc:       "Success",
			reqState:   "state",
			q:          url.Values{"state": []string{"state"}, "code": []string{"code"}},
			statusCode: 200,
		},
	}

	for _, test := range tests {
		serv, err := New(test.reqState)
		if err != nil {
			panic(err)
		}
		defer serv.Shutdown()

		u, err := url.Parse(serv.Addr)
		if err != nil {
			panic(err)
		}
		u.RawQuery = test.q.Encode()

		resp, err := http.DefaultClient.Do(
			&http.Request{
				Method: "GET",
				URL:    u,
			},
		)

		if err != nil {
			panic(err)
		}

		if resp.StatusCode != test.statusCode {
			if test.statusCode == 200 {
				t.Errorf("TestServer(%s): got StatusCode == %d, want StatusCode == 200", test.desc, resp.StatusCode)
				res := serv.Result(ctx)
				if res.Err == nil {
					t.Errorf("TestServer(%s): Result.Err == nil, want Result.Err != nil", test.desc)
				}
				continue
			}
			t.Errorf("TestServer(%s): got StatusCode == %d, want StatusCode == %d", test.desc, resp.StatusCode, test.statusCode)
			res := serv.Result(ctx)
			if res.Err == nil {
				t.Errorf("TestServer(%s): Result.Err == nil, want Result.Err != nil", test.desc)
			}
			continue
		}
		if resp.StatusCode != 200 {
			continue
		}

		content, err := ioutil.ReadAll(resp.Body)
		if err != nil {
			panic(err)
		}

		if test.failPage {
			if !strings.Contains(string(content), "Authentication Failed") {
				t.Errorf("TestServer(%s): got okay page, want failed page", test.desc)
			}

			res := serv.Result(ctx)
			if res.Err == nil {
				t.Errorf("TestServer(%s): Result.Err == nil, want Result.Err != nil", test.desc)
			}
			continue
		}

		if !strings.Contains(string(content), "Authentication Complete") {
			t.Errorf("TestServer(%s): got failed page, okay page", test.desc)
		}

		res := serv.Result(ctx)
		if diff := pretty.Compare(Result{Code: "code"}, res); diff != "" {
			t.Errorf("TestServer(%s): -want/+got:\n%s", test.desc, diff)
		}
	}
}
```

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -110,16 +110,20 @@ const (
...
+	CorrelationID       string
+	Endpoints           Endpoints
+	ClientID            string
+	Redirecturi         string
+	HomeaccountID       string
+	Username            string
+	Password            string
+	Scopes              []string
+	AuthorizationType   AuthorizeType
+	State               string
```
element-of-surprise, line 123: Sees you added a few.  For new ones we add, we should document them.  If you know the others enough to document them, that would be great too.

apps/public/public.go
```diff
@@ -251,3 +251,11 @@ func (pca Client) AcquireTokenByAuthCode(ctx context.Context, scopes []string, o
 func (pca Client) Accounts() []Account {
 	return pca.base.Accounts()
 }
+
+// AcquireTokenInteractive acquires a security token from the authority using the default web browser to select the account.
+func (pca Client) AcquireTokenInteractive(ctx context.Context, scopes []string) (AuthResult, error) {
```
element-of-surprise, line 256: We should have an end to end test for this.  But this interacts with the browser.

@abhidnya13 @keegan-caruso do you know how we are going to handle this?

Issue #59, opened 2020-10-27
[Feature Request] Interactive auth with browser


jennyf19:

jhendrixMSFT:
Fixed in `dev`

PR #150, opened 2021-01-26
Add support for interactive auth flow

jhendrixMSFT:
Implementation for interactive authentication with PKCE.
Launches the default browser with select_account prompt.


jhendrixMSFT:
Fixes https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/59

apps/internal/base/base.go
```diff
@@ -257,6 +268,72 @@ func (b Client) AcquireTokenByAuthCode(ctx context.Context, authCodeParams Acqui
 	return b.AuthResultFromToken(ctx, authParams, token, true)
 }
 
+// AcquireTokenByInteractive initiates the interactive auth flow.
```
jhendrixMSFT, line 271: @abhidnya13 @keegan-caruso is there canonical documentation for interactive login that I could paste a link to?  I did a little searching but couldn't find anything.

apps/public/public.go
```diff
@@ -251,3 +251,11 @@ func (pca Client) AcquireTokenByAuthCode(ctx context.Context, scopes []string, o
 func (pca Client) Accounts() []Account {
 	return pca.base.Accounts()
 }
+
+// AcquireTokenInteractive acquires a security token from the authority using the default web browser to select the account.
+func (pca Client) AcquireTokenInteractive(ctx context.Context, scopes []string) (AuthResult, error) {
```
jhendrixMSFT, line 256: @abhidnya13 @keegan-caruso Porting question from previous PR about automated testing thing.

In `azidentity` we have a test hook that overrides `browser.Open` with a simulated response.  This would be half the solution, still need to mock the other API calls.

apps/internal/base/base.go
keegan-caruso: unless @abhidnya13 has something better

https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-authentication-flows#interactive-and-non-interactive-authentication

apps/internal/base/base.go
```diff
@@ -257,6 +268,72 @@ func (b Client) AcquireTokenByAuthCode(ctx context.Context, authCodeParams Acqui
...
+func (b Client) AcquireTokenByInteractive(ctx context.Context, params AcquireTokenInteractiveParameters) (AuthResult, error) {
```
keegan-caruso, line 272: This is strictly a public client concern right? Is there a reason this can't live in the public package then?

apps/internal/base/internal/local/server.go
```diff
@@ -0,0 +1,161 @@
...
+	case <-ctx.Done():
+		return Result{Err: ctx.Err()}
+	case r := <-s.resultCh:
+		return r
+	}
+}
+
+// Shutdown shuts down the server.
+func (s *Server) Shutdown() {
+	// Note: You might get clever and thing you can do this in handler() as a defer, you can't.
```
keegan-caruso, line 112: thing -> think

apps/internal/base/internal/local/server.go
```diff
@@ -0,0 +1,161 @@
+// Package local contains a local HTTP server used with interactive authentication.
```
keegan-caruso, line 1: copyright header

apps/internal/base/internal/local/server_test.go
```diff
@@ -0,0 +1,129 @@
+package local
```
keegan-caruso, line 1: copyright header

apps/internal/base/base.go
jhendrixMSFT: I emulated the other APIs as they appear to perform the heavy lifting inside the base client.  I can certainly move it (and dependencies) around if that's preferred.

apps/internal/base/internal/local/server_test.go
jhendrixMSFT: Fixed

apps/internal/base/internal/local/server.go
jhendrixMSFT: Fixed

apps/internal/base/internal/local/server.go
jhendrixMSFT: Fixed

apps/internal/base/base.go
```diff
@@ -257,6 +268,72 @@ func (b Client) AcquireTokenByAuthCode(ctx context.Context, authCodeParams Acqui
...
+	authParams.CodeChallengeMethod = "S256"
+	authParams.State = uuid.New().String()
+	authParams.Prompt = "select_account"
+	res, err := b.browserLogin(ctx, authParams)
+	if err != nil {
+		return AuthResult{}, err
+	}
+	authParams.Redirecturi = res.redirectURI
+
+	req, err := accesstokens.NewCodeChallengeRequest(authParams, accesstokens.ATPublic, nil, res.authCode, cv)
```
abhidnya13, line 292: I think this function is a little misleading. It mentions new codeChallengeRequest when actually we are sending a code-verifier to the token endpoint. Code challenge is something we send in the request to authcode endpoint. That would also mean changes [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/oauth/ops/accesstokens/accesstokens.go#L189) and [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/oauth/ops/accesstokens/accesstokens.go#L218)

apps/internal/base/internal/local/server.go
```diff
@@ -0,0 +1,164 @@
...
+	resultCh chan Result
+	s        *http.Server
+	reqState string
+}
+
+// New creates a local HTTP server and starts it.
+func New(reqState string) (*Server, error) {
+	var l net.Listener
+	var err error
+	for i := 0; i < 10; i++ {
```
abhidnya13, line 62: Any reason as to why we have 10 retries to listen? Just understanding where that number came from and some sort of documentation for maintenance

apps/internal/base/internal/local/server.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+		break
+	}
+	if err != nil {
+		return nil, err
+	}
+
+	serv := &Server{
+		Addr:     "http://" + l.Addr().String(),
+		s:        &http.Server{Addr: "localhost:0"},
```
abhidnya13, line 75: Not sure what is happening here ? Does this mean Addr in the line above will be different that on this line ? We would want them to be the same right ?

apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -208,7 +208,7 @@ func (c Client) FromAuthCode(ctx context.Context, req AuthCodeRequest) (TokenRes
 			return TokenResponse{}, err
 		}
 	case ATPublic:
-		// Nothing needs to be done, exept to not error.
+		qv = url.Values{}
```
abhidnya13, line 211: Is this different that what is declared on line 196 ?

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -110,16 +110,20 @@ const (
...
+	Redirecturi         string
+	HomeaccountID       string
+	Username            string
+	Password            string
+	Scopes              []string
+	AuthorizationType   AuthorizeType
+	State               string
+	CodeChallenge       string
+	CodeChallengeMethod string
+	Prompt              string
```
abhidnya13, line 126: Not sure if this should be a part of AuthParams, looks a bit odd conceptually, `AuthParams` are parameters used for authorization for token acquistion. These parameters are used for auth_code_acquistion. Having `code` in here would make more sense instead as that is something we use for token_acquistion.

apps/internal/base/internal/local/server.go
element-of-surprise: This was just me being OVERLY cautious.  And I chose it out of my [expletive].  I figured if it tried 10 times to get a port and couldn't, that it was just broken.  

This would only happen if all 65535 ports were in use and you were trying to get one.  I have seen this happen at Google, but this is probably overkill.

Joel, if you want to kill it for 1 attempt, you can.

apps/internal/base/internal/local/server.go
element-of-surprise: When you use a client, you need to have the calling method (such as http:// or https://). 

When you do the server, it works differently.  It doesn't need to know the protocol, it gets that from other settings.  It just needs to know what to listen on.

So http.Server.Addr is so the server knows what to listen on (and doesn't need the method).
Server.Addr is to give our client the whole string it needs to dial the server.

apps/internal/oauth/ops/accesstokens/accesstokens.go
element-of-surprise: Yeah, that tripped me up when I saw it.  

Yes, 196 has a value of nil.
211 has a value of url.Values{} with nothing in it.

So, if the using function got nil and did this:
qv.Get("key")
it would crash.

if it got an empty map, it would not.

apps/internal/base/base.go
abhidnya13: https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-desktop-acquire-token?tabs=dotnet#acquire-a-token-interactively 

I think we should have a wiki page that looks something like this to help customers use our stuff easily. https://github.com/AzureAD/microsoft-authentication-library-for-java/wiki/Acquiring-Token-Interactively
Can track it an a different issue.

jhendrixMSFT: @abhidnya13 since your link doesn't cover Go (for obvious reasons) I'm thinking we should use the link provided by @keegan-caruso for now.  Once we have a better link we can update the docs.  Does that work?

apps/internal/base/internal/local/server.go
abhidnya13: Hahaha just was trying to understand where the magic number came from :) I am fine using 10, rest iterations will just be skipped if it finds it sooner !

apps/internal/base/base.go
abhidnya13: Sounds good !

apps/internal/oauth/ops/authority/authority.go
jhendrixMSFT: It's also used by `Client.AuthCodeURL()` to "creates a URL used to acquire an authorization code.".  So really at present it's used for both code and token acquisition.

I do see your point though that everything is lumped in here together making it a grab-bag of data with no clarity in what flows it's used.  If we want to refactor this I suggest we do it separately.  Thoughts?

apps/internal/base/base.go
jhendrixMSFT: Not sure I follow.  You still have to send the code verifier when requesting the token (the code verifier hash is sent in the challenge when requesting the auth code, the code verifier is sent with the token request).

apps/internal/base/base.go
jhendrixMSFT: fixed

apps/public/public.go
jhendrixMSFT: I've added the hook so we can simulate the auth code request portion of the flow, and stubbed out a test for this.

There is no fake `oauth.Client` exposed at present (you can't even publicly override it) so I can't mock the token request part.  Poking around the code, there are bits and pieces (e.g. `fakeAccessTokens`) but nothing cohesive (unless I missed it).

apps/internal/base/base.go
abhidnya13: Yes, what I mean is we are sending a CodeVerifier in this function named  `NewCodeChallengeRequest` along with an authCode to get back an accessToken. 
However, the name `NewCodeChallengeRequest` is a little bit odd because :
1. It is not a request to get a CodeChallenge
2. It is not requesting something using a CodeChallenge
3. And then in this [line](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/oauth/ops/accesstokens/accesstokens.go#L218) we set a code_verifier with a value of something named `CodeChallenge` which is again misleading.

Let me know if this makes sense, happy to sync offline to elaborate.

apps/internal/oauth/ops/authority/authority.go
abhidnya13: Okay, an issue tracking this would be good.

apps/internal/base/base.go
jhendrixMSFT: I see your point now.  Agreed the name isn't ideal, perhaps something like `NewRequestWithCodeVerifier()` or something along those lines.  It's not public surface area so we can change it now or at any time.

apps/internal/oauth/ops/authority/authority.go
jhendrixMSFT: Opened https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/154

apps/internal/base/base.go
abhidnya13: `NewRequestwithCodeVerifier()` is good. That would probably also mean changes [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/oauth/ops/accesstokens/accesstokens.go#L189) and [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/oauth/ops/accesstokens/accesstokens.go#L218). But its fine if these are not changed in this PR. I can make these changes after this is merged in.

element-of-surprise (commented):
How about NewCodeVerifierReq(). I don't even like that, but NewRequestwithCodeVerifier() is like a sentence.  I didn't like the other one either.

apps/public/public.go
element-of-surprise: Thinking about this a little, maybe this is one ETOE that we put in another place where we can do overrides.  We definitely don't want to expose the oauth.Client to the outside.

apps/internal/base/base.go
```diff
@@ -60,6 +66,12 @@ type AcquireTokenAuthCodeParameters struct {
 	Credential *accesstokens.Credential
 }
 
+// AcquireTokenInteractiveParameters contains the parameters required to acquire an access token using
+// the interactive auth code flow.
+type AcquireTokenInteractiveParameters struct {
```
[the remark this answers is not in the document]
jhendrixMSFT: This type has evaporated after the last refactor.

apps/internal/base/base.go
```diff
@@ -303,3 +380,15 @@ func convertStrUnixToUTCTime(unixTime string) (time.Time, error) {
 	}
 	return time.Unix(timeInt, 0).UTC(), nil
 }
+
+// creates a code verifier string along with its SHA256 hash
```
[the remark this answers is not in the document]
jhendrixMSFT: Fixed

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -110,16 +110,20 @@ const (
...
-	AuthorizationType AuthorizeType
+	AuthorityInfo       Info
+	CorrelationID       string
+	Endpoints           Endpoints
+	ClientID            string
+	Redirecturi         string
+	HomeaccountID       string
+	Username            string
+	Password            string
+	Scopes              []string
```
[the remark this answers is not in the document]
jhendrixMSFT: Fixed.

apps/internal/base/base.go
jhendrixMSFT: We can change it but we should still be mindful of the terminology.  While both legs of the flow send it via the `code_challenge` param, in the first leg of the flow it's called the challenge which is where we send the S256 hash of the code verifier.  The second leg of the flow sends the actual code verifier (same query param, different values).
So for APIs used in the first leg we should still call the param a `challenge`, and only in the second leg of the flow we should call it `codeVerifier`.

jhendrixMSFT: I started to do this but then noticed some other oddities with auth code flow which need to be fixed.  I've opened https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/155 to track refactoring all of this.

apps/public/public.go
jhendrixMSFT: While I can support not exposing `oauth.Client`, without having some kind of fake we can't test this (or other methods) end-to-end (I assume that's what ETOE means).  Did you have something else in mind?

element-of-surprise:
So I think it depends.  If we only have to fake this in a single etoe test, then we could just write the test in the public package where we have access to the oauth.  We can define a private interface there that oauth.Client implements and we can fake.  It would be a one off.

If we must do this more than once, we are going to have to come up with something more elegant.

John Doak
jdoak@microsoft.com

jhendrixMSFT:
Presumably we'd want something general-purpose so we can add automated tests for all auth flows.

jhendrixMSFT:
In `oauth_test.go` we already have some of this machinery in place.  What if we exported it so other tests can access it?

apps/internal/base/base.go
abhidnya13: Yeah that makes sense! This needs to be refactored.

element-of-surprise:
I don't disagree with it in principal, but I'm pretty sure it won't work.  I had never tried to import a test from a test.

So I tried it to do it and it doesn't work.  Test code only compiles when you are running a test, but it compiles the test code for that package only it seems.  You can't important code in a test in another module because that code isn't seen.  Either that or I did something wrong.

You'd have to export this to another regular package, which is fine (as it is internal and won't be seen).  Or do a little copy and paste (copy/paste gets a bad wrap).

John Doak
jdoak@microsoft.com

jhendrixMSFT:
I tried it too, doesn't work.

I would prefer moving it into an `oauth_fake.go` file so that other tests can import it as needed.  While I agree with you regarding copy/paste, it's ~100 LOC to be copying around which seems a bit excessive.  If we all agree on this plan I'll open an issue to track this work.

jhendrixMSFT:
To clarify, place this in `./internal/oauth/fake` package.

element-of-surprise:
Yep, that's fine.

John Doak
jdoak@microsoft.com

jhendrixMSFT:
Unfortunately making the fakes internally available isn't sufficient.  The fields on `oauth.Client` aren't exported (nor are their types).  I see a few solutions.

1. We make the interfaces optional parameters so we can pass in the fakes to `oauth.New()`
2. We export the fields on the client so they can be changed
3. We add something like `oauth.NewFake()` which prepopulates the fields with fakes

My preference is for option 2, it's the simplest plus exposing the underlying interface means they can be faked in other ways.

Thoughts or other ideas?

element-of-surprise:
#2.  Its internal so its fine.

John Doak
jdoak@microsoft.com

jhendrixMSFT:
Test for interactive auth is now enabled.

abhidnya13 (approved):
🚢

Issue #119, opened 2021-01-07
[Feature Request] Support custom http client


abhidnya13:
Refer : https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/117#discussion_r553532502

element-of-surprise:
This should hold until a customer gives a use case we can't support in another way.

element-of-surprise:
If added at some point, should look like:

```go
// HTTP provides an HTTP client that will be used for all HTTP calls. This interface is not covered by any
// compatibility promise. Additions to the stdlib http.Client will be added here and may break compatibility without
// a semantic version change.
type HTTP interface {
    CloseIdleConnections()
    Do(req *Request) (*http.Response, error)
    Get(url string) (*http.Response, error)
    Head(url string) (*http.Response, error)
    Post(url, contentType string, body io.Reader) (*http.Response,error)
    PostForm(url string, data url.Values) (*http.Response,error)
}
```

jhendrixMSFT:
There's no need to add `Get()`, `Head()`, `Post()`, and `PostForm()` as they're convenience methods which aren't applicable to our needs given they don't honor a user-provided context.

I think we can also omit `CloseIdleConnections()`, at least for now until some compelling use-case makes it a requirement (for reference, in the Go SDK we've defined our client interface with just the Do() method and have had no need for this in 4+ years).

element-of-surprise:
A couple things here:

Do we have a current customer for this, if not we should punt this? If that is Joel, you already have code that needs this?
If so, then let's move forward, but if not, we should punt on this.

Joels is right in that we can use just Do() instead of Get(), Head(), Post(), and PostForm().

We cannot do without CloseIdleConnections().

This goes back to the rule: you should never (almost) have a public interface a user can implement.  

Here's why:

HTTP design for today:
```go
type HTTP interface{
  Do(req *Request) (*http.Response, error)
}
```

Our Client's implementation:
```
type HTTP struct {
  client *http.Client
}

func (h *HTTP)  Do(req *http.Request) (*http.Response, error) {
  // Do some custom stuff in here.
  return client.Do(req)
}

HTTP design tomorrow:
```go
type HTTP interface{
  Do(req *Request) (*http.Response, error)
  CloseIdleConnections() // <-added here
}
```

This immediately breaks our clients, as their client is no longer compatible.

[...]

PR #151, opened 2021-01-27
Add support for custom HTTP clients

jhendrixMSFT:
This adds an option for public and confidential clients to specify their
own HTTP client.  By default, the clients use a shared *http.Client.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -21,6 +21,7 @@ import (
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/cache"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/accesstokens"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/authority"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/shared"
@@ -181,6 +182,10 @@ type Options struct {
 	// The default is https://login.microsoftonline.com/common. This can be changed using the
 	// WithAuthority() option.
 	Authority string
+
+	// The HTTP client used for making requests.
+	// It defaults to a shared http.Client.
+	HTTPClient ops.HTTPClient
 }
 
 func (o Options) validate() error {
@@ -212,17 +217,20 @@ func WithAccessor(accessor cache.ExportReplace) Option {
 	}
 }
 
-// tokener has a shared oauth.Token object. I (jdoak) am not a fan. But at this point, that
-// object is internal/ and I don't want to pull it out. A confidential.Client is mean to be made
-// per user, so we don't want to be creating a bunch of oauth.Token objects.
-var tokener = oauth.New()
+// WithHTTPClient allows for a custom HTTP client to be set.
+func WithHTTPClient(httpClient ops.HTTPClient) Option {
+	return func(o *Options) {
+		o.HTTPClient = httpClient
+	}
+}
 
 // New is the constructor for Client. userID is the unique identifier of the user this client
 // will store credentials for (a Client is per user). clientID is the Azure clientID and cred is
 // the type of credential to use.
 func New(clientID string, cred Credential, options ...Option) (Client, error) {
 	opts := Options{
-		Authority: base.AuthorityPublicCloud,
+		Authority:  base.AuthorityPublicCloud,
+		HTTPClient: shared.DefaultClient,
 	}
 
 	for _, o := range options {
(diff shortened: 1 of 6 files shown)


jhendrixMSFT:
Fixes https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/119

apps/confidential/confidential.go
```diff
@@ -212,17 +217,20 @@ func WithAccessor(accessor cache.ExportReplace) Option {
 	}
 }
 
-// tokener has a shared oauth.Token object. I (jdoak) am not a fan. But at this point, that
```
jhendrixMSFT, line 215: Presumably this comment was stale and meant oauth.Client.

One issue I do see though is the cache inside authorityEndpoint.  Obviously this change means there's no longer a shared cache which I'm assuming isn't great.

abhidnya13: This would mean every new instance of confidential client app will do the tenant discovery network call to get the endpoints and wont be able to use the cached endpoints. I think the cost is basically coming down to one network call for each instance of confidential client app to form the authorize and token endpoints in the cache. 
I think that isn't great but its not too bad either.

abhidnya13 (commented):
I think I just have one question, do we need a interface for the response object too ? Is *http.Response a standard in Go ?

In python we define an interface for response something like this:   https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/oauth2cli/http.py
This basically makes sure we enforce a response type should have all the fields we intend to use. 

But I know in Golang interfaces define functionality rather than data, so then we wont need a response interface. Everyone plugging their implementation will have to return a standard http.Response object right ?

element-of-surprise:
*http.Response is the standard in Go.  It has a lot of fields, making an interface in Go isn't feasible.  

We could roll our own, but I don't see the value.  If they are using another http package that doesn't use http.Response, they need to do the translating.  I don't think that is something we should worry about.

In general, I think what we are looking to solve is not a non-standard HTTP package, but some wrappers around HTTP calls.

Note: Python doesn't really have interfaces.  That is a base class that you inherit from.  They are terrible because if you were to add a single call to that base class that was used inside your client, it would break your clients during runtime because they didn't implement it.  With go, that would at least cause your code to not compile, runtime is way worse.

> I think I just have one question, do we need a interface for the response object too ? Is *http.Response a standard in Go ?
> 
> In python we define an interface for response something like this: https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/oauth2cli/http.py
> This basically makes sure we enforce a response type should have all the fields we intend to use.
> 
> But I know in Golang interfaces define functionality rather than data, so then we wont need a response interface. Everyone plugging their implementation will have to return a standard http.Response object right ?

apps/internal/oauth/ops/ops.go
```diff
@@ -17,22 +17,22 @@ Usage is simple:
...
 import (
-	"net/http"
-
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/accesstokens"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/authority"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/internal/comm"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/wstrust"
 )
 
+type HTTPClient = comm.HTTPClient
```
element-of-surprise, line 26: I *think* you might need to copy the comments.  I can't remember if type equivalence copies them for the godoc or not.  Might be worth checking.

abhidnya13:
Thanks @element-of-surprise

Thanks for clarifying. Was trying to understand what this would mean for the customer trying to use anything other than standard http package in Go. We had a customer in Python who just wanted to plug a different library than `requests` to do their http requests and response. 

For us, we currently only have Azure SDK using this who just need it for some additional logging etc and they are still using the standard Go http client so they really wont be needing the conversion part to a standard http.Reponse. But if someone does decide  to use something else they WILL have to do that translation. I think that is fine because the compiler will tell them this. 

And, yes I agree with Python not really having an interface but a Base class creates runtime issues which is worse.

abhidnya13 (approved):

apps/confidential/confidential.go
jhendrixMSFT: Is opening an issue to track this good enough for the moment?  Assuming this doesn't actually break anything of course.

apps/internal/oauth/ops/ops.go
jhendrixMSFT: I think you're right, if anything it will shut up the linter in vs code.

apps/confidential/confidential.go
abhidnya13: Ok, I think issue tracking this would be good. I dont think this will break anything. Our e2e tests also make sure confidential client flows are working fine.

jhendrixMSFT: Opened https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/153

jhendrixMSFT:
Any other comments on this?

keegan-caruso (approved):
thanks!

jhendrixMSFT:
@element-of-surprise any other feedback?

element-of-surprise:
I'd probably comment the Do() and CloseIdleConnections(), but that is really a nit.

Otherwise it is lgtm.

John Doak
jdoak@microsoft.com

PR #174, opened 2021-02-02
Fixing acquire token silent in conf client scenarios with no accounts

### LINKED ISSUE #106: [Bug] Acquire token silent does not work for confidential client scenarios with no account
Acquire token silent with no account which is in case of confidential client app does not give back an access token from cache. 

After going through the code, it tries to find an IDtoken which is missing if the identity is an app and not a user. This is considered as an error and it just returns an empty token response from cache so the found access token is not returned either. 

Reference: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/internal/storage/storage.go#L88

abhidnya13:
Resolves #106 

This changes the way we do cache lookups when confidential client apps with no account are performing acquire token silent. 

Before this change, we would ALWAYS look for an access token, refresh token, id token and account in the cache. Meaning even if one of them is not found we return an error which mean silent call failed. This works for public client applications with an user identity meaning an account present, 

However, for confidential client scenarios with no user identity (no account), the only thing that gets populated in the cache is the access token. 
There is no ID token or account because there is no user identity. There is no refresh token as well, as the service does not send one for this use case. Reasoning is in this stackoverflow answer by @rayluo :  https://stackoverflow.com/a/43150260/728675

### DIFF
--- a/apps/internal/base/internal/storage/storage.go
+++ b/apps/internal/base/internal/storage/storage.go
@@ -86,7 +86,7 @@ func isMatchingScopes(scopesOne []string, scopesTwo string) bool {
 }
 
 // Read reads a storage token from the cache if it exists.
-func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams) (TokenResponse, error) {
+func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams, inputAccount shared.Account) (TokenResponse, error) {
 	homeAccountID := authParameters.HomeaccountID
 	realm := authParameters.AuthorityInfo.Tenant
 	clientID := authParameters.ClientID
@@ -106,30 +106,39 @@ func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams)
 		return TokenResponse{}, err
 	}
 
-	idToken, err := m.readIDToken(homeAccountID, metadata.Aliases, realm, clientID)
-	if err != nil {
-		return TokenResponse{}, err
-	}
+	if !inputAccount.IsZero() {
 
-	AppMetaData, err := m.readAppMetaData(metadata.Aliases, clientID)
-	if err != nil {
-		return TokenResponse{}, err
-	}
-	familyID := AppMetaData.FamilyID
+		idToken, err := m.readIDToken(homeAccountID, metadata.Aliases, realm, clientID)
+		if err != nil {
+			return TokenResponse{}, err
+		}
 
-	refreshToken, err := m.readRefreshToken(homeAccountID, metadata.Aliases, familyID, clientID)
-	if err != nil {
-		return TokenResponse{}, err
-	}
-	account, err := m.readAccount(homeAccountID, metadata.Aliases, realm)
-	if err != nil {
-		return TokenResponse{}, err
+		AppMetaData, err := m.readAppMetaData(metadata.Aliases, clientID)
+		if err != nil {
+			return TokenResponse{}, err
+		}
+		familyID := AppMetaData.FamilyID
+
+		refreshToken, err := m.readRefreshToken(homeAccountID, metadata.Aliases, familyID, clientID)
+		if err != nil {
+			return TokenResponse{}, err
+		}
+		account, err := m.readAccount(homeAccountID, metadata.Aliases, realm)
+		if err != nil {
+			return TokenResponse{}, err
+		}
+		return TokenResponse{
+			AccessToken:  accessToken,
+			RefreshToken: refreshToken,
+			IDToken:      idToken,
+			Account:      account,
+		}, nil
 	}
 	return TokenResponse{
 		AccessToken:  accessToken,
-		RefreshToken: refreshToken,
-		IDToken:      idToken,
-		Account:      account,
+		RefreshToken: accesstokens.RefreshToken{},
+		IDToken:      IDToken{},
+		Account:      shared.Account{},
 	}, nil
 }
 
--- a/apps/internal/base/internal/storage/storage.go
+++ b/apps/internal/base/internal/storage/storage.go
@@ -86,7 +86,7 @@ func isMatchingScopes(scopesOne []string, scopesTwo string) bool {
 }
 
 // Read reads a storage token from the cache if it exists.
-func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams, inputAccount shared.Account) (TokenResponse, error) {
+func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams, account shared.Account) (TokenResponse, error) {
 	homeAccountID := authParameters.HomeaccountID
 	realm := authParameters.AuthorityInfo.Tenant
 	clientID := authParameters.ClientID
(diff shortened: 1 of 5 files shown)


apps/internal/base/internal/storage/storage.go
```diff
@@ -106,30 +106,39 @@ func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams)
 		return TokenResponse{}, err
 	}
 
-	idToken, err := m.readIDToken(homeAccountID, metadata.Aliases, realm, clientID)
-	if err != nil {
-		return TokenResponse{}, err
-	}
+	if !inputAccount.IsZero() {
```
element-of-surprise, line 109: I'd reverse the IsZero() check:

```go
if inputAccount.IsZero() {
    return TokenResponse{
		AccessToken:  accessToken,
		RefreshToken: accesstokens.RefreshToken{},
		IDToken:      IDToken{},
		Account:      shared.Account{},
	}, nil
}

File: apps/internal/shared/shared.go [hunk 2]
```diff
@@ -47,5 +47,38 @@ func (acc Account) Key() string {
 	return strings.Join([]string{acc.HomeAccountID, acc.Environment, acc.Realm}, CacheKeySeparator)
 }
 
+//IsZero checks the zero value of account
```
element-of-surprise, line 50: Can't be //IsZero, must be // IsZero
Must end with a period.

apps/internal/shared/shared.go
```diff
@@ -47,5 +47,38 @@ func (acc Account) Key() string {
...
+func (acc Account) IsZero() bool {
```
element-of-surprise, line 51: This is getting long in the tooth.  I'm going to suggest using reflection here.

Here is the example in the playground: https://play.golang.org/p/AyFv-rmIjG4
```go
func (a Account) IsZero() bool {
	v := reflect.ValueOf(a)
	for i := 0; i < v.NumField(); i++ {
		field := v.Field(i)
		if !field.IsZero() {
			switch field.Kind() {
			case reflect.Map, reflect.Slice:
				if field.Len() == 0 {
					continue
				}
			}
			return false
		}
	}
	return true
}
```

This can be made even shorter, but a zero length AdditionalFields instead of nil will cause it to think it is not zero. 
It depends on how diligent you are on keeping it nil:

```go
func (a Account) IsZero() bool {
	return reflect.ValueOf(a).IsZero()
}
```

apps/tests/integration/integration_test.go
```diff
@@ -215,3 +215,36 @@ func TestUsernamePassword(t *testing.T) {
...
+	if err != nil {
+		panic(errors.Verbose(err))
+	}
+	scopes := []string{msIDlabDefaultScope}
+	result, err := app.AcquireTokenByCredential(context.Background(), scopes)
+	if err != nil {
+		t.Fatalf("TestConfidentialClientwithSecret: on AcquireTokenByCredential(): got err == %s, want err == nil", errors.Verbose(err))
+	}
+	if result.AccessToken == "" {
+		t.Fatal("TestConfidentialClientwithSecret: on AcquireTokenByCredential(): got AccessToken == '', want AccessToken == non-empty string")
```
element-of-surprise, line 240: want AccessToken != ""

apps/tests/integration/integration_test.go
```diff
@@ -215,3 +215,36 @@ func TestUsernamePassword(t *testing.T) {
...
+	}
+	silentResult, err := app.AcquireTokenSilent(context.Background(), scopes)
+	if err != nil {
+		t.Fatalf("TestConfidentialClientwithSecret: on AcquireTokenSilent(): got err == %s, want err == nil", errors.Verbose(err))
+	}
+	if silentResult.AccessToken == "" {
+		t.Fatal("TestConfidentialClientwithSecret: on AcquireTokenSilent(): got AccessToken == '', want AccessToken == non-empty string")
```
element-of-surprise, line 247: Same

apps/internal/base/internal/storage/storage.go
```diff
@@ -121,7 +129,7 @@ func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams)
 	if err != nil {
 		return TokenResponse{}, err
 	}
-	account, err := m.readAccount(homeAccountID, metadata.Aliases, realm)
+	account, err = m.readAccount(homeAccountID, metadata.Aliases, realm)
```
jhendrixMSFT, line 132: Will the value of `account` be different after this call?

apps/internal/shared/shared.go
abhidnya13: I saw there were two implementations of IsZero() in the codebase :
[this](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/1a6b20feab85f7e1afc14529ee38962c97b6ee1f/apps/internal/oauth/ops/accesstokens/tokens.go#L83) and [this](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/35b0b9bbf5df09a3154c841144a91822b1b49c4a/apps/internal/base/internal/storage/items.go#L154)
 I wasn’t sure if we should use the reflect thing for IsZero() or a switch case. Had a discussion with @keegan-caruso  that using reflection might get slower but the switch case can cause maintenance issues. But since account object is pretty standard meaning it is unlikely to change, we decided to go with the faster one - switch case. 
 
 What are your thought ? @element-of-surprise 
 
 I have changed it to the one you suggested in the latest commit to see how long the integration tests take to run with this approach and looks like the time it takes is pretty similar to the one with a switch case. So may be using the reflection would probably be better.

apps/internal/base/internal/storage/storage.go
abhidnya13: If it finds an account, then it will and should be the same as the input account.
If an err is returned, this would be a blank `shared.Account{}` object

apps/internal/shared/shared.go
keegan-caruso: the difference in speed will be much less than measurable at the integration test level. ms at most what I would predict, but would have to benchmark it to actually know.

This is more simple.

element-of-surprise: My thoughts are the speed is inconsequential.  You could benchmark it but I doubt it is worth it. 

I like the maintenance friendly version and I would probably just generalize this and use it everywhere.

keegan-caruso: 👍

abhidnya13: @element-of-surprise , @keegan-caruso - Created PR #177 to change it in the rest of the code. This already uses the one suggested by John.

apps/internal/base/internal/storage/storage.go
jhendrixMSFT: I see.  Since we're now accepting `Account` as a parameter do we even need to call `m.readAccount()`?  Seems redundant now.

henrik-me:
We should not have to worry about accounts in confidential client scenarios. seems like an obstacle we should hide away in the public interface, not something we do in other libraries but here we have an opportunity to do better (it will break everything else)

apps/internal/base/internal/storage/storage.go
abhidnya13: I think you are right but the readAccount() function performs a few more checks to verify it is the correct account exists. I see there is a homeaccount id check which should always be true, because that is extracted from account in the previous step. However there is a realm check and environment aliases check which is from the initial configuration. 

I can dig a little more later to check if this is really redundant. Does that work?

jhendrixMSFT: Yep!

abhidnya13: Tracking [here ](https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/179)

henrik-me (approved):
:shipit:

jhendrixMSFT (approved):

PR #176, opened 2021-02-04
Fixing Jwt decoding

### LINKED ISSUE #170: [TODO]Investigate JWT decoding
Investigate JWT decoding [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/1a6b20feab85f7e1afc14529ee38962c97b6ee1f/apps/internal/oauth/ops/accesstokens/tokens.go#L233)

abhidnya13:
Looked up the spec of JWT. It mentions this [here](https://tools.ietf.org/html/rfc7519#section-3):
``` 
A JWT is represented as a sequence of URL-safe parts separated by
period ('.') characters.  Each part contains a base64url-encoded
value.
```
The definition of base64url-encoded is borrowed by the JWT spec from this [spec](https://www.rfc-editor.org/rfc/rfc7515.html#section-2):
```
Base64 encoding using the URL- and filename-safe character set
defined in Section 5 of RFC 4648 [RFC4648], with all trailing '='
characters omitted (as permitted by Section 3.2) and without the
inclusion of any line breaks, whitespace, or other additional
characters.  Note that the base64url encoding of the empty octet
sequence is the empty string.  (See Appendix C for notes on
implementing base64url encoding without padding.)
```

Finally this [Appendix](https://www.rfc-editor.org/rfc/rfc7515.html#appendix-C) mentions how to decode the JWT.

Looks like the implementation is a little bit different from what we adopted from the stackoverflow answer and MSAL Python.
Also changed the encoding from `StdEncoding` to `URLEncoding` based on the above spec.

Resolves #170

### DIFF
--- a/apps/internal/oauth/ops/accesstokens/tokens.go
+++ b/apps/internal/oauth/ops/accesstokens/tokens.go
@@ -230,14 +230,19 @@ func findDeclinedScopes(requestedScopes []string, grantedScopes []string) []stri
 
 // decodeJWT decodes a JWT and converts it to a byte array representing a JSON object
 // Adapted from MSAL Python and https://stackoverflow.com/a/31971780 .
-// TODO(msal): This looks suspect. I know that JWT has headers and payloads base64 encoded.
-// This looks to be doing a decode of base64 JSON with padding?? Is all these uses really
-// JWT??
+// JWT has headers and payload base64url encoded without padding.
+// https://tools.ietf.org/html/rfc7519#section-3 and
+// https://tools.ietf.org/html/rfc7515#section-2
 func decodeJWT(data string) ([]byte, error) {
+	// The following code adds padding based on
+	// https://tools.ietf.org/html/rfc7515#appendix-C
 	if i := len(data) % 4; i != 0 {
+		if i == 1 {
+			return nil, errors.New("Illegal base64url string")
+		}
 		data += strings.Repeat("=", 4-i)
 	}
-	return base64.StdEncoding.DecodeString(data)
+	return base64.URLEncoding.DecodeString(data)
 }
 
 // RefreshToken is the JSON representation of a MSAL refresh token for encoding to storage.
--- a/apps/internal/oauth/ops/accesstokens/tokens.go
+++ b/apps/internal/oauth/ops/accesstokens/tokens.go
@@ -229,8 +229,7 @@ func findDeclinedScopes(requestedScopes []string, grantedScopes []string) []stri
 }
 
 // decodeJWT decodes a JWT and converts it to a byte array representing a JSON object
-// Adapted from MSAL Python and https://stackoverflow.com/a/31971780 .
-// JWT has headers and payload base64url encoded without padding.
+// JWT has headers and payload base64url encoded without padding
 // https://tools.ietf.org/html/rfc7519#section-3 and
 // https://tools.ietf.org/html/rfc7515#section-2
 func decodeJWT(data string) ([]byte, error) {
--- a/apps/internal/oauth/ops/accesstokens/tokens.go
+++ b/apps/internal/oauth/ops/accesstokens/tokens.go
@@ -233,15 +233,8 @@ func findDeclinedScopes(requestedScopes []string, grantedScopes []string) []stri
 // https://tools.ietf.org/html/rfc7519#section-3 and
 // https://tools.ietf.org/html/rfc7515#section-2
 func decodeJWT(data string) ([]byte, error) {
-	// The following code adds padding based on
 	// https://tools.ietf.org/html/rfc7515#appendix-C
-	if i := len(data) % 4; i != 0 {
-		if i == 1 {
-			return nil, errors.New("Illegal base64url string")
-		}
-		data += strings.Repeat("=", 4-i)
-	}
-	return base64.URLEncoding.DecodeString(data)
+	return base64.RawURLEncoding.DecodeString(data)
 }
 
 // RefreshToken is the JSON representation of a MSAL refresh token for encoding to storage.
--- a/apps/internal/oauth/ops/accesstokens/accesstokens_test.go
+++ b/apps/internal/oauth/ops/accesstokens/accesstokens_test.go
@@ -707,8 +707,8 @@ func TestAccessTokenFromSamlGrant(t *testing.T) {
 }
 
 func TestDecodeJWT(t *testing.T) {
-	encodedStr := "aGVsbG8"
-	expectedStr := []byte("hello")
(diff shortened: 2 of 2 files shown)


apps/internal/oauth/ops/accesstokens/tokens.go
```diff
@@ -229,15 +229,19 @@ func findDeclinedScopes(requestedScopes []string, grantedScopes []string) []stri
...
 // decodeJWT decodes a JWT and converts it to a byte array representing a JSON object
-// Adapted from MSAL Python and https://stackoverflow.com/a/31971780 .
-// TODO(msal): This looks suspect. I know that JWT has headers and payloads base64 encoded.
-// This looks to be doing a decode of base64 JSON with padding?? Is all these uses really
-// JWT??
+// JWT has headers and payload base64url encoded without padding
+// https://tools.ietf.org/html/rfc7519#section-3 and
+// https://tools.ietf.org/html/rfc7515#section-2
 func decodeJWT(data string) ([]byte, error) {
+	// The following code adds padding based on
```
element-of-surprise, line 236: This seems wrong to me, if I'm reading this right (I might not be), and this probably needs someone who has done this before:

```
    // The example
    static byte [] base64urldecode(string arg)
     {
       string s = arg;
       s = s.Replace('-', '+'); // 62nd char of encoding
       s = s.Replace('_', '/'); // 63rd char of encoding
       switch (s.Length % 4) // Pad with trailing '='s
       {
         case 0: break; // No pad chars in this case
         case 2: s += "=="; break; // Two pad chars
         case 3: s += "="; break; // One pad char
         default: throw new System.Exception(
           "Illegal base64url string!");
       }
       return Convert.FromBase64String(s); // Standard base64 decoder
     }
```

This seems based on base64.StdEncoding.  Also seems weird that their examples for not having padding it to have padding!??  I hate RFCs.

But it seems like there are two options in Golang:
- RawURLEncoding is the unpadded alternate base64 encoding defined in RFC 4648. It is typically used in URLs and file names. This is the same as URLEncoding but omits padding characters.
- URLEncoding is the alternate base64 encoding defined in RFC 4648. It is typically used in URLs and file names.

I would think that since it says we should NOT have padding (even though we add it), that maybe we should just do RawURLEncoding and get rid of:
```go
if i := len(data) % 4; i != 0 {
...
```

?????

But I'd think one of these is taking care of it for you.

abhidnya13: Haha, Spent a lot of time to understand what the spec says. So here is what I think it is saying :
We have a JWT which is base64 url encoded and without any padding. This is our input string to the decode function. 

I believe what this example is doing is it is trying to show us what all steps need to be taken to be able to convert this base64 url encoded and unpadded JWT to a string which is 
1. base64 encoded (not url safe):

```
s = s.Replace('-', '+'); // 62nd char of encoding
s = s.Replace('_', '/'); // 63rd char of encoding
```

2. Padded. 
```
switch (s.Length % 4) // Pad with trailing '='s
       {
         case 0: break; // No pad chars in this case
         case 2: s += "=="; break; // Two pad chars
         case 3: s += "="; break; // One pad char
         default: throw new System.Exception(
           "Illegal base64url string!");
       }
```

After this conversion now a standard base64 decoder can be used to decode the string.

Now in our case, 
The first part is not present because we use `URLEncoding` instead of `StdEncoding` which should take care of the replace that is needed. 

The second step is what we have coded manually. As you mentioned Go has `RawURLEncoding` which can be used to replace this manual code. 
```
if i := len(data) % 4; i != 0 {
...
}
```
And that is correct. However this `RawURLEncoding` was apparently introduced in Go 1.5 [here](https://golang.org/doc/go1.5#minor_library_changes). So anybody using less than will be in trouble. But thats too old of a version so maybe we can just make the change to use `RawURLEncoding`. What do you think ?

element-of-surprise: To be fair, I think I get about half of that explanation (no fault of yours).  Again, I think the result needs to be checked by someone other than me and a test with good test data should be added.

Now, if RawURLEncoding will do what is needed it can be used.  We have a go.mod file that requires go 1.15 or higher: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/go.mod

abhidnya13: Switched to use RawURLEncoding. Also the test data I have used is from the RFC. I think its a good representative data. The test case is failing for the implementation we currently have in the dev branch.

abhidnya13: @rayluo , @bgavrilMS Can you help review this ?

henrik-me: @abhidnya13 I recommend you double check this with @brentschmaltz and IdentityModel/.NET implementations.

abhidnya13: I have asked @brentschmaltz  to review this. Also went and looked at [.Net implementations](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/320776d9f5242c18ef5258e144b03180044672a9/src/client/Microsoft.Identity.Client/Utils/Base64UrlHelpers.cs#L45) and it looks like what is mentioned in the RFC which is what this is doing. Will need someone to review it and give it a go.

henrik-me (approved):
:shipit:

apps/internal/oauth/ops/accesstokens/tokens.go
rayluo: > @rayluo , @bgavrilMS Can you help review this ?

Yes, the [JWT is encoded by a padding-less, URL-safe, base64. And, because Python's built-in `urlsafe_b64decode()` does not handle padding-less input, we had to compensate that](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.9.0/msal/oauth2cli/oidc.py#L12-L28), as @abhidnya13 correctly pointed out in her PR description.

[This PR's implementation uses Go's `base64.RawURLEncoding.DecodeString(data)`](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/176/files#diff-15e3507e1ab5a26ecef3771be96b4a2d47219077d1d87e965f23e08951f227f3R237). If that can tolerate padding-less input, that we'd be fine. I'll leave that to you folks.

bgavrilMS: Changes look good to me. If you want to write more tests I'm happy to help run them through the .NET implementation which was copied directly from the spec.

Key thing here is that "base64" and "base64-url" encodings are quite different things.

bgavrilMS (approved):

PR #178, opened 2021-02-09
Parses device code response and checks error returned

### LINKED ISSUE #169: [TODO] Investigate authorization pending in device code flow
Investigate authorization pending/ slow down in device code flow message [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/1a6b20feab85f7e1afc14529ee38962c97b6ee1f/apps/internal/oauth/oauth.go#L196)

abhidnya13:
Resolves #169 

The error returned back from service is a structured error response with status code of 400 and an error field in the json body of response. This error field can either be `authorization_pending` or `slow_down` which means we keep polling.

This is mentioned in this spec here : https://tools.ietf.org/html/rfc8628#section-3.5 which also points to this rfc for error response structure : https://tools.ietf.org/html/rfc6749#section-5.2

Also added Unwrap, Is and As to the errors file. @element-of-surprise can you review this ? Not sure the correct Go way of doing it.


apps/errors/errors.go
```diff
@@ -75,3 +76,19 @@ func (e CallErr) Verbose() string {
...
 	return fmt.Sprintf("%s:\nRequest:\n%s\nResponse:\n%s", e.Err, prettyConf.Sprint(e.Req), prettyConf.Sprint(e.Resp))
 }
+
+// Unwrap returns the wrapped error
+func (e CallErr) Unwrap() error {
+	return e.Err
+}
+
+// Is reports whether any error in errors chain matches target.
+func Is(err, target error) bool {
```
jhendrixMSFT, line 86: I don't think we need this func or the As() func below.  They just appear to redefine the ones from the standard lib.

abhidnya13: The reason it was added is , if we are trying to use this like its used here : https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/178/files#diff-a92c747c9d32f8b8080b85c19048958628dae0c99f59cc4f92e0262572ffa4f4R203 , we will need two imports , one is the standard errors package and the other one will be ours exposing CallError and we will have to name either one to differentiate. I believe customers trying to do this would also face similar issues. So if we can expose it as a part of our errors package, it will be easier

apps/errors/errors.go
```diff
@@ -75,3 +76,19 @@ func (e CallErr) Verbose() string {
...
+	return errors.Is(err, target)
+}
+// As finds the first error in errors chain that matches target,
+// and if so, sets target to that error value and returns true.
+// Otherwise, it returns false.
+func As(err, target error) bool {
```
jhendrixMSFT, line 92: Here `target` should be an `interface{}`

apps/errors/errors.go
jhendrixMSFT: Oh yes I remember this discussion now.

apps/errors/errors.go
```diff
@@ -75,3 +76,19 @@ func (e CallErr) Verbose() string {
 	e.Resp.TLS = nil     // Same
 	return fmt.Sprintf("%s:\nRequest:\n%s\nResponse:\n%s", e.Err, prettyConf.Sprint(e.Req), prettyConf.Sprint(e.Resp))
 }
+
+// Unwrap returns the wrapped error
+func (e CallErr) Unwrap() error {
```
jhendrixMSFT, line 81: I think we should remove this for now, we can always add it later if needed (removing is much harder).  Having an Unwrap() means we could be leaking the inner error as part of our method contract.

See https://blog.golang.org/go1.13-errors, section "Whether to Wrap".

apps/internal/oauth/oauth.go
```diff
@@ -194,14 +195,28 @@ func (d DeviceCode) Token(ctx context.Context) (accesstokens.TokenResponse, erro
 	}
 }
 
-var waitRE = regexp.MustCompile("(authorization_pending|slow_down)")
+type DeviceCodeError struct {
```
jhendrixMSFT, line 198: Do we expect callers outside the `oauth` package to need this type?  If not then should not be exported.

apps/internal/oauth/oauth.go
```diff
@@ -194,14 +195,28 @@ func (d DeviceCode) Token(ctx context.Context) (accesstokens.TokenResponse, erro
...
 
-// TODO(msal): This is freaking terrible. The original just looked for the exact word in the error output.
-// I doubt this worked. I don't know if the service really does this, but it should send back a structured
-// error response. Anyways, I updated this to search the entire return error message, which will be the body
-// of the return.
 func isWaitDeviceCodeErr(err error) bool {
-	return waitRE.MatchString(err.Error())
+	var c errors.CallErr
+	if errors.As(err, &c) {
+		if c.Resp.StatusCode == 400 {
```
jhendrixMSFT, line 205: Lot of nesting here.  While this is mostly stylistic, most Go programs flip the conditions to avoid the nesting (some people call this the "happy line").

apps/errors/errors.go
abhidnya13: Fixed

apps/errors/errors.go
abhidnya13: Removed in this commit https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/178/commits/5fbd566d1d01afddf610abb63fa29e05fff314ce

apps/internal/oauth/oauth.go
abhidnya13: Changed it to not be exported in this commit https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/178/commits/5fbd566d1d01afddf610abb63fa29e05fff314ce

apps/internal/oauth/oauth.go
abhidnya13: Removed nesting in this commit https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/178/commits/5fbd566d1d01afddf610abb63fa29e05fff314ce

apps/internal/oauth/oauth.go
```diff
@@ -194,14 +195,31 @@ func (d DeviceCode) Token(ctx context.Context) (accesstokens.TokenResponse, erro
 	}
 }
 
-var waitRE = regexp.MustCompile("(authorization_pending|slow_down)")
+type deviceCodeError struct {
```
jhendrixMSFT, line 198: Don't we receive a standard, structured error response for all failed requests?  Assuming that's the case, we should have a standard error type so we don't have to create these one-off types.  In `azidentity` we called this [`AADAuthenticationFailedError `](https://github.com/Azure/azure-sdk-for-go/blob/master/sdk/azidentity/azidentity.go#L45).  We can leave that for a separate PR though.

apps/internal/oauth/oauth.go
```diff
@@ -194,14 +195,31 @@ func (d DeviceCode) Token(ctx context.Context) (accesstokens.TokenResponse, erro
...
+		defer c.Resp.Body.Close()
+		body, err := ioutil.ReadAll(c.Resp.Body)
+		if err != nil {
+			return false
+		}
+		err = json.Unmarshal(body, &dCErr)
+		if err != nil {
+			return false
+		}
+		if dCErr.Error == "authorization_pending" || dCErr.Error == "slow_down" {
```
jhendrixMSFT, line 218: Has `slow_down` been removed?  I can't find it doc'ed https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-device-code

jhendrixMSFT: (we do this in our track 1 ADAL lib, I have no problem with having this check, more of a curiosity why it isn't documented)

abhidnya13: I think I see it in the official spec : https://tools.ietf.org/html/rfc8628#section-3.5 , there probably needs an update to the microsoft docs after verifiying it has not been removed

apps/internal/oauth/oauth.go
```diff
@@ -194,14 +195,31 @@ func (d DeviceCode) Token(ctx context.Context) (accesstokens.TokenResponse, erro
...
+}
 
-// TODO(msal): This is freaking terrible. The original just looked for the exact word in the error output.
-// I doubt this worked. I don't know if the service really does this, but it should send back a structured
-// error response. Anyways, I updated this to search the entire return error message, which will be the body
-// of the return.
 func isWaitDeviceCodeErr(err error) bool {
-	return waitRE.MatchString(err.Error())
+	var c errors.CallErr
+	if errors.As(err, &c) {
```
jhendrixMSFT, line 204: Nit: I'd flip this condition to remove the nesting.

apps/internal/oauth/ops/internal/comm/comm.go
```diff
@@ -255,6 +255,7 @@ func (c *Client) do(ctx context.Context, req *http.Request) ([]byte, error) {
 	switch reply.StatusCode {
 	case 200, 201:
 	default:
+		reply.Body = ioutil.NopCloser(bytes.NewBuffer(data))
```
jhendrixMSFT, line 258: Can this be moved to before the `switch` statement?  If this block is ever refactored it would be easy to accidentally miss setting `reply.Body`.

apps/internal/oauth/oauth.go
abhidnya13: I had this conversation with @element-of-surprise while doing the error design doc, of having a standard error type for the server response, but since in most cases the error message can sufficiently tell what needs to change, we decided to not go for it and instead have a `CallErr` so that in the case they need to use something they can get the response body.

jhendrixMSFT: If this is the only time we need to unmarshall the error info to check some value then we can leave it as-is.  I just want to ensure we're not adding types that essentially do the same thing because there's no shared definition.

jhendrixMSFT (approved):

apps/internal/oauth/oauth.go
abhidnya13: For now, this is the only place we need to unmarshall to check some value. There might be some features where we need to unmarshall the error to provide specific error info to the app developer. For that this is what we plan to do, quoting from the error design doc
``` If further differentiation is required, we can add custom errors that use Go error wrapping on top of CallErr to achieve our diagnostic goals (such as detecting when to retry a call due to transient errors).``` 

I guess, we can add a type even internally if such a thing is needed in more cases.

apps/internal/oauth/oauth.go
abhidnya13: Resolved in this commit : https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/178/commits/da3abaa86143db33a44497c61f65f67cc79e6872

apps/internal/oauth/ops/internal/comm/comm.go
abhidnya13: Resolved in this commit : https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/178/commits/da3abaa86143db33a44497c61f65f67cc79e6872

PR #180, opened 2021-02-10
add tests for confidential client

jhendrixMSFT:


apps/confidential/confidential_test.go
```diff
@@ -66,3 +75,93 @@ lQFhpxvPgPNBuxVRnsxoH/sLOA==
...
+func fakeClient() (Client, error) {
+	cred, err := NewCredFromSecret("fake_secret")
+	if err != nil {
+		return Client{}, err
+	}
+	client, err := New("fake_client_id", cred, WithAuthority("https://fake_authority/fake"))
+	if err != nil {
+		return Client{}, err
+	}
+	client.base.Token.AccessTokens = &fake.AccessTokens{
```
jhendrixMSFT, line 88: Note that I faked the min set of fields in all these responses to get things working.

jhendrixMSFT: This should be better.

apps/confidential/confidential_test.go
```diff
@@ -66,3 +75,105 @@ lQFhpxvPgPNBuxVRnsxoH/sLOA==
...
+	tk, err = client.AcquireTokenSilent(context.Background(), tokenScope)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if tk.AccessToken != token {
+		t.Fatalf("unexpected access token %s", tk.AccessToken)
+	}
+}
+
+func TestAcquireTokenByAuthCode(t *testing.T) {
```
abhidnya13, line 154: This test will probably be a little different than s2s scenario. AcquireTokenByAuthCode in confidential client is the second step of Authorization code flow. 

In real use case scenario, it probably is a webapp which has done the first step of acquiring an auth code on its own.  And then passes that auth code to this function. That also makes me rethink that code cannot be an optional parameter here. We probably need to review this. Also this means that it can have a user present too.

abhidnya13 (commented):
I think we need to review this API. Added some comments.

apps/confidential/confidential_test.go
jhendrixMSFT: Yes agreed, the code should not be optional here.  It's not in the [C# impl](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/IConfidentialClientApplication.cs#L44).  I will make it a required param.

jhendrixMSFT: Fixed

apps/confidential/confidential_test.go
```diff
@@ -66,3 +75,105 @@ lQFhpxvPgPNBuxVRnsxoH/sLOA==
...
+		t.Fatal("unexpected nil error from AcquireTokenSilent")
+	}
+	tk, err := client.AcquireTokenByAuthCode(context.Background(), "fake_auth_code", tokenScope)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if tk.AccessToken != token {
+		t.Fatalf("unexpected access token %s", tk.AccessToken)
+	}
+	// second attempt should return the cached token
```
abhidnya13, line 171: Since we are doing auth code flow, there will be a user as well. So we would need to pick up an account and do an acquire_token_silent with account

jhendrixMSFT: I see.  So two questions.  How do I create the account (do I just use the result from `client.Accounts()` and should I be worried that the call succeeded without an account?

abhidnya13: Yeah, client.Accounts() should work. I am still thinking on the second question. I will have to take a closer look.

jhendrixMSFT: So it seems like the account isn't looked at when reading a token from the cache, presumably due to [this](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/base/internal/storage/storage.go#L109).

Unfortunately `Accounts()` returns a `nil` slice.  It's because my fake access token doesn't set the `IDToken` field so as a result [this](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/base/internal/storage/storage.go#L191) check doesn't happen and no account is added to the cache.

abhidnya13: Yeah Ideally, there should be `IDToken` returned as well if there is a user. So there is an Account in the cache as well and then that would mean Accounts() wont return a nil.
I believe in that case doing an acquireTokenSilent() with no account will also not give an anything from the cache. - Answering your second question - should I be worried that the call succeeded without an account?

jhendrixMSFT: That doesn't seem to be the case, at least for this test.  i.e. having no account returns the cached token.  Presumably this is just triggering the same code path as a token acquired from a secret.  I will update the test to form a proper token response for a user.

jhendrixMSFT: This has been fixed and also uncovered a bug which has also been fixed.

abhidnya13 (approved):

PR #181, opened 2021-02-11
Authority validation todos

### LINKED ISSUE #160: [TODO]Clean-up authority validation
Refer [this](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/1a6b20feab85f7e1afc14529ee38962c97b6ee1f/apps/internal/oauth/ops/authority/authority.go)

abhidnya13:
Resolves #160

### DIFF
--- a/apps/internal/oauth/ops/authority/authority.go
+++ b/apps/internal/oauth/ops/authority/authority.go
@@ -108,6 +108,12 @@ const (
 	ATRefreshToken
 )
 
+// These are all authority types
+const (
+	AAD  = "MSSTS"
+	ADFS = "ADFS"
+)
+
 // AuthParams represents the parameters used for authorization for token acquisition.
 type AuthParams struct {
 	AuthorityInfo Info
@@ -177,9 +183,9 @@ func NewInfoFromAuthorityURI(authorityURI string, validateAuthority bool) (Info,
 
 	tenant, err := firstPathSegment(u)
 	if tenant == "adfs" {
-		authorityType = "ADFS"
+		authorityType = ADFS
 	} else {
-		authorityType = "MSSTS"
+		authorityType = AAD
 	}
 
 	if err != nil {
--- a/apps/internal/oauth/ops/authority/authority.go
+++ b/apps/internal/oauth/ops/authority/authority.go
@@ -166,48 +166,7 @@ func firstPathSegment(u *url.URL) (string, error) {
 // NewInfoFromAuthorityURI creates an AuthorityInfo instance from the authority URL provided.
 func NewInfoFromAuthorityURI(authorityURI string, validateAuthority bool) (Info, error) {
 	authorityURI = strings.ToLower(authorityURI)
-
-	// TODO(msal): The original code I found had a validateAuthority() function.
-	// But the function just returned error == nil. I took this from the python and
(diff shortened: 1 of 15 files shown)


apps/internal/oauth/ops/authority/authority.go
```diff
@@ -166,48 +166,7 @@ func firstPathSegment(u *url.URL) (string, error) {
...
-		                "Unable to get authority configuration for {}. "
-		                "Authority would typically be in a format of "
-		                "https://login.microsoftonline.com/your_tenant_name".format(
-						authority_url))
-	*/
-
-	// todo: check for other authority types...
-	authorityType := "MSSTS"
-
+	var authorityType string
```
abhidnya13, line 169: We do this in `resolvers.go`  [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/oauth/resolvers.go#L48)
We delay doing this here because we want to avoid making a network call unless absolutely needed. So this code wont run while initializing app object here.

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -307,8 +272,6 @@ func (c Client) UserRealm(ctx context.Context, authParams AuthParams) (UserRealm
 	err := c.Comm.JSONCall(
 		ctx,
 		endpoint,
-		// TODO(jdoak): not thrilled about this, because all calls should have this but
```
abhidnya13, line 310: I think we add Correlation Id to all the calls that are part of a user flow for telemetry. I see thats what we do in MSAL Python. It was added here : https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/103 
Could not find a spec though. @henrik-me thoughts ?

apps/internal/oauth/resolvers.go
```diff
@@ -123,13 +123,15 @@ func (m *authorityEndpoint) addCachedEndpoints(authorityInfo authority.Info, use
...
 func (m *authorityEndpoint) openIDConfigurationEndpoint(ctx context.Context, authorityInfo authority.Info, userPrincipalName string) (string, error) {
-	if authorityInfo.ValidateAuthority && !authority.TrustedHost(authorityInfo.Host) {
+	if authorityInfo.Tenant != "adfs" && authorityInfo.ValidateAuthority && !authority.TrustedHost(authorityInfo.Host) {
 		resp, err := m.rest.Authority().AADInstanceDiscovery(ctx, authorityInfo)
 		if err != nil {
 			return "", err
 		}
 
 		return resp.TenantDiscoveryEndpoint, nil
+	} else if authorityInfo.Tenant == "adfs" {
```
jhendrixMSFT, line 133: If you make this this first condition in the `if` block you can remove the `!=` check that's currently on line 126.

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -217,6 +176,12 @@ func NewInfoFromAuthorityURI(authorityURI string, validateAuthority bool) (Info,
 	}
 
 	tenant, err := firstPathSegment(u)
+	if tenant == "adfs" {
+		authorityType = "ADFS"
```
jhendrixMSFT, line 180: This and `MSSTS` should be constants.

jhendrixMSFT (approved):

PR #183, opened 2021-02-21
Honor the reply url in the redirectURI if specified

bdwyertech:
The interactive login process should honor the Redirect URI if specified.

Had an issue with this trying to leverage the AZ CLI client-id & application and it seems that only localhost is an authorized URL, not 127.0.0.1.

Seems to be because `net.Listener`'s  Addr() function returns 127.0.0.1 as opposed to localhost -- not sure if that is desired or not.

https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/696213fb452a5c21c892c7f87f3dda19f0cc1313/apps/internal/local/server.go#L80

In case you are curious what I am trying to do, I'd like to get rid of the need for Python Azure CLI completely for `az login` functionality.  I have an internal platform CLI tool which performs this function for AWS.  I'd like to add similar functionality for Azure.  Right now the tool just shells out to the Azure CLI, but I'd like to get rid of that dependency.

It would also help slim down Terraform containers for Azure deployments, as Terraform also has a dependency on the Azure CLI, which my understanding is purely for sourcing credentials.

### DIFF
--- a/apps/public/public.go
+++ b/apps/public/public.go
@@ -372,7 +372,11 @@ func (pca Client) browserLogin(ctx context.Context, redirectURI *url.URL, params
 		return interactiveAuthResult{}, err
 	}
 	defer srv.Shutdown()
-	authURL, err := pca.base.AuthCodeURL(ctx, params.ClientID, srv.Addr, params.Scopes, params)
+	replyUrl := srv.Addr
+	if redirectURI != nil {
+		replyUrl = redirectURI.String()
+	}
+	authURL, err := pca.base.AuthCodeURL(ctx, params.ClientID, replyUrl, params.Scopes, params)
 	if err != nil {
 		return interactiveAuthResult{}, err
 	}
@@ -387,7 +391,7 @@ func (pca Client) browserLogin(ctx context.Context, redirectURI *url.URL, params
 	}
 	return interactiveAuthResult{
 		authCode:    res.Code,
-		redirectURI: srv.Addr,
+		redirectURI: replyUrl,
 	}, nil
 }
 


abhidnya13:
Hey @bdwyertech , Thanks for the contribution. I am trying to understand your scenario, are you getting an error message for 127.0.0.1 ? I see you were trying to use AZ CLI client-id and application. However, eventually, you will have your own application right ?

bdwyertech:
@abhidnya13 I'm actually trying to create a Golang-equivalent of the `az login` functionality, e.g. the thing that generates `~/.azure/accessTokens.json`

I'm writing this to use the same application ID as the CLI for this reason.

apps/public/public.go
```diff
@@ -372,7 +372,11 @@ func (pca Client) browserLogin(ctx context.Context, redirectURI *url.URL, params
 		return interactiveAuthResult{}, err
 	}
 	defer srv.Shutdown()
-	authURL, err := pca.base.AuthCodeURL(ctx, params.ClientID, srv.Addr, params.Scopes, params)
+	replyUrl := srv.Addr
+	if redirectURI != nil {
+		replyUrl = redirectURI.String()
```
jhendrixMSFT, line 377: For interactive auth, the redirect URI is always going to be `http://localhost` (it's currently the loop-back address which we need to change).

The API takes a URI parameter as a means to specify a custom port number only (kinda clunky IMO but it's how the MSALs work).

bdwyertech:
I'd be perfectly fine with localhost as that would work as a callback to
the well-known CLI application ID.

As a side note, are there any plans to have a cleaner integration with the
azure-autorest SDK's? Currently I'm using the return from this and then
doing a few type conversions to make it work over there.  I know this is
all pre-beta, but Go is the language of choice for all our internal
platform tooling (large enterprise customer).

apps/public/public.go
bdwyertech: I'd be perfectly fine with localhost as that would work as a callback to
the well-known CLI application ID.

As a side note, are there any plans to have a cleaner integration with the
azure-autorest SDK's? Currently I'm using the return from this and then
doing a few type conversions to make it work over there.  I know this is
all pre-beta, but Go is the language of choice for all our internal
platform tooling (large enterprise customer).

jhendrixMSFT: The short answer is yes.

The long answer.  We have a new authentication package, [`azidentity`](https://github.com/Azure/azure-sdk-for-go/tree/master/sdk/azidentity) to be used to authenticate with our "track 2" (i.e. next-gen) Azure SDK for Go (the content under the [`./sdk`](https://github.com/Azure/azure-sdk-for-go/tree/master/sdk) directory contains the track 2 packages).  There's also an adapter, [`azidext`](https://github.com/jongio/azidext/tree/master/go/azidext), that you can use to adapt `azidentity` authentication with the track 1 SDK packages.  Once we have a preview of this MSAL package, we will retool the innards of `azidentity` to consume it.

jhendrixMSFT:
https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/187 replaces the loop-back address with localhost, so this should now be fixed.

apps/public/public.go
bdwyertech: Fantastic -- thank you for the insight, @jhendrixMSFT

PR #184, opened 2021-03-01
Exposing suggestedCacheKey in the cache interface for custom token cache

abhidnya13:
The PR adds a parameter named key to the cache interaface. This key can be used to partition the cache for better performance. 

After discussions, this PR also removed GetAccounts() API from confidential clients. 

We instead expose GetAccount(homeAccountId) which returns an Account assosciated with the corresponding homeAccountID

Wiki page : https://github.com/AzureAD/microsoft-authentication-library-for-go/wiki/Using-cache-key-to-partition-the-cache

CC: @jmprieur @jennyf19

### DIFF
--- a/apps/cache/cache.go
+++ b/apps/cache/cache.go
@@ -30,8 +30,8 @@ type Serializer interface {
 // ExportReplace is used export or replace what is in the cache.
 type ExportReplace interface {
 	// Replace replaces the cache with what is in external storage.
-	Replace(cache Unmarshaler)
+	Replace(cache Unmarshaler, key string)
 	// Export writes the binary representation of the cache (cache.Marshal()) to
 	// external storage. This is considered opaque.
-	Export(cache Marshaler)
+	Export(cache Marshaler, key string)
 }
--- a/apps/cache/cache.go
+++ b/apps/cache/cache.go
@@ -30,8 +30,10 @@ type Serializer interface {
 // ExportReplace is used export or replace what is in the cache.
 type ExportReplace interface {
 	// Replace replaces the cache with what is in external storage.
+	// key is the suggested key which can be used for partioning the cache
 	Replace(cache Unmarshaler, key string)
 	// Export writes the binary representation of the cache (cache.Marshal()) to
 	// external storage. This is considered opaque.
+	// key is the suggested key which can be used for partioning the cache
 	Export(cache Marshaler, key string)
 }
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -37,8 +37,8 @@ type manager interface {
 
 type noopCacheAccessor struct{}
 
-func (n noopCacheAccessor) Replace(cache cache.Unmarshaler) {}
-func (n noopCacheAccessor) Export(cache cache.Marshaler)    {}
+func (n noopCacheAccessor) Replace(cache cache.Unmarshaler, key string) {}
+func (n noopCacheAccessor) Export(cache cache.Marshaler, key string)    {}
 
 // AcquireTokenSilentParameters contains the parameters to acquire a token silently (from cache).
 type AcquireTokenSilentParameters struct {
@@ -202,8 +202,9 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	authParams.HomeaccountID = silent.Account.HomeAccountID
 
 	if s, ok := b.manager.(cache.Serializer); ok {
-		b.cacheAccessor.Replace(s)
-		defer b.cacheAccessor.Export(s)
+		suggestedCacheKey := getKeyFromRequest(authParams)
+		b.cacheAccessor.Replace(s, suggestedCacheKey)
+		defer b.cacheAccessor.Export(s, suggestedCacheKey)
 	}
 
 	storageTokenResponse, err := b.manager.Read(ctx, authParams, silent.Account)
@@ -241,6 +242,7 @@ func (b Client) AcquireTokenByAuthCode(ctx context.Context, authCodeParams Acqui
 	var cc *accesstokens.Credential
 	if authCodeParams.AppType == accesstokens.ATConfidential {
 		cc = authCodeParams.Credential
+		authParams.IsConfidentialClient = true
 	}
 
 	req, err := accesstokens.NewCodeChallengeRequest(authParams, authCodeParams.AppType, cc, authCodeParams.Code, authCodeParams.Challenge)
@@ -262,8 +264,9 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 	}
 
 	if s, ok := b.manager.(cache.Serializer); ok {
-		b.cacheAccessor.Replace(s)
-		defer b.cacheAccessor.Export(s)
+		suggestedCacheKey := getKeyFromResponse(authParams, token.ClientInfo.HomeAccountID())
+		b.cacheAccessor.Replace(s, suggestedCacheKey)
+		defer b.cacheAccessor.Export(s, suggestedCacheKey)
 	}
 
 	account, err := b.manager.Write(authParams, token)
@@ -275,8 +278,9 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 
 func (b Client) Accounts() []shared.Account {
 	if s, ok := b.manager.(cache.Serializer); ok {
-		b.cacheAccessor.Replace(s)
-		defer b.cacheAccessor.Export(s)
+		suggestedCacheKey := getKeyFromRequest(b.AuthParams)
+		b.cacheAccessor.Replace(s, suggestedCacheKey)
+		defer b.cacheAccessor.Export(s, suggestedCacheKey)
 	}
 
 	accounts, err := b.manager.AllAccounts()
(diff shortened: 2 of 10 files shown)


apps/internal/base/base.go
```diff
@@ -302,3 +306,30 @@ func convertStrUnixToUTCTime(unixTime string) (time.Time, error) {
 	}
 	return time.Unix(timeInt, 0).UTC(), nil
 }
+
+func getKeyFromRequest(authParams authority.AuthParams) string {
+	if authParams.AuthorizationType == authority.ATClientCredentials {
```
bgavrilMS, line 311: There is also the OBO case

apps/cache/cache.go
```diff
@@ -30,8 +30,8 @@ type Serializer interface {
 // ExportReplace is used export or replace what is in the cache.
 type ExportReplace interface {
 	// Replace replaces the cache with what is in external storage.
```
keegan-caruso, line 32: Comments need updated to explain what the key is, and the contract for it, when it is empty / etc

apps/internal/base/base.go
```diff
@@ -302,3 +306,30 @@ func convertStrUnixToUTCTime(unixTime string) (time.Time, error) {
...
+		return homeaccountIDfromResponse
+	}
+	return ""
+}
+
+func getAppKey(authParams authority.AuthParams) string {
+	if authParams.AuthorityInfo.Tenant != "" {
+		return fmt.Sprintf("%s_%s_AppTokenCache", authParams.ClientID, authParams.AuthorityInfo.Tenant)
+	}
+	return fmt.Sprintf("%s_%s_AppTokenCache", authParams.ClientID, "")
```
keegan-caruso, line 334: why sprintf on the second param if it is always empty?

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -133,6 +133,8 @@ type AuthParams struct {
 	CodeChallengeMethod string
 	// Prompt specifies the user prompt type during interactive auth.
 	Prompt string
+	// IsConfidentialClient specifies if it is a confidential client
+	IsConfidentialClient bool
```
keegan-caruso, line 137: is this just a proxy for AuthorizationType == accesstokens.ATConfidential?

apps/internal/base/base.go
abhidnya13: Thanks for the review Bogdan ! 
We don't support OBO flow in MSAL Go yet 
The second one is interesting, we just have GetAccounts() in the library right now, may be discuss little bit more about this one offline ? Have some questions.

apps/internal/base/base.go
```diff
@@ -302,3 +306,30 @@ func convertStrUnixToUTCTime(unixTime string) (time.Time, error) {
 	}
 	return time.Unix(timeInt, 0).UTC(), nil
 }
+
+func getKeyFromRequest(authParams authority.AuthParams) string {
```
element-of-surprise, line 310: So this looks like something that should just be attached to authority.AuthParams.  Renamed to CacheKey().

apps/internal/base/base.go
```diff
@@ -302,3 +306,30 @@ func convertStrUnixToUTCTime(unixTime string) (time.Time, error) {
...
+	if authParams.AuthorizationType == authority.ATClientCredentials {
+		return getAppKey(authParams)
+	}
+	if authParams.IsConfidentialClient || authParams.AuthorizationType == authority.ATRefreshToken {
+		return homeaccountIDfromResponse
+	}
+	return ""
+}
+
+func getAppKey(authParams authority.AuthParams) string {
```
element-of-surprise, line 330: Same, AppKey() and attach to authority.AuthParams.

apps/internal/base/base.go
```diff
@@ -302,3 +306,30 @@ func convertStrUnixToUTCTime(unixTime string) (time.Time, error) {
...
+	if authParams.AuthorizationType == authority.ATClientCredentials {
+		return getAppKey(authParams)
+	}
+	if authParams.AuthorizationType == authority.ATRefreshToken {
+		return authParams.HomeaccountID
+	}
+	return ""
+}
+
+func getKeyFromResponse(authParams authority.AuthParams, homeaccountIDfromResponse string) string {
```
element-of-surprise, line 320: So, want to note a few trends I see.  

Java names:  homeaccountIDfromResponse, homeAccountID would be fine.  
getKeyFromResponse: There are rare times you should use "get" in a name, and that is almost always because you have a name conflict.  keyFromResponse() says the same thing.  keyFromResp() is even better.

I'd think about attaching this to accesstokens.TokenResponse as CacheKey(authParams authority.AuthParams).

apps/cache/cache.go
element-of-surprise: Even better would be a link to some doc somewhere that explains this whole thing.

bgavrilMS: I'm not sure we have that, so let me try to explain here. MSAL's token caching primitives are "store this blob" and "give me the blob back". This works fine for a public client application, which typically has <10 users logged in. 

In a confidential client app, the number of users of that app is in the millions. This poses 2 problems: 

a) memory pressure, i.e. you'll run out of memory since 1 entry (AT, RT, IDT and account metadata) is around 10KB. 
b) token cache queries are O(n) and for large values of N we've seen searches taking 10+ seconds. 

So the public client caching strategy is misleading / wrong and we need to try to correct it. To solve the above problems, the standard way is to use a distributed cache (e.g. Redis) and so there is a need for a KEY. 

# Usage in web site

- web site middlerware gets you an authorization code and ID token
- developer intercepts this and retrieves the home_account_id
- create a CCA and call `GetAccount(home_account_id)`. CCA loads only the cache node associated with that `home_account_id`
- call `AcquireTokenSilent(account)` - CCA loads only the cache node associated with account.HomeAccountId
- or call `AcquireTokenByAuthCode(auth_code)` - and when the result come in, the CCA will save the tokens in based on `response.home_account_id`

# Usage in S2S

- for S2S, there is no user involved. However, there could be millions of tenants involved and the problems above are still hit.
- `AcquireTokenSilent()` with no account will load the cache in node `AppTokenCache_<tid>` and will search for a token only in that cache partition
- `AcquireTokenForClient()` will go to ESTS and get a new token. Then it will save all in the cache node `AppTokenCache_<tid>`

bgavrilMS: I also want to point out that you can redesign your cache primitives to be more like 

`Read(key): byte[]` 
`Write(key, byte[])`
`Remove(key)` 

For public client apps, the key does not really matter, it can be "PCA" or a constant. 
For confidential client apps, the key can be as per what Abhi implemented.

apps/internal/base/base.go
abhidnya13: I have exposed `GetAccount(home_accountID)` equivalent in the latest commit

abhidnya13: I have scheduled a meeting to discuss if we need GetAccounts() in MSAL Go.

apps/cache/cache.go
abhidnya13: I have added a note as to what the partitioning key is,

I can convert the explanantion above in a wiki page to help understand things better if someone wants to use the suggested Cache key.

apps/internal/base/base.go
abhidnya13: resolved in https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/184/commits/71dea81ddef28df17b4741bef3e8643d75e7ccd9

apps/internal/oauth/ops/authority/authority.go
abhidnya13: Yes it is, but we set these only in certain structs and we are extracting the information we need for the cache key mostly from authParams,

I think if we do not want it here in AuthParams, we can add it as a parameter to the CacheKey() function. What do you think ?

apps/internal/base/base.go
abhidnya13: Resolved in https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/184/commits/71dea81ddef28df17b4741bef3e8643d75e7ccd9

apps/internal/base/base.go
abhidnya13: Resoved in https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/184/commits/71dea81ddef28df17b4741bef3e8643d75e7ccd9

apps/internal/base/base.go
abhidnya13: Resoved in 71dea81

apps/internal/base/base.go
```diff
@@ -202,8 +203,13 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	authParams.HomeaccountID = silent.Account.HomeAccountID
 
 	if s, ok := b.manager.(cache.Serializer); ok {
-		b.cacheAccessor.Replace(s)
-		defer b.cacheAccessor.Export(s)
+		var isAppCache bool
+		if silent.Account.IsZero() {
+			isAppCache = true
+		}
```
jennyf19, line 209: is this only available to cca? not sure as it's in base.go

jennyf19 (approved):
:shipit:

apps/internal/base/base.go
abhidnya13: Thanks for this, moved it to confidential.go in the latest commit : https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/184/commits/37a493c54450f34ef29b42c6252c7989b05b45e4

PR #186, opened 2021-03-04
Only trigger CI/CD when a PR is labelled 

abhidnya13:
This branch changes the PR to be built only when a label has been added.
Verified this works as expected.


keegan-caruso (approved):
thanks!

PR #187, opened 2021-03-04
Use http://localhost for interactive flow

jhendrixMSFT:
Don't use the loop-back address.  Updated test to check for it.


abhidnya13 (approved):
:shipit:

PR #188, opened 2021-03-04
Make redirect URI a required parameter for auth code flow

jhendrixMSFT:
Don't hard-code the nativeclient redirect URI.
Make the auth code a parameter for the public client (the same was
already done for confidential client).


jhendrixMSFT:
Change is based on the discussion [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/96/files#r539627995).

abhidnya13 (approved):
LGTM, 

I think we should also mention may be in the comments/reference docs that this RedirectUri has to be the same as the one used in the first leg of this flow.

PR #192, opened 2021-03-17
Adding support for SubjectName/Issuer (SNI) authentication

### LINKED ISSUE #190: Add support for SubjectName/Issuer (SNI) authentication
See https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/60 for the python impl.

abhidnya13:
This PR resolves #190.

This feature can be enabled by setting WithX5C() option to true when initializing the confidential client application something like this:
``` Go
confidential.New("client_id", credential, confidential.WithAuthority("authority"), confidential.WithX5C())
```

Alternate way to expose this can be by providing this as option to the Credential object here: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/confidential/confidential.go#L132

Also, I was able to test this e2e using an MSIT cert and configuring the tenant to have MSIT as a Trusted Certificate Authority.

RFC : https://tools.ietf.org/html/rfc7515#section-4.1.6

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -187,7 +187,7 @@ type Options struct {
 	// It defaults to a shared http.Client.
 	HTTPClient ops.HTTPClient
 
-	// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS
+	// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS.
 	SendX5C bool
 }
 
@@ -227,7 +227,7 @@ func WithHTTPClient(httpClient ops.HTTPClient) Option {
 	}
 }
 
-// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS to enable Subject Name Issuer Authentication
+// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS to enable Subject Name Issuer Authentication.
 func SendX5C() Option {
 	return func(o *Options) {
 		o.SendX5C = true
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -140,8 +140,8 @@ type Credential struct {
 // code requires that client.go, requests.go and confidential.go share a credential type without
 // having import recursion. That requires the type used between is in a shared package. Therefore
 // we have this.
-func (c Credential) toInternal() *accesstokens.Credential {
-	return &accesstokens.Credential{Secret: c.secret, Cert: c.cert, Key: c.key}
+func (c Credential) toInternal(sendX5C bool) *accesstokens.Credential {
+	return &accesstokens.Credential{Secret: c.secret, Cert: c.cert, Key: c.key, SendX5C: sendX5C}
 }
 
 // NewCredFromSecret creates a Credential from a secret.
@@ -186,6 +186,9 @@ type Options struct {
 	// The HTTP client used for making requests.
 	// It defaults to a shared http.Client.
 	HTTPClient ops.HTTPClient
+
+	// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS
+	SendX5C bool
 }
 
 func (o Options) validate() error {
@@ -224,6 +227,13 @@ func WithHTTPClient(httpClient ops.HTTPClient) Option {
 	}
 }
 
+// SendX5c allows to send X5C value
+func SendX5C() Option {
+	return func(o *Options) {
+		o.SendX5C = true
+	}
+}
+
 // New is the constructor for Client. userID is the unique identifier of the user this client
 // will store credentials for (a Client is per user). clientID is the Azure clientID and cred is
 // the type of credential to use.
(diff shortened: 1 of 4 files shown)


apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -92,6 +92,9 @@ type Credential struct {
 	// Key is the private key for signing if we are doing any auth other than secret.
 	Key crypto.PrivateKey
 
+	// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS
+	SendX5C bool
```
keegan-caruso, line 96: This isn't really part of the credential. Would this make more sense to be modeled as part of AuthParams? thoughts?

apps/confidential/confidential.go
```diff
@@ -186,6 +186,9 @@ type Options struct {
 	// The HTTP client used for making requests.
 	// It defaults to a shared http.Client.
 	HTTPClient ops.HTTPClient
+
+	// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS
```
element-of-surprise, line 190: Must end in period "."

apps/confidential/confidential.go
```diff
@@ -224,6 +227,13 @@ func WithHTTPClient(httpClient ops.HTTPClient) Option {
 	}
 }
 
+// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS to enable Subject Name Issuer Authentication
```
element-of-surprise, line 230: Same period thing.

apps/confidential/confidential.go
```diff
@@ -140,8 +140,8 @@ type Credential struct {
 // code requires that client.go, requests.go and confidential.go share a credential type without
 // having import recursion. That requires the type used between is in a shared package. Therefore
 // we have this.
-func (c Credential) toInternal() *accesstokens.Credential {
-	return &accesstokens.Credential{Secret: c.secret, Cert: c.cert, Key: c.key}
+func (c Credential) toInternal(sendX5C bool) *accesstokens.Credential {
```
element-of-surprise, line 143: Nit: I think x5c would suffice over sendX5C

apps/confidential/confidential.go
element-of-surprise: As a follow up, I am certainly not an expert on all the auth methods.  But it seems to me that various users tend to see they need to do X, so they look to enable X.  x5c doesn't mean much to the user.

Keegan and yourself would know better, but I might call this:

EnableSNIAuth()

apps/internal/oauth/ops/accesstokens/accesstokens.go
element-of-surprise: I think semantically, makes more sense in AuthParams.  

My caveat to that is how hard it is to add to authParams vs where it is. If that starts to get messy, I'd just leave it.

element-of-surprise (changes requested):

apps/confidential/confidential.go
abhidnya13: I think I agree with you. EnableSNIAuth really makes more sense. Right now this is how things are done in .NET and Java. @jmprieur Is sendX5C chosen for a specific reason? Can we use EnableSNIAuuth() instead here ?

apps/internal/oauth/ops/accesstokens/accesstokens.go
abhidnya13: Added it in AuthParams in this commit https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/192/commits/77406fefb57f82aad9027449990cd080a67a7adf. Does not get that messy. So I think we can keep it in AuthParams. Please take a look.

apps/confidential/confidential.go
jmprieur: This is the result of compromises in API reviews, @abhidnya13.
It's not so much that we enable SNI Authentication: we enable the same kind of confidential client flows, but we send the public key of a subject/name issuer cert to ease the cert rotation.

apps/internal/base/base.go
```diff
@@ -130,6 +130,13 @@ func WithCacheAccessor(ca cache.ExportReplace) Option {
 	}
 }
 
+// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS to enable Subject Name Issuer Authentication
```
element-of-surprise, line 133: Still missing the period at the end of the comment.

apps/confidential/confidential.go
```diff
@@ -224,6 +227,13 @@ func WithHTTPClient(httpClient ops.HTTPClient) Option {
 	}
 }
 
+// SendX5C specifies if x5c claim(public key of the certificate) should be sent to STS to enable Subject Name Issuer Authentication.
```
element-of-surprise, line 230: We adopted a "With" prefix for options.  So this will need to be WithSendX5C().  Though, if we aren't going to do the SNI name, then WithX5C() would be better.

keegan-caruso (approved):
LGTM

jhendrixMSFT (approved):

PR #195, opened 2021-03-23
Avoid making copies of Contract

jhendrixMSFT:
Now that the cache manager is per client, there's no need to copy and
update the Contract (copying was the source of slow cache updates for a
large number of tokens).
Removed atomic implementation in favor of a RWMutex.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 *.dylib
 *.golangci.yml
 *.swp
+*.pprof
 
 # OSX specific os files
 *.DS_Store
--- a/apps/internal/base/internal/storage/storage.go
+++ b/apps/internal/base/internal/storage/storage.go
@@ -16,7 +16,6 @@ import (
 	"fmt"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/json"
@@ -46,19 +45,18 @@ type TokenResponse struct {
 // updated on read/write calls. Unmarshal() replaces all data stored here with whatever
 // was given to it on each call.
 type Manager struct {
-	contract atomic.Value           // Stores a *Contract
-	requests aadInstanceDiscoveryer // *oauth.Token
+	contract   *Contract
+	contractMu sync.RWMutex
+	requests   aadInstanceDiscoveryer // *oauth.Token
 
-	mu sync.Mutex
-
-	cacheMu  sync.Mutex
-	aadCache map[string]authority.InstanceDiscoveryMetadata
+	aadCacheMu sync.RWMutex
+	aadCache   map[string]authority.InstanceDiscoveryMetadata
 }
 
 // New is the constructor for Manager.
 func New(requests *oauth.Client) *Manager {
 	m := &Manager{requests: requests, aadCache: make(map[string]authority.InstanceDiscoveryMetadata)}
-	m.contract.Store(NewContract())
+	m.contract = NewContract()
 	return m
 }
 
@@ -145,9 +143,6 @@ const scopeSeparator = " "
 
 // Write writes a token response to the cache and returns the account information the token is stored with.
 func (m *Manager) Write(authParameters authority.AuthParams, tokenResponse accesstokens.TokenResponse) (shared.Account, error) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
-
 	authParameters.HomeaccountID = tokenResponse.ClientInfo.HomeAccountID()
 	homeAccountID := authParameters.HomeaccountID
 	environment := authParameters.AuthorityInfo.Host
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@
 *.dylib
 *.golangci.yml
 *.swp
+*.pprof
```
abhidnya13, line 9: Not sure what is this changing ?

apps/internal/base/internal/storage/storage.go
```diff
@@ -218,18 +213,13 @@ func (m *Manager) Write(authParameters authority.AuthParams, tokenResponse acces
...
-// Contract returns the Contract for read operations.
-func (m *Manager) Contract() *Contract {
-	return m.contract.Load().(*Contract)
-}
-
 func (m *Manager) getMetadataEntry(ctx context.Context, authorityInfo authority.Info) (authority.InstanceDiscoveryMetadata, error) {
-	m.cacheMu.Lock()
-	defer m.cacheMu.Unlock()
-
+	m.aadCacheMu.RLock()
```
abhidnya13, line 217: Cannot use `defer m.aadCacheMu.RUnlock()` here ?

.gitignore
jhendrixMSFT: This is a profiler data file, no need to commit those.

apps/internal/base/internal/storage/storage.go
jhendrixMSFT: No, m.aadMetadata() takes the write lock and RWMutext doesn't support lock promotion.  I've added a comment explaining this.

abhidnya13: Thanks for the comment ! Makes sense.

element-of-surprise: Sorry to unresolve this:

I'm looking at this and it looks like to me we could do the following to make this honor defers, put the locks in a single place and generally clean it up.  I think Joel this is how it would probably look had you been writing it fresh instead of this re-write.

```go
func (m *Manager) addMetadaFromCache(authorityInfo authority.Info) (authority.InstanceDiscoveryMetadata, error) {
  m.addCacheMu.RLock()
  defer m.addCacheMu.RUnlock()

  metadata, ok := m.aadCache[authorityInfo.Host]
  if ok {
    return metadata, nil
  }
  return metadata, fmt.Errorf("not found")
}

func (m *Manager) aadMetadata(ctx context.Context, authorityInfo authority.Info) (authority.InstanceDiscoveryMetadata, error) {
  if md, err := addMetadaFromCache(authorityInfo); err != nil {
    return md, nil
  }
  ...
}
```

Now in getMetadataEntry(), this should end up being:

```go
  metadata, err := m.aadMetadata(ctx, authorityInfo)
  ...
```
No more locks at that level, everything is deferred

apps/internal/base/internal/storage/storage.go
```diff
@@ -262,9 +256,10 @@ func (m *Manager) aadMetadata(ctx context.Context, authorityInfo authority.Info)
 }
 
 func (m *Manager) readAccessToken(homeID string, envAliases []string, realm, clientID string, scopes []string) (AccessToken, error) {
-	cache := m.Contract()
-
-	for _, at := range cache.AccessTokens {
+	m.contractMu.RLock()
+	defer m.contractMu.RUnlock()
+	// TODO: linear search (over a map no less) is slow for a large number of tokens
```
element-of-surprise, line 261: I'm not sure if the TODO is valid??  It's all going to depend on the average case.  If the average is small < 100 tokens or this isn't in a n^2 scenario, I doubt this will matter.

However, if our case doesn't resemble that, then we are either going to have to:
- write multiple maps keyed on HomeAccountID, Realm and ClientID and do a union of 3 hash lookups
- store another map that has a combo key just for this "HAI:Realm:CID"
- have a different data structure that works for this case

These are all actually expensive.  Your either making multiple maps or are dealing with a data structure that is:
A) Hidden 
or 
B) Serializes into key/values when using JSON.  

With this said, is this valid TODO, is this something we note why we are iterating over a map, .......??????

apps/internal/base/internal/storage/storage.go
```diff
@@ -277,11 +272,10 @@ func (m *Manager) readAccessToken(homeID string, envAliases []string, realm, cli
...
 
 func (m *Manager) writeAccessToken(accessToken AccessToken) error {
+	m.contractMu.Lock()
+	defer m.contractMu.Unlock()
 	key := accessToken.Key()
-
-	cache := m.Contract().copy()
-	cache.AccessTokens[key] = accessToken
-	m.contract.Store(cache)
+	m.contract.AccessTokens[key] = accessToken
```
element-of-surprise, line 278: As a note, not for this line....

Did we run against this with the race detector?

apps/internal/base/internal/storage/storage.go
jhendrixMSFT: After I made this change I was surprised to see that token retrieval wasn't O(1) like I expected, so I profiled it.  It's the dominating node in the profile for a large number of tokens.

I agree that it likely doesn't matter for real-world scenarios, however I wanted to ensure this behavior is documented in case it becomes a problem in the future.

apps/internal/base/internal/storage/storage.go
jhendrixMSFT: Fixed

apps/internal/base/internal/storage/storage.go
jhendrixMSFT: Yes, no problems were detected.

apps/internal/base/internal/storage/storage.go
element-of-surprise: Do me a favor, add the text of the above to give some explanation fo why this isn't done then.  This lets someone in the future understand the kind of things they need to do if this becomes a problem.

jhendrixMSFT: Done

element-of-surprise (approved):

PR #197, opened 2021-03-24
Fixing JSON marshal bug on empty map values

element-of-surprise:
Fixes a bug where the JSON marshaller would write bad JSON when encountering an empty map and the field did not have omitempty in the tag.


jhendrixMSFT:
Does this marshaller behave similar or different from the in-box marshaller WRT empty maps?

abhidnya13 (approved):
Other than the failing builds, functionality wise LGTM. Tested it using a confidential client storing the json in a file and a correct json was output.

element-of-surprise:
> marshaller

It works like encoding/json.  The only problem here was it just wasn't working :)

jhendrixMSFT (approved):

PR #201, opened 2021-03-29
Adding version file

abhidnya13:
Adding this file here to track version number.

### DIFF
--- a/apps/internal/version/version.go
+++ b/apps/internal/version/version.go
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft Corporation.
 // Licensed under the MIT license.
 
+// Package version keeps the version number of the client package.
 package version
 
 // Version is the version of this client package that is communicated to the server.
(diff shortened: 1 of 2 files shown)


jhendrixMSFT:
Is there a reason the wiring up to `comm.go` needs to happen in a separate PR?

abhidnya13:
I tried changing it here but vs code was complaining that there is no version folder to import this value from

element-of-surprise:
@abhidnya13  That is some type of local issue.  Probably will happen at the cmd line too.  

Please figure out what is going on and update the comm package to use this.  Also, go ahead and wire the version up to be 0.1.0 so we don't have to do this later.

abhidnya13:
@jhendrixMSFT , @element-of-surprise Made the changes here.

apps/internal/version/version.go
```diff
@@ -0,0 +1,7 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package version
```
element-of-surprise, line 4: I hate to say this, but you have to put a // Package version .... here

element-of-surprise (approved):

PR #221, opened 2021-06-11
This branch adds readtimeout and writetimeout to the server. It also adds ssytem shutdown signal, allowing us to gracefully shutdown

fadhilijuma:
This PR makes the following changes:
1. Handle the critical Http Server errors that we are swallowing

2. Add os syscall shutdown and interrupt alert handling

3. We are just calling Http Server Shutdown with a context that does not have a timeout. What if the shutdown fails or hangs forever?

4. We should also call shutdown and handle the error incase the shutdown fails. If it fails, then we should call Close( ). This will force the shutdown. If forced shutdown fails then thats a fatal error

### DIFF
--- a/apps/internal/local/server.go
+++ b/apps/internal/local/server.go
@@ -152,7 +152,20 @@ func (s *Server) Result(ctx context.Context) Result {
 // Shutdown shuts down the server.
 func (s *Server) Shutdown() {
 	// Note: You might get clever and think you can do this in handler() as a defer, you can't.
-	_ = s.s.Shutdown(context.Background())
+	const timeout = 5 * time.Second
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	err := s.s.Shutdown(ctx)
+	if err != nil {
+		//If gracefull shutdown fails, then we force shutdown.
+		err = s.s.Close()
+	}
+	if err != nil {
+		//We failed to shutdown completely.
+		//TODO we should pass a logger into this function
+		log.Fatalf("Shutdown: could not stop the server gracefully: %v", err)
+	}
+
 }
 
 func (s *Server) putResult(r Result) {
--- a/apps/internal/local/server.go
+++ b/apps/internal/local/server.go
@@ -162,7 +162,6 @@ func (s *Server) Shutdown() {
 	}
 	if err != nil {
 		//We failed to shutdown completely.
-		//TODO we should pass a logger into this function
 		log.Fatalf("Shutdown: could not stop the server gracefully: %v", err)
 	}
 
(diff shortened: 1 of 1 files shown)


element-of-surprise:
Hey Fadhili,

First, want to thank you for providing a PR.  Before we go any further, I want to understand the purpose of the PR from a practical perspective, since this doesn't come from a linked issue or a feature request.  

When I say this, what is the problem in production/testing that this PR solves?  I get the feeling (but please correct me if I'm wrong) that these may be perceived flaws?

The two main aspects that you appear to want to handle are:

1. A server enforced timeout of HTTP requests to the local server
2. Adding a grace period for a sigterm for requests

The server is local (in this context meaning that we only serve an on localhost) that serves up two static pages.  It is virtually impossible for there to be a timeout unless the entire local machine is so overloaded that nothing is happening.  Unlike a normal service serving customers, I don't see a need for a timeout or a reason we'd want to have one.  However, if during your testing or production run you are seeing this as an issue, we can discuss implementing timeouts.

I'm also curious how the grace period is affecting your test/production use.  If the local application running MSAL is terminating, it is in teardown.  It probably shouldn't have anything inflight.  Unlike a public http service where a graceful handling might be useful (though generally I prefer to drain at a LB and when connection count is 0 terminate), I don't know how this would cause an issue as the caller is the app itself.  So if you could describe the testing/production problem, we can see if this is the best way for us to solve your problem.  

Finally, logging.  We have specifically kept log libraries out of this package.  You will see if used in our samples, but not in the package itself (except for a single location that we need to remove : apps/internal/oauth/ops/wstrust/defs/wstrust_mex_document.go).  We may add support for OpenTelementry in the future and if the track2 SDK comes into GA, we may conform to their logging interface.  Otherwise we only use logging for samples and debugging during dev.

I'm happy discuss any of this with you and get any concrete use cases where you see this happening.  As you are Msft internal, please feel free to reach out to me on Teams at any time.

Cheers and thanks for the PR.

henrik-me:
+ 1 to @element-of-surprise 
@fadhilijuma please create a feature issue that outlines what this is for and what problem it solves.

element-of-surprise:
Follow up:

@henrik-me 

I sat down with Fadhili and we chatted about the changes.  The long story in a short excerpt is he's becoming familiar with the code base to help out in the future.  There is not a concrete problem to solve.  The end result is this PR is not going to move forward, however we might entertain switching from server.Shutdown() to server.Close().  He will be attending our MSAL meeting in a few weeks and where we can discuss this as well as getting him into helping with our bug/feature burndown list.  

He will be contacting @abhidnya13 to put the Shutdown()/Close() on the agenda.

Closing this out and again, thanks Fadhili for reaching out to us.  Great conversation and look forward to working with you in the future.

Issue #172, opened 2021-02-01
[TODO] Sample TODOs


abhidnya13:
- Auth code sample [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/248e6dc01f3d0eedbb227b2745a27806550df9d3/apps/tests/devapps/confidential_auth_code_sample.go#L13) and [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/0b43475868c96ac8fbe5a978fb54840de6bb85c6/apps/tests/devapps/main.go)
- Use flag package in samples main file instead of if...else chain [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/0b43475868c96ac8fbe5a978fb54840de6bb85c6/apps/tests/devapps/main.go)

abhidnya13:
Make the samples self contained. Refer:  https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/228#issuecomment-865307266

PR #228, opened 2021-06-21
Modifying confidential client secret sample to use the latest API

abhidnya13:
Removing the commented sample which was using an older unreleased API and adding one which uses the latest API surface.

### DIFF
--- a/apps/tests/devapps/client_secret_sample.go
+++ b/apps/tests/devapps/client_secret_sample.go
@@ -3,34 +3,31 @@
 
 package main
 
-/*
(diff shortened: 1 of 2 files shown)


element-of-surprise (approved):

tonyeung:
I'd like to make a suggestion. Please make these samples self contained. It took me a good while to realize where the `cacheAccessor` came from.

abhidnya13:
@tonyeung Thanks for the suggestion ! Merging this one for now, there is a TODO to improve samples here #172. I have made a note of this suggestion in the issue so we track this.

PR #234, opened 2021-06-24
Add confidential example test

fadhilijuma:
This PR adds example tests to the confidential package.

### DIFF
--- a/apps/confidential/examples_test.go
+++ b/apps/confidential/examples_test.go
@@ -71,17 +71,17 @@ func ExampleAcquireTokenByAuthCode() {
 	// We could also use an assertion (NewCredFromAssertion) or a certificate (NewCredFromCert) to obtain a credential.
 	cred, err := NewCredFromSecret(secret)
 	if err != nil {
-		log.Fatalf("ExampleAcquireTokenByAuthCode: acquring token by credential %v", err)
+		log.Fatalf("ExampleAcquireTokenByAuthCode: acquring token by auth code %v", err)
 	}
 	client, err := New("fake_client_id", cred)
 	if err != nil {
-		log.Fatalf("ExampleAcquireTokenByAuthCode: acquring token by credential %v", err)
+		log.Fatalf("ExampleAcquireTokenByAuthCode: acquring token by auth code %v", err)
 	}
 
 	ctx := context.Background()
 	token, err := client.AcquireTokenByAuthCode(ctx, "fake_auth_code", "fake_redirect_uri", tokenScope)
 	if err != nil {
-		log.Fatalf("ExampleAcquireTokenByAuthCode: acquring token by credential %v", err)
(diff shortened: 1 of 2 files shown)


apps/confidential/examples_test.go
```diff
@@ -35,3 +36,53 @@ func ExampleNewCredFromCert_pem() {
...
+	token, err := client.AcquireTokenByCredential(ctx, tokenScope)
+	if err != nil {
+		log.Fatalf("ExampleAcquireTokenByCredential: acquring token by credential %v", err)
+	}
+	fmt.Println(token)
+
+}
+
+// ExampleAcquireTokenByAuthCode gives an example of acquiring token by auth code.
+func ExampleAcquireTokenByAuthCode() {
```
abhidnya13, line 66: Same as above about naming. https://blog.golang.org/examples#TOC_4.

apps/confidential/examples_test.go
```diff
@@ -35,3 +36,53 @@ func ExampleNewCredFromCert_pem() {
...
+	}
+	client, err := New("client_id", cred)
+	if err != nil {
+		log.Fatalf("ExampleAcquireTokenByAuthCode: acquring token by auth code %v", err)
+	}
+
+	ctx := context.Background()
+	var tokenScope = []string{"the_scope"}
+
+	token, err := client.AcquireTokenByAuthCode(ctx, "auth_code", "redirect_uri", tokenScope)
```
abhidnya13, line 82: this should be named result as `result`  instead of `token`.
We return a token inside of [AuthResult](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b1a1be475e73ac76ca97d2c399f8aa7b8fe0b986/apps/internal/base/base.go#L68)

abhidnya13 (changes requested):

apps/confidential/examples_test.go
```diff
@@ -35,3 +36,51 @@ func ExampleNewCredFromCert_pem() {
...
+	}
+	client, err := New("client_id", cred)
+	if err != nil {
+		log.Fatalf("ExampleAcquireTokenByAuthCode: acquring token by auth code %v", err)
+	}
+
+	ctx := context.Background()
+	var tokenScope = []string{"the_scope"}
+
+	result, err := client.AcquireTokenByAuthCode(ctx, "auth_code", "redirect_uri", tokenScope)
```
jhendrixMSFT, line 80: While this shows how to invoke the API, I think it needs more documentation.  Where does `auth_code` come from?  What value do I use for `redirect_uri`?  The auth code flow is a bit complicated, so we should have clear docs on how to use it end-to-end.  If that exists elsewhere that's fine, we should then add a code comment with a link to them.

fadhilijuma: Thank you, let me fix that.

abhidnya13:
Closing this due to inactivity.

PR #237, opened 2021-06-26
Leaf Certificate detection

fadhilijuma:
This PR adds leaf certificate detection. We currently are only reading the first item in a pool of certificates and making the assumption that the first item will always be the leaf certificate. The easier and more predictable approach would be to check if the certificate is marked IsCA or not. Root and Intermediary Certificates are marked IsCA as TRUE because these are certificates that generate/issue other certificates. A leaf certificate has IsCA marked as false because its an end entity certificate. 

The PR also renames `base` variable to `baseClient` to avoid collision with the imported package `github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base`

Resources:
RFC 5280, x509 package

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -59,6 +59,9 @@ type AuthResult = base.AuthResult
 
 type Account = shared.Account
 
+// LeafCertificateNotFoundError is returned when we cannot find a leaf certificate from a chain.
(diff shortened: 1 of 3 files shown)


abhidnya13:
Any changes to the public API should be discussed with the team. Lets create an issue and track it.

PR #264, opened 2021-07-16
BugFix - Compare scopes insensitive 

abhidnya13:
Scenario: 

We write scopes in Access Token Cache object like they are sent by AAD. AAD sends scopes like these:
``` "User.Read User.Write"```
If a user requests `AcquireTokenSilent` with `user.read`, MSAL would not find an accessToken because of scopes not matching.

This PR fixes the above condition by making the scope comparison case insensitive.

### DIFF
--- a/apps/internal/base/internal/storage/storage.go
+++ b/apps/internal/base/internal/storage/storage.go
@@ -70,8 +70,10 @@ func checkAlias(alias string, aliases []string) bool {
 
 func isMatchingScopes(scopesOne []string, scopesTwo string) bool {
 	newScopesTwo := strings.Split(scopesTwo, scopeSeparator)
+	toLower(newScopesTwo)
 	scopeCounter := 0
 	for _, scope := range scopesOne {
+		scope = strings.ToLower(scope)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: apps/internal/base/internal/storage/storage.go [hunk 1]
```diff
@@ -70,8 +70,10 @@ func checkAlias(alias string, aliases []string) bool {
 
 func isMatchingScopes(scopesOne []string, scopesTwo string) bool {
 	newScopesTwo := strings.Split(scopesTwo, scopeSeparator)
+	toLower(newScopesTwo)
 	scopeCounter := 0
 	for _, scope := range scopesOne {
+		scope = strings.ToLower(scope)
```
jhendrixMSFT, line 76: Instead of doing this conversion, take a look at [`strings.EqualFold`](https://pkg.go.dev/strings#EqualFold).

abhidnya13: This is exactly what I wanted, making the change, thanks @jhendrixMSFT

jhendrixMSFT (approved):

PR #274, opened 2021-11-10
Adding inmemory partitioned token cache for OBO

abhidnya13:
This PR adds a default inmemory partitioned cache for OnBehalfOf flow only.

The summary of changes as quoted by Bogdan in the linked issue:
- today, we store tokens in memory in maps, e.g. at_key -> at_payload
- this is not efficient when you have a lot of tokens and need to find a single 1
- the problem is partially solved by SuggestedCacheKey, i.e. MSAL helps app developers use a distributed cache
- many app developers do not want to use a distributed cache, for example for low traffic apps.
- proposal is to store tokens like map[partition_key] -> [map[token_key]-> token],

- The internal partition key is
    - AT, RT: user assertion hash
    - ID, Accounts: home account ID.
    - App metadata - not partitioned.

Here are some results from benchmarking that were done:

|              Method | Partitions | TokensPerPartition |          Mean |       Median |       StdDev |          Min |          Max |   
|-------------------- |-----------:|-------------------:|-------------:|------------:|------------:|-------------:|-------------:|
| AcquireTokenOnBehalfOf |          1 |              10000 |       574.00&nbsp;μs   |  540.09&nbsp;μs |403.3185&nbsp;μs |3.089&nbsp;μs |  10.10&nbsp;ms |   
| AcquireTokenOnBehalfOf |          1 |              100000 |   13.756&nbsp;ms     |  13.699&nbsp;ms | 8.034&nbsp;ms| 11.364&nbsp;μs|  52.594&nbsp;ms|   
| AcquireTokenOnBehalfOf |          100|              10000 |     1.351&nbsp;ms     | 1.327&nbsp;ms  |899.3&nbsp;μs | 6.19&nbsp;μs| 52.254&nbsp;ms  |   
| AcquireTokenOnBehalfOf |          1000 |              10000 |  1.375&nbsp;ms        | 1.355&nbsp;ms  | 865.24&nbsp;μs| 7.63&nbsp;μs|   21.324&nbsp;ms|   
| AcquireTokenOnBehalfOf |          10000|              100 |    18.828&nbsp;μs|  18.013&nbsp;μs |35.605&nbsp;μs | 3.756&nbsp;μs|  29.626&nbsp;ms |


apps/internal/base/base.go
```diff
@@ -207,18 +224,33 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
...
-	storageTokenResponse, err := b.manager.Read(ctx, authParams, silent.Account)
-	if err != nil {
-		return AuthResult{}, err
+	authParams.AuthorizationType = silent.AuthorizationType
+	authParams.UserAssertion = silent.UserAssertion
+
+	var storageTokenResponse storage.TokenResponse
+	var err error
+	if authParams.AuthorizationType == authority.ATOnBehalfOf {
+		if s, ok := b.pmanager.(cache.Serializer); ok {
```
jhendrixMSFT, line 234: It looks like this block of code is duplicated in the `else` clause below.  Can it be moved out of the `if` so the duplication is removed?

abhidnya13: This block actually is using the partitioned_manager and the one in the else clause uses a storage manager. So the serializers for both of them are different as the structure of data is different.

jhendrixMSFT: Ah ok I missed that earlier.

apps/internal/base/internal/storage/partitioned_storage.go
```diff
@@ -0,0 +1,426 @@
...
+			tokenResponse.ExtExpiresOn.T,
+			target,
+			tokenResponse.AccessToken,
+		)
+		if authParameters.AuthorizationType == authority.ATOnBehalfOf {
+			accessToken.UserAssertionHash = userAssertionHash // get Hash method on this
+		}
+
+		// Since we have a valid access token, cache it before moving on.
+		if err := accessToken.Validate(); err == nil {
```
jhendrixMSFT, line 124: What happens if `err != nil`?

apps/internal/base/internal/storage/partitioned_storage.go
```diff
@@ -0,0 +1,426 @@
...
+	defer m.aadCacheMu.RUnlock()
+	metadata, ok := m.aadCache[authorityInfo.Host]
+	if ok {
+		return metadata, nil
+	}
+	return metadata, errors.New("not found")
+}
+
+func (m *PartitionedManager) aadMetadata(ctx context.Context, authorityInfo authority.Info) (authority.InstanceDiscoveryMetadata, error) {
+	m.aadCacheMu.Lock()
```
jhendrixMSFT, line 188: Assuming that `m.requests.AADInstanceDiscovery` below makes an IO request.  It's usually preferred to not hold a lock while performing IO.

apps/internal/base/internal/storage/partitioned_storage.go
```diff
@@ -0,0 +1,426 @@
...
+}
+
+func matchFamilyRefreshTokenObo(rt accesstokens.RefreshToken, userAssertionHash string, envAliases []string) bool {
+	return rt.UserAssertionHash == userAssertionHash && checkAlias(rt.Environment, envAliases) && rt.FamilyID != ""
+}
+
+func matchClientIDRefreshTokenObo(rt accesstokens.RefreshToken, userAssertionHash string, envAliases []string, clientID string) bool {
+	return rt.UserAssertionHash == userAssertionHash && checkAlias(rt.Environment, envAliases) && rt.ClientID == clientID
+}
+func (m *PartitionedManager) readRefreshToken(envAliases []string, familyID, clientID, userAssertionHash, partitionKey string) (accesstokens.RefreshToken, error) {
```
jhendrixMSFT, line 247: Nit: Missing new-line before method declaration.

jhendrixMSFT (commented):
Few more comments.

apps/internal/base/internal/storage/partitioned_storage.go
abhidnya13: Good catch, handled it !

apps/internal/base/internal/storage/partitioned_storage.go
abhidnya13: Thats a good point. I believe the lock is for writing in the aadCache on line 198. I have changed the code to acquire the lock after the IO request. Please take a look.

jhendrixMSFT (approved):

PR #275, opened 2021-11-10
fix(README.md): unify usage naming

ksyx:
The case is inconsistent with others surrounding.


abhidnya13 (approved):
Thanks for noticing this !

PR #286, opened 2022-01-16
Add support for returning RefreshToken

bdwyertech:
Not sure why this wasn't returned in the response -- doesn't hurt if its empty.  I am using this lib to create a golang version of `az login`.  The Azure CLI creates `~/.azure/msal_token_cache.json` which may contain a refresh token.

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -84,6 +84,7 @@ type AuthResult struct {
 	Account        shared.Account
 	IDToken        accesstokens.IDToken
 	AccessToken    string
+	RefreshToken   string
 	ExpiresOn      time.Time
 	GrantedScopes  []string
 	DeclinedScopes []string
@@ -97,6 +98,7 @@ func AuthResultFromStorage(storageTokenResponse storage.TokenResponse) (AuthResu
 
 	account := storageTokenResponse.Account
 	accessToken := storageTokenResponse.AccessToken.Secret
+	refreshToken := storageTokenResponse.RefreshToken.Secret
 	grantedScopes := strings.Split(storageTokenResponse.AccessToken.Scopes, scopeSeparator)
 
 	// Checking if there was an ID token in the cache; this will throw an error in the case of confidential client applications.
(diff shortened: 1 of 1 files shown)


juliusl:
I happen to need this refresh token as well. Any reason the library can't return the entire oauth response? 

The previous ADAL go library did, so currently migrating to MSAL is hard without the same results from the client, 

https://github.com/Azure/go-autorest/blob/71d02b67e17a3d80abf8db171f1ca80c39eb2e90/autorest/adal/token.go#L100

abhidnya13:
MSAL does not expose the refresh token to the end user by design. It is stored and used internally under the hood when you need a new access token. You should call acquireTokenSilent each time you need an access token and MSAL will manage the tokens by either returning a cached token to you or using the refresh token to acquire a new access token.

bdwyertech:
> MSAL does not expose the refresh token to the end user by design. It is stored and used internally under the hood when you need a new access token. You should call acquireTokenSilent each time you need an access token and MSAL will manage the tokens by either returning a cached token to you or using the refresh token to acquire a new access token.

This functionality is broken if you don't already have a token with a matching scope.  #287 fixes that.

abhidnya13:
We do provide you the interface to cache tokens in a persistent storage using a cache accessor on your app object. 
We have a sample cache accessor using a file [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/tests/devapps/sample_cache_accessor.go)
And an example of how to pass it to the application object [here]( https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/tests/devapps/username_password_sample.go#L16)

juliusl:
> It is stored and used internally under the hood when you need a new access token.

What if we want a refresh token and not an access token? There are situations where you want to pass a refresh token to an opaque service which will manage token acquisition on its own from that point on. In this case the downstream service doesn't need to understand know anything about MSAL, it just needs to understand OAuth. 

Also, providing the refresh token in the response doesn't limit MSAL from working as originally designed, but **not** providing it causes friction with existing oauth libraries in the ecosystem. 

@abhidnya13 is it possible to rethink this design choice? Currently, the workaround would be to pass a cache interface so that the cache gets serialized, then deserializing the cache immediately to reclaim the refresh token. However, this feels a bit over-kill?

henrik-me:
The token cache should be used as the storage of the RT. You can and should be serializing the cache. MSAL takes care of refreshing the AT as needed.

PR #297, opened 2022-02-28
Add `context` to `ExportReplace` interface

bmorton:
We're looking to use this library to replace the support we built ourselves in our app.  However, in order to instrument the usage of the cache, we need a `context` to be passed via the ExportReplace interface.  This changes that interface and adds the underlying plumbing to make sure the context can be passed through.

This would be a breaking API change, so I'm willing to work around that if it's necessary.  Instead of this change, we could allow adding a `CacheAccessorWithContext` or similar, but it would add some complexity and the breaking API changes aren't onerous.  I'd opt for the breaking change given this is still preview, but I'll defer to y'all there if you decide to accept this change.

Thanks for this library!  Let me know if there's anything I need to change so that this can be considered.

### DIFF
--- a/apps/cache/cache.go
+++ b/apps/cache/cache.go
@@ -11,6 +11,8 @@ implementers on the format being passed.
 */
 package cache
 
+import "context"
+
 // Marshaler marshals data from an internal cache to bytes that can be stored.
 type Marshaler interface {
 	Marshal() ([]byte, error)
@@ -31,9 +33,9 @@ type Serializer interface {
 type ExportReplace interface {
 	// Replace replaces the cache with what is in external storage.
 	// key is the suggested key which can be used for partioning the cache
-	Replace(cache Unmarshaler, key string)
+	Replace(ctx context.Context, cache Unmarshaler, key string)
 	// Export writes the binary representation of the cache (cache.Marshal()) to
 	// external storage. This is considered opaque.
 	// key is the suggested key which can be used for partioning the cache
-	Export(cache Marshaler, key string)
+	Export(ctx context.Context, cache Marshaler, key string)
 }
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -381,12 +381,12 @@ func (cca Client) AcquireTokenOnBehalfOf(ctx context.Context, userAssertion stri
 }
 
 // Account gets the account in the token cache with the specified homeAccountID.
-func (cca Client) Account(homeAccountID string) Account {
-	return cca.base.Account(homeAccountID)
+func (cca Client) Account(ctx context.Context, homeAccountID string) Account {
+	return cca.base.Account(ctx, homeAccountID)
 }
 
 // RemoveAccount signs the account out and forgets account from token cache.
-func (cca Client) RemoveAccount(account Account) error {
-	cca.base.RemoveAccount(account)
+func (cca Client) RemoveAccount(ctx context.Context, account Account) error {
+	cca.base.RemoveAccount(ctx, account)
 	return nil
 }
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -45,8 +45,8 @@ type partitionedManager interface {
 
 type noopCacheAccessor struct{}
 
-func (n noopCacheAccessor) Replace(cache cache.Unmarshaler, key string) {}
-func (n noopCacheAccessor) Export(cache cache.Marshaler, key string)    {}
+func (n noopCacheAccessor) Replace(ctx context.Context, cache cache.Unmarshaler, key string) {}
+func (n noopCacheAccessor) Export(ctx context.Context, cache cache.Marshaler, key string)    {}
 
 // AcquireTokenSilentParameters contains the parameters to acquire a token silently (from cache).
 type AcquireTokenSilentParameters struct {
@@ -233,8 +233,8 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	if authParams.AuthorizationType == authority.ATOnBehalfOf {
 		if s, ok := b.pmanager.(cache.Serializer); ok {
 			suggestedCacheKey := authParams.CacheKey(silent.IsAppCache)
-			b.cacheAccessor.Replace(s, suggestedCacheKey)
-			defer b.cacheAccessor.Export(s, suggestedCacheKey)
+			b.cacheAccessor.Replace(ctx, s, suggestedCacheKey)
+			defer b.cacheAccessor.Export(ctx, s, suggestedCacheKey)
 		}
 		storageTokenResponse, err = b.pmanager.Read(ctx, authParams, silent.Account)
 		if err != nil {
@@ -243,8 +243,8 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	} else {
 		if s, ok := b.manager.(cache.Serializer); ok {
 			suggestedCacheKey := authParams.CacheKey(silent.IsAppCache)
-			b.cacheAccessor.Replace(s, suggestedCacheKey)
-			defer b.cacheAccessor.Export(s, suggestedCacheKey)
+			b.cacheAccessor.Replace(ctx, s, suggestedCacheKey)
+			defer b.cacheAccessor.Export(ctx, s, suggestedCacheKey)
 		}
 		authParams.AuthorizationType = authority.ATRefreshToken
 		storageTokenResponse, err = b.manager.Read(ctx, authParams, silent.Account)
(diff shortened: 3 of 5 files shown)


bmorton:
Hey @abhidnya13 👋!  Is there anything else I need to do to have this PR considered for merge?

bmorton:
I'm using this successfully in my production application now using the branch on my fork.  Now that I've passed the context through, I can see the cache requests in my telemetry, which is essential to running this in production for us.

bmorton:
@rayluo, I saw there was a new release recently and I was wondering if y'all would consider including this change in a future release?

rayluo:
@chlowell ^^^

chlowell:
I like it. We might eventually need this to support other scenarios as well, like distributed caching. However, despite being preview, this project is supported in production, so the change should be additive. I imagine that would look something like this:
```diff
// callers check whether the accessor implements ExportReplace or ExportReplaceWithContext
+++type ExportReplaceWithContext interface {
       Replace(cache Unmarshaler, key string)
       Export(cache Marshaler, key string)
+++    ReplaceContext(ctx context.Context, cache Unmarshaler, key string)
+++    ExportContext(ctx context.Context, cache Marshaler, key string)
}

+++func (cca Client) AccountContext(ctx context.Context, homeAccountID string) Account

// calls AccountContext with context.Background
func(cca Client) Account(homeAccountID string) Account

+++func (cca Client) RemoveAccountContext(ctx context.Context, account Account) error

// calls RemoveAccountContext with context.Background
func (cca Client) RemoveAccount(account Account) error
```
And then we can consider making the break in v1.0. What do you think @bmorton?

apps/confidential/confidential.go
```diff
@@ -381,12 +381,12 @@ func (cca Client) AcquireTokenOnBehalfOf(ctx context.Context, userAssertion stri
 }
 
 // Account gets the account in the token cache with the specified homeAccountID.
-func (cca Client) Account(homeAccountID string) Account {
-	return cca.base.Account(homeAccountID)
+func (cca Client) Account(ctx context.Context, homeAccountID string) Account {
```
bgavrilMS, line 384: @chlowell @element-of-surprise - is there a way in Go to add the context param as suggested without a breaking change? Varargs + interface maybe (http://changelog.ca/log/2015/01/30/golang) ?

apps/cache/cache.go
```diff
@@ -31,9 +33,9 @@ type Serializer interface {
 type ExportReplace interface {
 	// Replace replaces the cache with what is in external storage.
 	// key is the suggested key which can be used for partioning the cache
-	Replace(cache Unmarshaler, key string)
+	Replace(ctx context.Context, cache Unmarshaler, key string)
```
bgavrilMS, line 36: @chlowell @element-of-surprise - is there a way is Go to evolve an interface without breaking change?

In .NET / Java there are 3 ways to do this: 

1. Define a new interface `ExportRelace2` which inherits from `ExportReplace` and defines `Replace(ctx, cache, key)`. Any existing method returning `ExportReplace` now returns `ExportReplace2` - this is not a breaking change. Sometimes app developer has to cast from ExportReplace to ExportReplace2

element-of-surprise: You can't change the interface, that change is breaking.  You can offer a superset:

```go
type ExportReplaceCtx interface {
  ExportReplace

  ReplaceCtx(ctx context.Context, cache Unmarshaler, key string)
  ExportCtx(ctx context.Context, cache Marshaler, key string)
```

```go
if s, ok := b.pmanager.(cache.Serializer); ok {
  suggestedCacheKey := authParams.CacheKey(silent.IsAppCache)
  // If the implementation supports our new ExportReplaceCtx.
  if cache, ok := b.cacheAccessor.(ExportReplaceCtx); ok {
    cache.ReplaceCtx(ctx, s, suggestedCacheKey)
    defer cache.ExportCtx(ctx, s, suggestedCacheKey)
    cache.ReplaceCtx(ctx, s, suggestedCacheKey)
    defer cache.ExportCtx(ctx, s, suggestedCacheKey)
  }else{  // Use the old way.
    b.cacheAccessor.Replace(s, suggestedCacheKey)
    defer b.cacheAccessor.Export(s, suggestedCacheKey)
    b.cacheAccessor.Replace(ctx, s, suggestedCacheKey)
    defer b.cacheAccessor.Export(ctx, s, suggestedCacheKey)
}
```

apps/confidential/confidential.go
chlowell: Yes, however it would be bizarre to take a `Context` as anything but the first argument. Google established that convention in the standard library and chose to add duplicate `*Context` or `*WithContext` methods where needed ([for example](https://pkg.go.dev/net/http#NewRequest)).

element-of-surprise: If we want to support the change, I suggest the superset as I mention in the thread below (I could assign this to myself and make it happen).

Since we don't treat this as a preview, we shouldn't make a breaking change.  For us, a general release means that we support a very specific set of features in MSAL.  

I think for us to consider doing a breaking release, we should have a laundry list of things we want to change or better support.    Also, we should probably be fairly close to having features that let us make a general release.  That way we at least know we won't need/want to make future changes to support the feature set.

element-of-surprise: @bgavrilMS -- You want me to take this and make the changes?

bgavrilMS: Go right ahead @element-of-surprise

apps/cache/cache.go
bgavrilMS: Right, I forgot that method overloading is not available in Go, by design.

element-of-surprise:
@bgavrilMS @chlowell , looking for a yea/nay or any other comments on this.

chlowell:
That looks reasonable to me. Any thoughts @jhendrixMSFT?

@ellismg @vhvb1989, having implemented `ExportReplace`, what do you think of the proposed `ExportReplaceCtx`?

vhvb1989:
LGTM. thank you for the heads up

jhendrixMSFT:
The typical pattern is to add a `WithContext` suffix (e.g. `http.NewRequestWithContext`).  While a bit verbose, I think there's enough prior art on the pattern that callers just get it.

element-of-surprise:
I think you are right that it is a typical pattern in constructors, but I don't think it is in interfaces.

The verbosity of ExportReplaceWithContext or ExportReplaceWithCtx is starting to reach lengths I'm not particularly fond of.  No matter which we choose, Ctx is a pretty standard abbreviation.  And I'm not too worried that people who are implementing this sort of thing are going to get confused at what is going on (but I've certainly been wrong before).  My "feeling" is to forgo "with" and use "Ctx" to keep it as compact as possible.

@chlowell @vhvb1989 @bgavrilMS , thoughts?

bgavrilMS:
Naming is hard and don't want to get into a heated debat. Agreed with current proposal :)

Note that in the devdiv and in some Windows SDK, they just go with numbers - ExportReplace, ExportReplace2. Kind of makes sense when you get to ExportReplace20 :)

chlowell:
I'm fine with `ExportReplaceWithCtx`. What I'd really like to do is replace `ExportReplace`, but that's a future discussion 😄

element-of-surprise:
I understand Charles, there are definitely things worth cleaning up.

An incompatible API change really has a large barrier to entry.  The benefits of consistency and not breaking client code needs to be weighed against the benefits.  Those benefits usually need to be something very drastic.

I think that is what makes Go and the Linux kernel so successful.  Go doesn't break previous programs and Linux doesn't break userspace.  Both could use a huge dose of cleanup and consistency, but it is a high price to pay.

Maybe it will be worth paying it, but I think if we are going to do that, we do it for everything all at once when we are going to achieve GR status.  And then we make a conscious decision not to break the API ever again.

As for the name, it seems the general consensus is to go with the shorter names, and no one really likes tagging Ctx on things, but it is what it is.  And if we introduce breaking changes, ExportReplaceCtx and ExportReplace will be replaced with a single ExportReplace taking the form of ExportReplaceCtx.

Charles, Joel, Bogdan...  I'm thinking we should start tracking things we want to break in the future and put them in list.  Go has this concept of Go2 (which will never happen), but its a list of the breaking changes they'd like to see (context propogation, generics use, ...).  We could create a "GR breaking" list and start compiling.  As we approach that capability, we'd be able to look at this list and weigh it to see if it meets the high bar we set.  We can figure out what that bar is in future discussions.  How does that sound?

John Doak
Manager, Layer 1 Reliability Engineering
***@***.***

bgavrilMS:
Agreed, let's use a label, for example https://github.com/AzureAD/microsoft-authentication-library-for-go/labels/major%20version%20change (Major Version Change) to track these.

chlowell:
This was superseded by #378. Thanks for suggesting it @bmorton

Issue #305, opened 2022-04-14
Unmarshaling AuthResult fails on IDToken


maxsokolovsky:
**Which version of MSAL Go are you using?** v0.4.0.
Note that to get help, you need to run the latest version. 

**Where is the issue?**
* Public client
    * [ ] Device code flow
    * [ ] Username/Password (ROPC grant)
    * [ ] Authorization code flow 
* Confidential client
    * [ ] Authorization code flow 
    * [ ] Client credentials:
        * [ ] client secret
        * [ ] client certificate
* Token cache serialization
     * [ ] In-memory cache
* Other (please describe)
The IDToken field of the AuthResult value has custom logic for getting itself unmarshalled from JSON.
The logic relies on the order of JSON keys, because it splits the raw string and then uses a hardcoded index in the array (1) to find the actual JWT. Thus the order of keys must be specific for the code to find the JWT. 
**Is this a new or an existing app?**

**What version of Go are you using (`go version`)?**
1.17
<pre>
$ go version
</pre>

**What operating system and processor architecture are you using (`go env`)?**
macOS x86
<details><summary><code>go env</code> Output</summary><br><pre>
$ go env

</pre></details>

**Repro**

Marshal a value of type confidential.AuthResult, then try unmarshaling it. Observe an error.

**Expected behavior**
A clear and concise description of what you expected to happen (or code).
The raw JSON representing a value of type confidential.AuthResult should properly unmarshal without errors.
The order of keys should not be important.

**Actual behavior**
A clear and concise description of what happens, e.g. an exception is thrown, UI freezes.
Error:
unable to unmarshal IDToken: invalid character '¢' looking for beginning of value - this is due to the wrong part of the RawResponse string being parsed as a JWT.

**Possible solution**

**Additional context / logs / screenshots**
Add any other context about the problem here, such as logs and screenshots.

PR #311, opened 2022-05-14
Fix: correctly unmarshal additional fields from IDToken

w1ndy:
This fix correctly unmarshal additional fields (e.g., the `roles` claim) in the IDToken into ``IDToken.AdditionalFields``.

### DIFF
--- a/apps/internal/oauth/ops/accesstokens/tokens.go
+++ b/apps/internal/oauth/ops/accesstokens/tokens.go
@@ -76,6 +76,22 @@ func (i *IDToken) UnmarshalJSON(b []byte) error {
 	token.RawToken = jwt
 
 	*i = IDToken(token)
+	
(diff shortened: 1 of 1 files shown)


jennyf19:
#305

w1ndy:
> #305

This pr doesn't fix #305 since that issue is probably related to this line of code:
https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/e1614e74a7270690a127aa40ddb60f70063d59de/apps/internal/oauth/ops/accesstokens/tokens.go#L65
However, I can't give a concrete fix for that issue because I can't reproduce it. Splitting and decoding IDToken worked fine in my case, but additional fields like the ``roles`` claim were missing.

jennyf19:
Got it, thanks for the clarification @w1ndy

bgavrilMS:
We would need some unit tests for this change.

Issue #292, opened 2022-02-08
[Feature Request] Simplify updating client assertions


chlowell:
Because client assertions are baked into `Credential` instances, which are in turn baked into `confidential.Client` instances, it's necessary to create a new instance of each when an assertion expires. Assertions will commonly expire, so it would be great to have a more efficient API for rotating them in MSAL. For example, instead of a string parameter, `NewCredFromAssertion()` could take a callback that returns assertions (something like a `func (context.Context) (string, error)`).

PR #313, opened 2022-05-19
Regenerate client certificate assertions after they expire

### LINKED ISSUE #312: Confidential client doesn't regenerate certificate assertions
There are a couple causes. The operative one today is that `Client.Credential()` only calls `Credential.JWT()` until that method succeeds. Once `Credential` has generated an assertion, `Client` uses that assertion ever after:
https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/e1614e74a7270690a127aa40ddb60f70063d59de/apps/internal/oauth/oauth.go#L104-L109

If it were called again, `Credential.JWT()` would check whether the assertion has expired. However, it simply returns expired assertions:
https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/e1614e74a7270690a127aa40ddb60f70063d59de/apps/internal/oauth/ops/accesstokens/accesstokens.go#L109-L111

chlowell:
This makes `accesstokens.Credential` fully responsible for providing assertions and ensures `Credential.JWT()` generates new assertions as needed from client certificates. This fixes #312 but maintains the current behavior for user-provided assertions--`JWT()` returns these after they expire. I left that as is because `JWT()` can't generate replacements for user-provided assertions and there's no API for applications to do so (#292).


jhendrixMSFT (approved):
LGTM but I'd like an ack from an MSAL expert.

keegan-caruso (approved):

element-of-surprise:
Hey guys, 

Don't wait for me on this one.  If Keegan + Henrik are good, I'm good.  I'm currently getting ready to transit the country and finalizing my book before it releases, so I'm a little out of sorts for extra reviews ATM.

apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -106,14 +109,18 @@ func (c *Credential) JWT(authParams authority.AuthParams) (string, error) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
-	if c.Expires.Before(time.Now()) && c.Assertion != "" {
+	if c.Expires.After(time.Now()) {
```
bgavrilMS, line 112: Can we stop caching assertions completely?  We should rely on token caching not on assertion caching. 

An assertion is not tied just to a certificate, to re-create it you need the `client_id` and the `tenant_id` (for normal AAD assertions). There will be other assertions like ManagedIdentity assertions, which are different.

bgavrilMS: (we've had quite a few bugs because we were caching assertions)

chlowell: > Can we stop caching assertions completely? We should rely on token caching not on assertion caching.
> 

Sure, can it wait for a future PR? I'm making a tactical fix here to get certificate authentication working without breaking anything else. Removing or unexporting the `Expires` or `Assertion` fields entails rewriting unrelated tests and maybe some broad refactoring. I'm uncertain of the side effects but don't think they would be trivial.

bgavrilMS (approved):

PR #316, opened 2022-05-27
Release 0.5.1

rayluo:
The following line will become release note.

* Bugfix: MSAL Go now regenerate client certificate assertions after they expire (#312, #313)

CC: @chlowell , @keegan-caruso


keegan-caruso (approved):

apps/internal/version/version.go
```diff
@@ -5,4 +5,4 @@
 package version
 
 // Version is the version of this client package that is communicated to the server.
-const Version = "0.5.0"
+const Version = "0.5.1"
```
chlowell, line 8: The code change is a patch, but we also raised the minimum required Go version
```suggestion
const Version = "0.6.0"
```

chlowell (approved):

PR #319, opened 2022-06-14
Enable updating client assertions via callback

### LINKED ISSUE #292: [Feature Request] Simplify updating client assertions

chlowell:
This closes #292 by replacing the static string assertion with a callback that enables applications to generate assertions as needed. Following up on https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/313#discussion_r879838372, I also removed assertion caching; `Credential` will now acquire or create a new assertion for each token request.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -136,7 +136,7 @@ type Credential struct {
 	cert *x509.Certificate
 	key  crypto.PrivateKey
 
-	assertion string
+	assertionCallback func(context.Context) (string, error)
 }
 
 // toInternal returns the accesstokens.Credential that is used internally. The current structure of the
@@ -144,7 +144,7 @@ type Credential struct {
 // having import recursion. That requires the type used between is in a shared package. Therefore
 // we have this.
 func (c Credential) toInternal() *accesstokens.Credential {
-	return &accesstokens.Credential{Secret: c.secret, Cert: c.cert, Key: c.key, Assertion: c.assertion}
+	return &accesstokens.Credential{Secret: c.secret, Cert: c.cert, Key: c.key, AssertionCallback: c.assertionCallback}
 }
 
 // NewCredFromSecret creates a Credential from a secret.
@@ -156,11 +156,20 @@ func NewCredFromSecret(secret string) (Credential, error) {
 }
 
 // NewCredFromAssertion creates a Credential from a signed assertion.
+//
+// Deprecated: a Credential created by this function can't refresh the
+// assertion when it expires. Use NewCredFromAssertionCallback instead.
 func NewCredFromAssertion(assertion string) (Credential, error) {
 	if assertion == "" {
 		return Credential{}, errors.New("assertion can't be empty string")
 	}
-	return Credential{assertion: assertion}, nil
+	return NewCredFromAssertionCallback(func(context.Context) (string, error) { return assertion, nil }), nil
+}
+
+// NewCredFromAssertionCallback creates a Credential that invokes a callback to get assertions
+// authenticating the application.
+func NewCredFromAssertionCallback(callback func(context.Context) (string, error)) Credential {
+	return Credential{assertionCallback: callback}
 }
 
 // NewCredFromCert creates a Credential from an x509.Certificate and a PKCS8 DER encoded private key.
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -167,7 +167,7 @@ func NewCredFromAssertion(assertion string) (Credential, error) {
 }
 
 // NewCredFromAssertionCallback creates a Credential that invokes a callback to get assertions
-// authenticating the application.
+// authenticating the application. The callback must be thread safe.
 func NewCredFromAssertionCallback(callback func(context.Context) (string, error)) Credential {
 	return Credential{assertionCallback: callback}
 }
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -20,6 +20,7 @@ import (
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/cache"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/exported"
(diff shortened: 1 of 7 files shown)


apps/confidential/confidential.go
```diff
@@ -156,11 +156,20 @@ func NewCredFromSecret(secret string) (Credential, error) {
...
 func NewCredFromAssertion(assertion string) (Credential, error) {
 	if assertion == "" {
 		return Credential{}, errors.New("assertion can't be empty string")
 	}
-	return Credential{assertion: assertion}, nil
+	return NewCredFromAssertionCallback(func(context.Context) (string, error) { return assertion, nil }), nil
+}
+
+// NewCredFromAssertionCallback creates a Credential that invokes a callback to get assertions
+// authenticating the application.
```
keegan-caruso, line 170: should we document that this needs to be thread safe?

chlowell: Yes, thanks for suggesting it.

apps/confidential/confidential.go
```diff
@@ -156,11 +156,20 @@ func NewCredFromSecret(secret string) (Credential, error) {
...
+func NewCredFromAssertionCallback(callback func(context.Context) (string, error)) Credential {
+	return Credential{assertionCallback: callback}
```
rayluo, line 172: Overall, this PR's introduction of an assert callback looks good to me. My Golang skills (or the lack thereof) does not see a thread safety issue in this line, so I'll leave that part to the experts.

rayluo (approved):
Discussed with Charles offline. This PR's introduction of an assert callback looks good to me.

apps/confidential/confidential.go
```diff
@@ -156,11 +156,20 @@ func NewCredFromSecret(secret string) (Credential, error) {
...
 // NewCredFromAssertion creates a Credential from a signed assertion.
+//
+// Deprecated: a Credential created by this function can't refresh the
+// assertion when it expires. Use NewCredFromAssertionCallback instead.
 func NewCredFromAssertion(assertion string) (Credential, error) {
 	if assertion == "" {
 		return Credential{}, errors.New("assertion can't be empty string")
 	}
-	return Credential{assertion: assertion}, nil
+	return NewCredFromAssertionCallback(func(context.Context) (string, error) { return assertion, nil }), nil
```
bgavrilMS, line 166: What data is available in the `Context`? TokenEndpoint (used as `aud` claim in the jwt) and `ClientId` (used as `iss` claim in the JWT) should be there to help app developer create the JWT.

jhendrixMSFT: The `Context` is used to control the lifetime of IO operation.  If we need to pass additional data to the callback it should be in explicit parameters.

apps/confidential/confidential.go
```diff
@@ -156,11 +156,20 @@ func NewCredFromSecret(secret string) (Credential, error) {
...
+}
+
+// NewCredFromAssertionCallback creates a Credential that invokes a callback to get assertions
+// authenticating the application. The callback must be thread safe.
+func NewCredFromAssertionCallback(callback func(context.Context) (string, error)) Credential {
```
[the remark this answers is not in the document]
chlowell: No problem, that's good feedback, and Go is quite different from other languages in concurrency. Any Go function may run concurrently and invoke other functions concurrently without returning a special type like `Task` or using async/await syntax. So, this callback is as async as the application wants it to be.

apps/internal/exported/exported.go
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+// package exported contains internal types that are re-exported from a public package
+package exported
+
+// AssertionRequestOptions has information required to generate a client assertion
+type AssertionRequestOptions struct {
```
keegan-caruso, line 8: any reason not to just define this as part of the accesstokens package and then do the same re-export from there?

That is closer to the existing patterns followed in this project I believe.

chlowell: We adopted this convention in the Azure SDK to make pubternal API more explicit. This helps SDK developers avoid breaking public APIs with apparently internal changes, and improves documentation by preventing pkgsite from showing users stuff that may not be exported ([like this](https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential#AuthResult)).

It is a new pattern for MSAL though. I don't mind moving the current aliased types into `exported` in another PR or, if you feel strongly about maintaining the current pattern, following it here.

keegan-caruso: makes sense, thanks for explaining.

keegan-caruso (approved):

apps/confidential/confidential.go
```diff
@@ -163,12 +167,12 @@ func NewCredFromAssertion(assertion string) (Credential, error) {
...
 		return Credential{}, errors.New("assertion can't be empty string")
 	}
-	return NewCredFromAssertionCallback(func(context.Context) (string, error) { return assertion, nil }), nil
+	return NewCredFromAssertionCallback(func(context.Context, AssertionRequestOptions) (string, error) { return assertion, nil }), nil
 }
 
 // NewCredFromAssertionCallback creates a Credential that invokes a callback to get assertions
 // authenticating the application. The callback must be thread safe.
-func NewCredFromAssertionCallback(callback func(context.Context) (string, error)) Credential {
+func NewCredFromAssertionCallback(callback func(context.Context, AssertionRequestOptions) (string, error)) Credential {
```
bgavrilMS, line 175: I am not familiar with Go, but please ensure this does not break backwards compat.

bgavrilMS (approved):
LGTM

apps/confidential/confidential.go
chlowell: This comment apparently landed on the diff between two of this PR's commits. There's no compatibility risk here because `NewCredFromAssertionCallback` is entirely new.

Issue #322, opened 2022-06-22
[Bug] wrong scope when getting token silently


marvinkurock:
**Which version of MSAL Go are you using?**
v0.51

**Where is the issue?**
* Public client
    * [ ] Device code flow
    * [ ] Username/Password (ROPC grant)
    * [x] Authorization code flow 

**Is this a new or an existing app?**

**What version of Go are you using (`go version`)?**

<pre>
$ go version
</pre>
go version go1.17.5 darwin/arm64

**What operating system and processor architecture are you using (`go env`)?**

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
╰─ go env
GO111MODULE=""
GOARCH="arm64"
GOBIN=""
GOCACHE="/Users/marvinkurock/Library/Caches/go-build"
GOENV="/Users/marvinkurock/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/marvinkurock/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/marvinkurock/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_arm64"
GOVCS=""
GOVERSION="go1.17.5"
GCCGO="gccgo"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/marvinkurock/development/go/tmw-cli/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -arch arm64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/01/3y5lb3fs17d6g4jkgmt9fm8c0000gn/T/go-build1713909736=/tmp/go-build -gno-record-gcc-switches -fno-common"

</pre></details>
[...]

PR #323, opened 2022-06-23
[Hotfix] incorrect authParams passed to Refresh token method

mkurock:
fixed wrong authParams passed to Refresh function

Regarding this issue:
https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/322


mkurock:
No problem, thanks for the great library.

chlowell:
I incorporated this into #327, and it has now shipped in v0.5.3. Thanks again!

PR #326, opened 2022-07-05
Support certificate chains in SNI authentication

### LINKED ISSUE #267: Can't send certificate chain in SNI authentication
I observe NewCredFromCert takes a single `*x509.Certificate` (although CertFromPEM returns `[]*x509.Certificate`). `Credential` therefore carries a single cert, making it impossible to send a chain via the "x5c" header, as allowed by [RFC 7515](https://datatracker.ietf.org/doc/html/rfc7515#section-4.1.6). Does AAD require only the leaf cert?

(Related: #122)

chlowell:
Closes #267 by adding a `Credential` factory that takes multiple certs. The new test certs happen to have PKCS1 encoded private keys, so I also added support for that format to `CertFromPEM`.


apps/confidential/confidential.go
```diff
@@ -176,10 +185,36 @@ func NewCredFromAssertionCallback(callback func(context.Context, AssertionReques
 	return Credential{assertionCallback: callback}
 }
 
-// NewCredFromCert creates a Credential from an x509.Certificate and a PKCS8 DER encoded private key.
-// CertFromPEM() can be used to get these values from a PEM file storing a PKCS8 private key.
+// NewCredFromCert creates a Credential from an x509.Certificate and an RSA private key.
+// CertFromPEM() can be used to get these values from a PEM file.
 func NewCredFromCert(cert *x509.Certificate, key crypto.PrivateKey) Credential {
-	return Credential{cert: cert, key: key}
+	cred, _ := NewCredFromCertChain([]*x509.Certificate{cert}, key)
```
keegan-caruso, line 191: what is the story on the discarded error here?

chlowell: `NewCredFromCertChain` returns an error when the cert and key are mismatched, or the key isn't an RSA key. These are unlikely errors when the arguments come from `CertFromPEM`, but because handling cert chains entails discovering them, I had `NewCredFromCertChain` report them. `NewCredFromCert` can't return an error, so I maintained its current behavior of allowing problems with its arguments to surface as signing or authentication errors.

apps/confidential/confidential_test.go
```diff
@@ -232,3 +238,120 @@ func TestAcquireTokenByAuthCode(t *testing.T) {
...
+		{"../testdata/test-cert.pem", 1},
+		{"../testdata/test-cert-chain.pem", 2},
+		{"../testdata/test-cert-chain-reverse.pem", 2},
+	} {
+		f, err := os.Open(filepath.Clean(file.path))
+		if err != nil {
+			t.Fatal(err)
+		}
+		defer f.Close()
+		pemData, err := ioutil.ReadAll(f)
```
keegan-caruso, line 256: since we moved up the minimum go version this should be using io.ReadAll

apps/confidential/confidential.go
```diff
@@ -92,19 +91,28 @@ func CertFromPEM(pemData []byte, password string) ([]*x509.Certificate, crypto.P
...
 			}
 			certs = append(certs, cert)
 		case "PRIVATE KEY":
 			if priv != nil {
-				return nil, nil, fmt.Errorf("found multiple blocks labelled 'PRIVATE KEY'")
+				return nil, nil, errors.New("found multiple private key blocks")
 			}
 
 			var err error
 			priv, err = parsePrivateKey(block.Bytes)
```
keegan-caruso, line 103: nit: might be good to rename 'parsePrivateKey' now that we are supporting more than one private key type

chlowell: I inlined it.

apps/confidential/confidential.go
```diff
@@ -92,16 +91,25 @@ func CertFromPEM(pemData []byte, password string) ([]*x509.Certificate, crypto.P
...
 			if priv != nil {
-				return nil, nil, fmt.Errorf("found multiple blocks labelled 'PRIVATE KEY'")
+				return nil, nil, errors.New("found multiple private key blocks")
 			}
 
 			var err error
-			priv, err = parsePrivateKey(block.Bytes)
+			priv, err = x509.ParsePKCS8PrivateKey(block.Bytes)
+			if err != nil {
+				return nil, nil, fmt.Errorf("could not decode private key: %w", err)
```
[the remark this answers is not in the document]
chlowell: No, I was just following along with the existing code. I see these `x509` methods call `errors.New`, so I don't see immediate danger in wrapping. But that implies applications can't have taken a dependency on an error type here and it would therefore be safe to switch to `%v` throughout this function. I'll do that.

jhendrixMSFT (approved):

apps/confidential/confidential.go
```diff
@@ -61,12 +63,9 @@ type AuthResult = base.AuthResult
...
 // CertFromPEM converts a PEM file (.pem or .key) for use with NewCredFromCert(). The file
-// must have the public certificate and the private key encoded. The private key must be encoded
-// in PKCS8 (not PKCS1). This is usually denoted by the section "PRIVATE KEY" (instead of PKCS1's
-// "RSA PRIVATE KEY"). If a PEM block is encoded and password is not an empty string, it attempts
-// to decrypt the PEM blocks using the password. This will return multiple x509 certificates,
-// though this use case should have a single cert. Multiple certs are due to certificate
-// chaining for use cases like TLS that sign from root to leaf.
+// must contain the public certificate and the private key. If a PEM block is encrypted and
+// password is not an empty string, it attempts to decrypt the PEM blocks using the password.
+// Multiple certs are due to certificate chaining for use cases like TLS that sign from root to leaf.
```
rayluo, line 68: :+1:

rayluo (approved):

PR #327, opened 2022-07-08
Fix silent auth scopes and refresh behavior

### LINKED ISSUE #322: [Bug] wrong scope when getting token silently

chlowell:
This fixes two problems with silent authentication:
  1. it ignores the requested scopes (closes #322; thanks @mkurock)
  1. it uses refresh tokens only to refresh expired access tokens i.e., silent authentication for a given scope succeeds only when the cache contains an access token (expired or not) for that scope

Fixing the first problem is straightforward: just pass the right argument. The second problem is more complex. It arises in the cache's `Read` method, which assumes the caller requires an access token matching the given parameters and returns an error when it doesn't find one. If the caller could in principle use a previously cached refresh token to authenticate, it never gets the chance because `Read` doesn't return that refresh token. I fixed this by having `Read` continue past an access token cache miss to fetch more data, such as a refresh token, before returning. This means `Read` now returns a zero-valued access token in some cases. This is okay because `base.Client` is the only caller and already validates access tokens before returning them to the application.

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -262,7 +262,7 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	result, err := AuthResultFromStorage(storageTokenResponse)
 	if err != nil {
 		if reflect.ValueOf(storageTokenResponse.RefreshToken).IsZero() {
-			return AuthResult{}, errors.New("no refresh token found")
+			return AuthResult{}, errors.New("no token found")
 		}
 
 		var cc *accesstokens.Credential
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -270,7 +270,7 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 			cc = silent.Credential
 		}
 
-		token, err := b.Token.Refresh(ctx, silent.RequestType, b.AuthParams, cc, storageTokenResponse.RefreshToken)
+		token, err := b.Token.Refresh(ctx, silent.RequestType, authParams, cc, storageTokenResponse.RefreshToken)
 		if err != nil {
 			return AuthResult{}, err
 		}
--- a/apps/internal/base/internal/storage/items.go
+++ b/apps/internal/base/internal/storage/items.go
@@ -103,8 +103,14 @@ func (a AccessToken) Key() string {
 	)
 }
 
+// FakeValidate enables tests to fake access token validation
+var FakeValidate func(AccessToken) error
+
 // Validate validates that this AccessToken can be used.
 func (a AccessToken) Validate() error {
+	if FakeValidate != nil {
+		return FakeValidate(a)
+	}
 	if a.CachedAt.T.After(time.Now()) {
 		return errors.New("access token isn't valid, it was cached at a future time")
 	}
--- a/apps/internal/base/internal/storage/storage.go
+++ b/apps/internal/base/internal/storage/storage.go
@@ -94,10 +94,7 @@ func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams,
 		return TokenResponse{}, err
 	}
 
-	accessToken, err := m.readAccessToken(homeAccountID, metadata.Aliases, realm, clientID, scopes)
-	if err != nil {
-		return TokenResponse{}, err
-	}
+	accessToken := m.readAccessToken(homeAccountID, metadata.Aliases, realm, clientID, scopes)
 
 	if account.IsZero() {
 		return TokenResponse{
@@ -249,7 +246,7 @@ func (m *Manager) aadMetadata(ctx context.Context, authorityInfo authority.Info)
 	return m.aadCache[authorityInfo.Host], nil
 }
 
-func (m *Manager) readAccessToken(homeID string, envAliases []string, realm, clientID string, scopes []string) (AccessToken, error) {
+func (m *Manager) readAccessToken(homeID string, envAliases []string, realm, clientID string, scopes []string) AccessToken {
 	m.contractMu.RLock()
 	defer m.contractMu.RUnlock()
 	// TODO: linear search (over a map no less) is slow for a large number (thousands) of tokens.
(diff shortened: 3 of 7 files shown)


apps/internal/base/base.go
```diff
@@ -262,15 +262,15 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
-			return AuthResult{}, errors.New("no refresh token found")
+			return AuthResult{}, errors.New("no token found")
 		}
 
 		var cc *accesstokens.Credential
 		if silent.RequestType == accesstokens.ATConfidential {
 			cc = silent.Credential
 		}
 
-		token, err := b.Token.Refresh(ctx, silent.RequestType, b.AuthParams, cc, storageTokenResponse.RefreshToken)
```
chlowell, line 273: This is problem 1. `b.AuthParams` holds common parameters for all auth requests. Its type has a scopes field but on this instance that field always has a zero value (default scopes are handled elsewhere). `authParams` is an instance of the same type which carries overriding values for this auth request, and the scopes requested by the user application.

apps/internal/base/internal/storage/storage.go
```diff
@@ -94,10 +94,7 @@ func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams,
 		return TokenResponse{}, err
 	}
 
-	accessToken, err := m.readAccessToken(homeAccountID, metadata.Aliases, realm, clientID, scopes)
```
chlowell, line 97: This is problem 2. This helper returns an error when it misses. A miss isn't fatal and using an error to signal one requires the caller to inspect the error to determine whether it's safe to proceed. I could've used a sentinel error for misses but that seemed excessive because the helper really doesn't need to return an error--it only searches a map.

bgavrilMS (commented):
LGTM, at least by reading the tests.

keegan-caruso (approved):

chlowell:
Thanks for the additional context. I added a test to verify granted scopes are handled correctly in caching. Clients appear to add "openid profile offline_access" scopes as you describe--at least, I see this in live traffic and the code looks right--but that doesn't have unit test coverage today (#329).

apps/internal/base/internal/storage/items.go
```diff
@@ -103,8 +103,14 @@ func (a AccessToken) Key() string {
 	)
 }
 
+// FakeValidate enables tests to fake access token validation
+var FakeValidate func(AccessToken) error
+
 // Validate validates that this AccessToken can be used.
 func (a AccessToken) Validate() error {
+	if FakeValidate != nil {
```
ivarprudnikov, line 111: It looks like a bad idea. It should be possible to just pass a valid token in the test.

chlowell: To give some context, the cache calls this method before writing a token and won't write that token if this method returns an error. This method will return an error if the token expires within the next 5 minutes. A test that requires the cache to contain an expired token therefore can't pass in a valid token unless it then waits for that token to expire. Perhaps the cache shouldn't check expiry before writing tokens (too big a change to contemplate here) but there are of course other ways to work around this. The test could cache a token that expires in 5 minutes and 1 millisecond then `time.Sleep(2*time.Millisecond)`. I like that approach, but I still prefer `FakeValidate` because it may prove useful in other test cases and doesn't require the one I'm adding in this PR to take a dependency on timing or behavior not under test.

ivarprudnikov (approved):

PR #333, opened 2022-07-28
Provide Accounts() with confidential client

FStelzer:
The public client provides this method and the confidential one should do so too.
Otherwise we can't get the home id for acquiring a silent token when we already have a cache.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -418,6 +418,12 @@ func (cca Client) AcquireTokenOnBehalfOf(ctx context.Context, userAssertion stri
 	return cca.base.AcquireTokenOnBehalfOf(ctx, params)
 }
 
+// Accounts gets all the accounts in the token cache.
+// If there are no accounts in the cache the returned slice is empty.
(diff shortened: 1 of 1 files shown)


apps/confidential/confidential.go
```diff
@@ -418,6 +418,12 @@ func (cca Client) AcquireTokenOnBehalfOf(ctx context.Context, userAssertion stri
 	return cca.base.AcquireTokenOnBehalfOf(ctx, params)
 }
 
+// Accounts gets all the accounts in the token cache.
+// If there are no accounts in the cache the returned slice is empty.
+func (cca Client) Accounts() []Account {
```
bgavrilMS, line 423: No, in confidential client we'd like to avoid `GetAccounts` operation. Instead, use `GetAccount(id)`. 
The reason for this is that `GetAccounts()` does not play well with distributed caching. 

In both web site and web api scenario, the token cache cannot be stored in memory or in a file, since its size is proportional to the number of users.

bgavrilMS:
Please rely on `GetAccount(id)` for confidential client.

PR #335, opened 2022-07-29
added password flow for cca

Andrei-Predoiu:
The password flow is useful especially for Accessing service account emails without using Applciation permissions.

It is missing from golang MSAL Confidential client, so i just copied it from the public one.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -430,7 +430,7 @@ func (cca Client) RemoveAccount(account Account) error {
 }
 
 // AcquireTokenByUsernamePassword acquires a security token from the authority, via Username/Password Authentication.
-// NOTE: this flow is NOT recommended.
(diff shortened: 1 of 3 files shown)


apps/confidential/confidential.go
```diff
@@ -428,3 +428,9 @@ func (cca Client) RemoveAccount(account Account) error {
 	cca.base.RemoveAccount(account)
 	return nil
 }
+
+// AcquireTokenByUsernamePassword acquires a security token from the authority, via Username/Password Authentication.
+// NOTE: this flow is NOT recommended.
```
bgavrilMS, line 433: I would detail this statement as

"The ROPC flow is not recommended, because it will block conditional access and multi-factor authentication, as these protection measures involve user interaction through a browser.

 It can still be used in special cases, such as for testing purposes and in scenarios where the resource does not allow application tokens - client_credentials."

PR #340, opened 2022-08-11
Upgrade linter and address new warnings

chlowell:


apps/internal/local/server.go
```diff
@@ -84,7 +85,7 @@ func New(reqState string, port int) (*Server, error) {
 
 	serv := &Server{
 		Addr:     fmt.Sprintf("http://localhost:%s", portStr),
-		s:        &http.Server{Addr: "localhost:0"},
+		s:        &http.Server{Addr: "localhost:0", ReadHeaderTimeout: time.Second},
```
chlowell, line 88: This is the [Slowloris](https://en.wikipedia.org/wiki/Slowloris_(computer_security)) fix. We could simply disable the check because there's no vulnerability here--this is the only server in MSAL and it runs on localhost. If users want to Slowloris the local server trying to help them authenticate, that's okay. But the header timeout won't hurt anything and it feels cleaner to me than disabling a warning :shrug:

apps/internal/base/internal/storage/items_test.go
```diff
@@ -204,13 +203,7 @@ func TestAppMetaDataMarshal(t *testing.T) {
 }
 
 func TestContractUnmarshalJSON(t *testing.T) {
-	jsonFile, err := os.Open(testFile)
-	if err != nil {
-		panic(err)
-	}
-	defer jsonFile.Close()
```
jhendrixMSFT, line 211: Was this inadvertently removed?

jhendrixMSFT (approved):

apps/internal/base/internal/storage/items_test.go
chlowell: No, I changed `os.Open -> ioutil.ReadAll` to simply `os.ReadFile`.

jhendrixMSFT: Oh yes, I read that wrong.

keegan-caruso (approved):

PR #343, opened 2022-08-17
Add per-request tenant ID option

### LINKED ISSUE #296: [Feature Request] Enable changing tenant after client construction
Acquiring tokens as a multi-tenant application is painful because tenants are baked into clients at construction. It's therefore necessary to create a new client instance for each tenant. Please provide an API for overriding the tenant specified at construction during a single token request or enable changing it after construction. MSAL.NET has implemented the former and is evidently moving to the latter:  https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/368#issuecomment-945631545.

chlowell:
This closes #296 by adding a new option, `WithTenantID`, for public and confidential client token acquisition methods. Tenant resolution follows the logic laid out in [#296](https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/296#issuecomment-1122156757); see the new tests. Feel free to just look at the tests if you aren't interested in the Go plumbing that makes this new option work.

`WithTenantID` is straightforward to use in an application but its implementation is complex because it must function alongside options which apply only to particular methods, those methods must accept it without a breaking change to their signatures, and we still want the compiler to prevent users from passing invalid options to a method. Credit goes to @element-of-surprise for the pattern I implemented here.

I exported the new `*Option` interfaces supporting `WithTenantID` to enable users to fake `Client` with their own interfaces (but users can't implement the `*Option` interfaces). Sadly, this clutters the public API with overlapping names, for example:
```go
type AcquireTokenByAuthCodeOption func(a *AcquireTokenByAuthCodeOptions)
type AcquireByAuthCodeOption interface { ... }
type AcquireTokenByAuthCodeOptions struct { ... }
```


apps/confidential/confidential.go
```diff
@@ -363,10 +407,23 @@ func WithSilentAccount(account Account) AcquireTokenSilentOption {
...
-func (cca Client) AcquireTokenSilent(ctx context.Context, scopes []string, options ...AcquireTokenSilentOption) (AuthResult, error) {
+//
+// Options:
+//   - [WithSilentAccount]
+//   - [WithTenantID]
+func (cca Client) AcquireTokenSilent(ctx context.Context, scopes []string, options ...acquireTokenSilentOption) (AuthResult, error) {
 	opts := AcquireTokenSilentOptions{}
 	for _, o := range options {
-		o(&opts)
+		switch t := o.(type) {
```
chlowell, line 417: This is ugly. We could replace it with a call to `ApplyOptions` by having `AcquireTokenSilentOption` implement `CallOption`. That would be okay functionally because the compiler would still prohibit passing an `AcquireTokenSilentOption` to other methods (because `AcquireTokenSilentOption` doesn't implement e.g.  `acquireTokenByAuthCodeOption`). However, implementing `CallOption` on `AcquireTokenSilentOption` requires adding a public method to it or moving the definition of `CallOption` into this module i.e., implementing `CallOption` separately in `public` and `confidential`. I'm not sure eliminating this switch is worth it. What do you think?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/AzureAD/microsoft-authentication-library-for-go
 
-go 1.17
+go 1.18
```
jhendrixMSFT, line 3: Are we using something new in Go 1.18?

element-of-surprise:
There are two things here:

1.17 is out of support. Go team supports two versions, which means 1.18 and 1.19

1.18 introduces generics.  I haven't looked but if he ported my libs into here it uses generics in the apply loop.

Charles and I have been talking, and we think we should probably switch these shortly after Go team drops support for the previous version.

What do you think Joel?

Charles, I haven't looked at this yet, but it will probably be Monday because I took a couple days off this week.  You may need to ping me, as I tend to forget stuff when I go off for 4 days.

-- Doak

go.mod
jhendrixMSFT: I missed that generics were introduced in this change, so that part makes sense.

Regarding moving forward to the supported N-1 release when no version-specific features are used, I'm not quite convinced.  While I don't think we should go out of our way to support unsupported versions of Go, I don't think we should just drop them.

apps/internal/oauth/ops/authority/authority_test.go
```diff
@@ -317,3 +318,44 @@ func TestCreateAuthorityInfoFromAuthorityUri(t *testing.T) {
...
+		authority, expectedAuthority, tenant string
+		expectError                          bool
+	}{
+		{authority: host + "common", tenant: uuid1, expectedAuthority: host + uuid1},
+		{authority: host + "consumers", tenant: uuid1, expectedAuthority: host + uuid1},
+		{authority: host + "organizations", tenant: uuid1, expectedAuthority: host + uuid1},
+		{authority: host + uuid1, tenant: uuid2, expectedAuthority: host + uuid2},
+		{authority: host + uuid1, tenant: "common", expectError: true},
+		{authority: host + uuid1, tenant: "organizations", expectError: true},
+		{authority: host + uuid1, tenant: uuid2, expectedAuthority: host + uuid2},
```
rayluo, line 336: Did we mean to have this line here? Seems duplicated with line 333.

chlowell: Nope, good eye! I'll delete it.

apps/confidential/confidential.go
```diff
@@ -400,16 +400,60 @@ func (cca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string,
...
+	shared.CallOption
+} {
+	return struct {
+		AcquireByAuthCodeOption
+		AcquireByCredentialOption
+		AcquireOnBehalfOfOption
+		AcquireSilentOption
+		shared.CallOption
+	}{
+		CallOption: shared.NewCallOption(
```
bgavrilMS, line 418: I would add another restriction here, i.e. we should not allow the `tenantID` to be `common` or `organizations` because the intent of this option is to help multi-tenant apps get tokens for specific tenants only. It's ok to go from general to specific, i.e. authority = "l.m.o/common" and tenant = "12345-..", but there aren't any scenarios where you'd want to go from "common" to "common" or from "12345" to "common".

bgavrilMS: Also, both in client_credentials flow and in OBO flow, calls should not be made over "common" / "organizations". 

- in client_credentials we should outright ban it, because "common" is only supposed to work with user flows
- in OBO, if you've got a multi-tenant app, the tenant should be `incomming_assertion.tid`

chlowell: "common" is the default confidential client tenant here, so I opened #348 to track addressing your first point. For OBO, you mean that the application should request a token from the same tenant the user authenticated in i.e., the value of the assertion's `tid` claim?

apps/public/public.go
```diff
@@ -235,10 +342,23 @@ func WithChallenge(challenge string) AcquireTokenByAuthCodeOption {
...
+//
+// Options:
+//   - [WithChallenge]
+//   - [WithTenantID]
+func (pca Client) AcquireTokenByAuthCode(ctx context.Context, code string, redirectURI string, scopes []string, options ...AcquireByAuthCodeOption) (AuthResult, error) {
 	opts := AcquireTokenByAuthCodeOptions{}
 	for _, o := range options {
-		o(&opts)
+		switch t := o.(type) {
+		case AcquireTokenByAuthCodeOption:
```
element-of-surprise, line 353: I looked at this a while and I've come down on the side that we should just make all options here use the new methodology.  Then you can just use the ApplyOptions().  

This makes all options have the same look and feel.  I mean, once you run into one shared option, you gotta deal with it.  Better to make them all consistent.  That also frees us from these switch statements.  

A note as well, you might add in some private doc a link to the explanation on how this all works:
https://blog.devgenius.io/go-call-option-that-can-be-used-with-multiple-methods-6c81734f3dbe

apps/confidential/confidential.go
```diff
@@ -400,16 +400,60 @@ func (cca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string,
...
 // These are set by using various AcquireTokenSilentOption functions.
 type AcquireTokenSilentOptions struct {
 	// Account represents the account to use. To set, use the WithSilentAccount() option.
 	Account Account
+
+	tenantID string
+}
+
+// AcquireSilentOption is implemented by options for AcquireTokenSilent
+type AcquireSilentOption interface {
```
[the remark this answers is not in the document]
chlowell: Yes, to enable customers to fake the method with a parameter of this type.

apps/internal/shared/options.go
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package shared
```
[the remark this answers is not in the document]
chlowell: Me too. Moved this into `internal/options`.

apps/internal/options/options.go
```diff
@@ -0,0 +1,52 @@
...
+// https://blog.devgenius.io/go-call-option-that-can-be-used-with-multiple-methods-6c81734f3dbe
+// for an explanation of the usage pattern.
+type CallOption interface {
+	Do(any) error
+	callOption()
+}
+
+// ApplyOptions applies all the callOptions to options. options must be a pointer to a struct and
+// callOptions must be a list of objects that implement CallOption.
+func ApplyOptions[O, C any](options O, callOptions []C) error {
```
jhendrixMSFT, line 21: Given the comment, you could use a constraint on `C`.  Or `C` isn't a generic parameter and is just `[]CallOption`?

apps/public/public.go
```diff
@@ -221,32 +327,62 @@ func (pca Client) AcquireTokenByDeviceCode(ctx context.Context, scopes []string)
 // AcquireTokenByAuthCodeOptions contains the optional parameters used to acquire an access token using the authorization code flow.
 type AcquireTokenByAuthCodeOptions struct {
 	Challenge string
+
+	tenantID string
+}
+
+// AcquireByAuthCodeOption is implemented by options for AcquireTokenByAuthCode
+type AcquireByAuthCodeOption interface {
```
jhendrixMSFT, line 335: Does this, and the other interfaces, need to be exported?  Seems like clutter.

jhendrixMSFT: nvm, I see this is used publicly.

apps/internal/options/options.go
chlowell: `C` is unconstrained to allow options of all types. For example, `AcquireTokenSilent` takes `...AcquireTokenSilentOption`. In practice those options will be anonymous structs implementing `CallOption`. However, `AcquireTokenSilentOption` itself doesn't implement `CallOption`. If `C` were simply a `[]CallOption`, passing `[]AcquireTokenSilentOption` would get a spurious compile error. You could call that error useful and the `any` a kludge. I can live with this in an internal API but am open to better ideas. The best alternative that comes to my mind would be to implement `CallOption` on the various option types, which in this design would mean adding a new public method to each of them or moving `CallOption` into `public` and `confidential` to avoid exporting that method.

apps/confidential/confidential.go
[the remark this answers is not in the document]
element-of-surprise: True enough :)

apps/confidential/confidential.go
```diff
@@ -400,40 +401,101 @@ func (cca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string,
 	return cca.base.AuthCodeURL(ctx, clientID, redirectURI, scopes, cca.base.AuthParams)
 }
 
+// WithTenantID specifies a tenant for a single authentication. It may be different than the tenant set in [New] by [WithAuthority].
```
element-of-surprise, line 404: You might want to use the 1.19 doc comments to list which calls this works with.

apps/confidential/confidential.go
```diff
@@ -400,40 +401,101 @@ func (cca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string,
...
 // AcquireTokenSilentOption changes options inside AcquireTokenSilentOptions used in .AcquireTokenSilent().
 type AcquireTokenSilentOption func(a *AcquireTokenSilentOptions)
 
+func (AcquireTokenSilentOption) acquireSilentOption() {}
+
 // WithSilentAccount uses the passed account during an AcquireTokenSilent() call.
-func WithSilentAccount(account Account) AcquireTokenSilentOption {
-	return func(a *AcquireTokenSilentOptions) {
-		a.Account = account
+func WithSilentAccount(account Account) interface {
```
element-of-surprise, line 459: Same doc thing.

element-of-surprise (approved):
Reviewed from the options part of the code, which lgtm.

apps/internal/options/options.go
jhendrixMSFT: Does this mean there are cases where the type-assertion on line 23 fails and we expect this?

chlowell: No, it means the assertion never fails but I don't see a good way to obviate it.

jhendrixMSFT: I get how this works now, and I agree that compromising on the internal API is a good choice (compared to the other options you listed).

jhendrixMSFT (approved):
LGTM although somebody from the MSAL should review the MSAL-specific parts.

bgavrilMS (approved):

rayluo (approved):
Thanks for the discussion on the feature. This implementation looks good to me.

Issue #301, opened 2022-03-25
Consistency track: please provide consistent APIs to disable authority validation and instance discovery


xiangyan99:
MSAL Go has no API for this now but when it does, this is only to track the work and to be sure it's consistent with MSAL best practice

jennyf19:
@xiangyan99 Can you provide some examples of which APIs you are referring to? Thanks.

xiangyan99:
FYI. In msal Python, when creating ClientApplication, we can opt out validate_authority to false. https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/application.py#L164.

User can use it when they want to use a private authority.

.net, java, js have similar functionalities but we have not yet added this support for go.

henrik-me:
potentially dupe of: https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/289

chlowell:
Closed by #362, which added a `WithInstanceDiscovery()` option

PR #344, opened 2022-08-18
Enable using confidential.Client only for token caching

### LINKED ISSUE #302: [Feature Request] Enable Azure.Identity to benefit from MSAL.NET token cache when used for Managed identity
Copied from https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3137:

Is your feature request related to a problem? Please describe.
Azure.Identity (Azure SDK) is using MSAL.NET for some flows, and in that case it benefits from MSAL.NET cache. But in other cases, it goes directly to the MSI endpoint, and does not cache the token.
The ask is to have a way for the Azure SDK to benefit from MSAL.NET cache

Describe the solution you'd like
The spec is [here](https://microsoft.sharepoint.com/:w:/t/activedirectory/acmi/EUyj9w-qLt5IqsqtXydnFdAB0nIbjbBQKbjTLl-RuWSpCw?e=QftciK)

Describe alternatives you've considered
See the spec

Additional context
This will be a consistency item,

chlowell:
Closes #302 by adding `confidential.NewCredFromTokenProvider()`, which enables creating a confidential client that invokes a callback to acquire access tokens. Applications can use this to handle all the details of authentication themselves while relying on the MSAL client for token caching. I mostly followed the design doc but want to call out one change: `TokenProviderResult` here doesn't have a `RefreshInSeconds` field. I omitted this because MSAL for Go currently uses a hardcoded refresh window and I don't want to imply to users that setting `RefreshInSeconds` would have an effect. Adding this field later won't break the API.

To prevent unnecessary authority validation and metadata discovery, I added an internal known authority API. This isn't part of the public API but is a start on #301.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -366,7 +366,23 @@ func New(clientID string, cred Credential, options ...Option) (Client, error) {
 		return Client{}, err
 	}
 
-	base, err := base.New(clientID, opts.Authority, oauth.New(opts.HTTPClient), base.WithX5C(opts.SendX5C), base.WithCacheAccessor(opts.Accessor), base.WithRegionDetection(opts.AzureRegion))
+	baseOpts := []base.Option{
+		base.WithCacheAccessor(opts.Accessor),
+		base.WithRegionDetection(opts.AzureRegion),
+		base.WithX5C(opts.SendX5C),
+	}
+	if cred.tokenProvider != nil {
+		// The caller will handle all details of authentication, using Client only as a token cache.
+		// Declaring the authority host known prevents unnecessary metadata discovery requests. (The
+		// authority is irrelevant to Client and friends because the token provider is responsible
+		// for authentication.)
+		parsed, err := url.Parse(opts.Authority)
+		if err != nil {
+			return Client{}, errors.New("invalid authority")
+		}
+		baseOpts = append(baseOpts, base.WithKnownAuthorityHosts([]string{parsed.Hostname()}))
+	}
+	base, err := base.New(clientID, opts.Authority, oauth.New(opts.HTTPClient), baseOpts...)
 	if err != nil {
 		return Client{}, err
 	}
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -140,6 +140,8 @@ type Credential struct {
 	x5c  []string
 
 	assertionCallback func(context.Context, AssertionRequestOptions) (string, error)
+
+	tokenProvider func(context.Context, TokenProviderParameters) (TokenProviderResult, error)
 }
 
 // toInternal returns the accesstokens.Credential that is used internally. The current structure of the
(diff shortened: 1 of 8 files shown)


apps/confidential/confidential.go
```diff
@@ -207,6 +216,19 @@ func NewCredFromCertChain(certs []*x509.Certificate, key crypto.PrivateKey) (Cre
 	return cred, nil
 }
 
+// TokenProviderParameters is the authentication parameters passed to token providers
+type TokenProviderParameters = exported.TokenProviderParameters
+
+// TokenProviderResult is the authentication result returned by custom token providers
+type TokenProviderResult = exported.TokenProviderResult
+
+// NewCredFromTokenProvider creates a Credential from a function that provides access tokens. This
```
bgavrilMS, line 225: For the public comments, please add a note that this extensibility point is only meant for Azure SDK to cache MSI tokens. We want to be explicit about this, so that other users do not try to use this. Long term we want to move all MSI logic into MSALs, so we could even drop this extensiblity point.

apps/confidential/confidential.go
```diff
@@ -324,7 +346,23 @@ func New(clientID string, cred Credential, options ...Option) (Client, error) {
...
 	}
 
-	base, err := base.New(clientID, opts.Authority, oauth.New(opts.HTTPClient), base.WithX5C(opts.SendX5C), base.WithCacheAccessor(opts.Accessor), base.WithRegionDetection(opts.AzureRegion))
+	baseOpts := []base.Option{
+		base.WithCacheAccessor(opts.Accessor),
+		base.WithRegionDetection(opts.AzureRegion),
+		base.WithX5C(opts.SendX5C),
+	}
+	if cred.tokenProvider != nil {
+		// The caller will handle all details of authentication, using Client only as a token cache.
```
bgavrilMS, line 355: That's a very nice detail

apps/internal/base/base.go
```diff
@@ -147,6 +147,15 @@ func WithCacheAccessor(ca cache.ExportReplace) Option {
 	}
 }
 
+// WithKnownAuthorityHosts specifies hosts Client shouldn't validate or request metadata for because they're known to the user
+func WithKnownAuthorityHosts(hosts []string) Option {
+	return func(c *Client) {
+		cp := make([]string, len(hosts))
+		copy(cp, hosts)
```
bgavrilMS, line 154: go newbie question - why do you need to copy the hosts slice to cp? Is it because MSAL might append data to it?

apps/internal/exported/exported.go
```diff
@@ -12,3 +12,23 @@ type AssertionRequestOptions struct {
 	// TokenEndpoint is the intended token endpoint. Used as the assertion's "aud" claim.
 	TokenEndpoint string
 }
+
+// TokenProviderParameters is the authentication parameters passed to token providers
+type TokenProviderParameters struct {
```
bgavrilMS, line 17: Consider naming this `AppTokenProvider...` since it only works for client_credential flow and not for user flows (auth_code, obo etc.)

bgavrilMS (approved):

keegan-caruso (approved):

apps/internal/base/base.go
chlowell: That's the gist. A slice value is a pointer. Appending copies a slice without modifying the original but elements are mutable otherwise. This explicit copying will prevent sharing memory with applications once this API is publicly accessible and `hosts` is provided by users.

rayluo (approved):

PR #354, opened 2022-10-11
Add WithLoginHint option for interactive authentication

### LINKED ISSUE #271: Support login_hint in auth code flow
I see some vestigial code for this:
https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/cc87fdbb7684389a230546db7fa22925b87557db/apps/internal/base/base.go#L190-L198

Please add at least `login_hint`; the Azure SDK would use it.

chlowell:
Closes #271 by adding an option to set `login_hint` for interactive auth and auth code URLs. Usage looks like this:
```go
token, err := client.AcquireTokenInteractive(context.Background(), scopes, public.WithLoginHint("upn")
```

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -219,6 +219,9 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
 	if authParams.CodeChallengeMethod != "" {
 		v.Add("code_challenge_method", authParams.CodeChallengeMethod)
 	}
+	if authParams.LoginHint != "" {
+		v.Add("login_hint", authParams.LoginHint)
+	}
 	if authParams.Prompt != "" {
 		v.Add("prompt", authParams.Prompt)
 	}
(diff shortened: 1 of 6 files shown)


bgavrilMS (approved):
LGTM

apps/internal/base/base.go
```diff
@@ -219,6 +219,9 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
 	if authParams.CodeChallengeMethod != "" {
 		v.Add("code_challenge_method", authParams.CodeChallengeMethod)
 	}
+	if authParams.LoginHint != "" {
```
bgavrilMS, line 222: A simple adition to this feature would be to expose the login_hint to the `GetAuthorizationUrl` API on confidential client

chlowell: Thanks for pointing this out! I've added the option to the public and confidential URL builders.

keegan-caruso (approved):

PR #355, opened 2022-10-13
Support client capabilities and claims

### LINKED ISSUE #263: [Feature Request]Claims request parameter support in MSAL
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/789600

chlowell:
Closes #263 by adding a `WithClientCapabilities` option for client constructors and a `WithClaims` option for token acquisition methods. Usage looks like this:
```go
client, err := New("id", cred, WithClientCapabilities([]string{"CP1"}))
token, err := client.AcquireTokenByCredential(ctx, scope, WithClaims("some JSON"))
```
(confidential and public clients have the same options)

I believe I've covered all the bases here (see the e2e tests) but please look closely; many code paths are impacted. Let me highlight a few subtleties:
- capabilities must interpolate into a valid JSON array but otherwise we have no opinion on their validity
- claims must be valid JSON. Callers are responsible for decoding challenges. This is documented and the error returned for invalid JSON calls this out.
- `AcquireTokenSilent(..., WithClaims("..."))` won't return a cached access token. This means it always fails for confidential clients. It can succeed for public clients that have a refresh token for the account. Similarly, `AcquireTokenOnBehalfOf` can succeed silently for confidential clients.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -284,6 +284,8 @@ type Options struct {
 
 	// Instructs MSAL Go to use an Azure regional token service with sepcified AzureRegion.
 	AzureRegion string
+
+	capabilities []string
 }
 
 func (o Options) validate() error {
@@ -312,6 +314,15 @@ func WithAuthority(authority string) Option {
 func WithAccessor(accessor cache.ExportReplace) Option {
 	return func(o *Options) {
 		o.Accessor = accessor
+	}
+}
+
+// WithClientCapabilities allows configuring one or more client capabilities such as "CP1"
+func WithClientCapabilities(capabilities []string) Option {
+	return func(o *Options) {
+		// there's no danger of sharing the slice's underlying memory with the application because
+		// this slice is simply passed to base.WithClientCapabilities, which copies its data
+		o.capabilities = capabilities
 	}
 }
 
@@ -369,6 +380,7 @@ func New(clientID string, cred Credential, options ...Option) (Client, error) {
 
 	baseOpts := []base.Option{
 		base.WithCacheAccessor(opts.Accessor),
+		base.WithClientCapabilities(opts.capabilities),
 		base.WithRegionDetection(opts.AzureRegion),
 		base.WithX5C(opts.SendX5C),
 	}
@@ -398,7 +410,7 @@ func (cca Client) UserID() string {
 
 // authCodeURLOptions contains options for AuthCodeURL
 type authCodeURLOptions struct {
-	loginHint, tenantID string
+	claims, loginHint, tenantID string
 }
 
 // AuthCodeURLOption is implemented by options for AuthCodeURL
@@ -409,6 +421,7 @@ type AuthCodeURLOption interface {
 // AuthCodeURL creates a URL used to acquire an authorization code. Users need to call CreateAuthorizationCodeURLParameters and pass it in.
 //
 // Options:
+// - [WithClaims]
 // - [WithLoginHint]
 // - [WithTenantID]
 func (cca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, scopes []string, opts ...AuthCodeURLOption) (string, error) {
@@ -420,6 +433,7 @@ func (cca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string,
 	if err != nil {
 		return "", err
 	}
+	ap.Claims = o.claims
 	ap.LoginHint = o.loginHint
 	return cca.base.AuthCodeURL(ctx, clientID, redirectURI, scopes, ap)
 }
@@ -438,6 +452,47 @@ func WithLoginHint(username string) interface {
 				switch t := a.(type) {
 				case *authCodeURLOptions:
 					t.loginHint = username
+				default:
+					return fmt.Errorf("unexpected options type %T", a)
+				}
+				return nil
+			},
+		),
+	}
+}
+
+// WithClaims sets additional claims to request for the token, such as those required by conditional access policies.
+// Use this option when Azure AD returned a claims challenge for a prior request. The argument must be decoded.
+// This option is valid for any token acquisition method.
+func WithClaims(claims string) interface {
+	AcquireByAuthCodeOption
+	AcquireByCredentialOption
+	AcquireOnBehalfOfOption
+	AcquireSilentOption
+	AuthCodeURLOption
+	options.CallOption
+} {
+	return struct {
+		AcquireByAuthCodeOption
+		AcquireByCredentialOption
+		AcquireOnBehalfOfOption
+		AcquireSilentOption
+		AuthCodeURLOption
+		options.CallOption
+	}{
+		CallOption: options.NewCallOption(
+			func(a any) error {
+				switch t := a.(type) {
+				case *AcquireTokenByAuthCodeOptions:
+					t.claims = claims
+				case *acquireTokenByCredentialOptions:
+					t.claims = claims
+				case *acquireTokenOnBehalfOfOptions:
+					t.claims = claims
+				case *AcquireTokenSilentOptions:
+					t.claims = claims
+				case *authCodeURLOptions:
+					t.claims = claims
 				default:
 					return fmt.Errorf("unexpected options type %T", a)
 				}
@@ -493,7 +548,7 @@ type AcquireTokenSilentOptions struct {
 	// Account represents the account to use. To set, use the WithSilentAccount() option.
 	Account Account
 
-	tenantID string
+	claims, tenantID string
 }
 
 // AcquireSilentOption is implemented by options for AcquireTokenSilent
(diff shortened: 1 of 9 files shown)


apps/public/public_test.go
```diff
@@ -234,6 +237,169 @@ func TestAcquireTokenWithTenantID(t *testing.T) {
...
+				}
+				mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, tenant)))
+				mockClient.AppendResponse(mock.WithBody(mock.GetTenantDiscoveryBody(lmo, tenant)))
+				switch method {
+				case "devicecode":
+					mockClient.AppendResponse(mock.WithBody([]byte(`{"device_code":".","expires_in":600}`)))
+				case "password":
+					mockClient.AppendResponse(mock.WithBody([]byte(`{"account_type":"Managed","cloud_audience_urn":".","cloud_instance_name":".","domain_name":"."}`)))
+				case "passwordFederated":
+					mockClient.AppendResponse(mock.WithBody([]byte(`{"account_type":"Federated","cloud_audience_urn":".","cloud_instance_name":".","domain_name":".","federation_protocol":".","federation_metadata_url":"."}`)))
```
rayluo, line 311: Outside of the scope of this PR, but a FYI: [MSALs are allowing confidential clients to use password flow (internal work item)](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2092736).

rayluo (approved):
Overall LGTM. Leaving a couple minor comments below.

apps/confidential/confidential_test.go
```diff
@@ -529,6 +531,185 @@ func TestNewCredFromTokenProviderError(t *testing.T) {
...
+		{
+			claims:   `{"id_token":{"auth_time":{"essential":true}}}`,
+			expected: `{"id_token":{"auth_time":{"essential":true}}}`,
+		},
+		{
+			capabilities: []string{"cp1", "cp2"},
+			claims:       `{"access_token":{"nbf":{"essential":true, "value":"42"}}}`,
+			expected:     `{"access_token":{"nbf":{"essential":true, "value":"42"}, "xms_cc":{"values":["cp1","cp2"]}}}`,
+		},
+	} {
```
bgavrilMS, line 584: That is a great unit test, very easy to read and understand.

bgavrilMS (approved):

bgavrilMS (approved):

apps/confidential/confidential_test.go
```diff
@@ -529,6 +531,185 @@ func TestNewCredFromTokenProviderError(t *testing.T) {
...
+				}
+				if err != nil {
+					t.Fatal(err)
+				}
+				if ar.AccessToken != accessToken {
+					t.Fatalf(`unexpected access token "%s"`, ar.AccessToken)
+				}
+				// silent auth should now succeed, provided no claims are requested, because the client has cached an access token
+				if method == "obo" {
+					ar, err = client.AcquireTokenOnBehalfOf(ctx, "assertion", tokenScope)
```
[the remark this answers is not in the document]
chlowell: This is to verify the client has a cached access token, because below we want to validate that `AcquireTokenOnBehalfOf` doesn't return this token when called with claims. (`AcquireTokenOnBehalfOf` handles both silent and non-silent auth; `AcquireTokenSilent` accesses a different cache and doesn't take a user assertion argument.)

apps/confidential/confidential.go
```diff
@@ -447,6 +458,47 @@ func WithLoginHint(username string) interface {
 	}
 }
 
+// WithClaims sets additional claims to request for the token, such as those required by conditional access policies.
+// Use this option when Azure AD returned a claims challenge for a prior request. The argument must be decoded.
+// This option is valid for any token acquisition method.
+func WithClaims(claims string) interface {
```
jhendrixMSFT, line 464: The implementation requires this to be a JSON object, we should doc that here.  In addition, the `string` type doesn't help enforce that contract.  What if we were to unmarshall the value into a JSON object here instead (then pass that through), would that make the error manifest earlier?

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -204,6 +210,84 @@ func (p AuthParams) WithTenant(ID string) (AuthParams, error) {
...
+	c := ClientCapabilities{}
+	var err error
+	if len(capabilities) > 0 {
+		cpbs := make([]string, len(capabilities))
+		for i := 0; i < len(cpbs); i++ {
+			cpbs[i] = fmt.Sprintf(`"%s"`, capabilities[i])
+		}
+		c.asJSON = fmt.Sprintf(`{"access_token":{"xms_cc":{"values":[%s]}}}`, strings.Join(cpbs, ","))
+		// note our JSON is valid but we can't stop users breaking it with garbage like "}"
+		err = json.Unmarshal([]byte(c.asJSON), &c.asMap)
```
jhendrixMSFT, line 286: Nit: seems like this could be done lazily as the map is only required if you need to merge.

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -204,6 +210,84 @@ func (p AuthParams) WithTenant(ID string) (AuthParams, error) {
...
+		} else {
+			// b does contain this key => recursively merge a[k] into b[k], provided both are maps. If a[k] or b[k] isn't
+			// a map, return an error because merging would overwrite some value in b. Errors shouldn't occur in practice
+			// because the challenge will be from AAD, which knows the capabilities format.
+			if A, ok := av.(map[string]any); ok {
+				if B, ok := bv.(map[string]any); ok {
+					return merge(A, B)
+				} else {
+					// b[k] isn't a map
+					return errors.New("challenge claims conflict with client capabilities")
```
jhendrixMSFT, line 254: How common do you think such a collision is?  Would it be would worth including details on the collision to help the caller fix the issue (could always be added later if it becomes a problem too)?

apps/internal/base/base.go
```diff
@@ -328,32 +365,33 @@ func (b Client) AcquireTokenByAuthCode(ctx context.Context, authCodeParams Acqui
...
 		AuthorizationType: authority.ATOnBehalfOf,
 		TenantID:          onBehalfOfParams.TenantID,
+		Claims:            onBehalfOfParams.Claims,
 	}
-	token, err := b.AcquireTokenSilent(ctx, silentParameters)
+	ar, err := b.AcquireTokenSilent(ctx, silentParameters)
+	if err == nil {
+		return ar, err
+	}
+	authParams, err := b.AuthParams.WithTenant(onBehalfOfParams.TenantID)
```
jhendrixMSFT, line 382: Previously, this was called before `b.AcquireTokenSilent()`.  Is the change in ordering significant?

apps/confidential/confidential.go
chlowell: We take a string because users get the value from AAD (by way of some RP) via a response header. They're supposed to simply send it back to AAD with a token request, which is the only useful thing they can do with it. In practice there shouldn't be an unmarshalling error because AAD is the sole author of these strings. Users have no reason to craft their own, and so the format and values aren't documented. Manifesting the error earlier would be awkward but I suppose we could do it. That wouldn't change the calling pattern or significantly shorten the time to failure though (an earlier error here wouldn't prevent a request, for example).

apps/internal/oauth/ops/authority/authority.go
chlowell: This is a hypothetical error. I think it would only happen to users who write their own claims challenge in ignorance of this merge step, which they have no reason to attempt because in normal usage AAD provides the challenge.

apps/internal/base/base.go
chlowell: This is an attempt to clarify the logic. `AcquireTokenSilent()` doesn't use `authParams`, so there's no reason to assign it before the call.

apps/confidential/confidential.go
jhendrixMSFT: I see, thanks for the explanation.

apps/confidential/confidential.go
```diff
@@ -531,15 +583,17 @@ func WithSilentAccount(account Account) interface {
...
-//   - [WithSilentAccount]
-//   - [WithTenantID]
+// Options: [WithClaims], [WithSilentAccount], [WithTenantID]
 func (cca Client) AcquireTokenSilent(ctx context.Context, scopes []string, opts ...AcquireSilentOption) (AuthResult, error) {
 	o := AcquireTokenSilentOptions{}
 	if err := options.ApplyOptions(&o, opts); err != nil {
 		return AuthResult{}, err
 	}
 
+	if o.claims != "" {
```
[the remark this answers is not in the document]
chlowell: You raise a good point. It will one day be possible for this method to succeed when given claims; we can add the option then. The value of taking the option today is in maintaining the calling pattern used in all other cases (see e.g. [azidentity](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/client_secret_credential.go#L51-L62)):
1. call `AcquireTokenSilent` (check the cache)
2. if that returned an error, call `AcquireTokenBy*` (request a new token)

Applications expect errors from `AcquireTokenSilent` because it commonly returns them, which is okay because that's just a cache miss. Having an option that guarantees it will return an error therefore doesn't make users write more, different, or tricky code. On the other hand, if this method doesn't take claims, the probably miniscule population of users who use this feature will have to write conditional code around it, then rewrite that code when this method does take claims. But sure, it's weird to have an anti-useful option.

Okay, how do you feel about it now? I lean toward keeping the option but might flop to the other side of the fence if you push hard enough 😆

apps/internal/oauth/ops/authority/authority.go
chlowell: It's not clear in the diff, but this function is called during client construction, and `capabilities` comes from the application. Unmarshalling here therefore creates a constructor error for a class of illegal user input values.

apps/confidential/confidential.go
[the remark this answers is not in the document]
richardpark-msft: I think I get what you mean - it's definitely not a straightforward case and it sounds like it's critical to consider that it's running as part of an overall pattern of calls, not just on its own (if I'm following your logic).

Resolving for now, explanation makes sense to me.

jhendrixMSFT (approved):

PR #357, opened 2022-10-26
Adds support for custom success page

### LINKED ISSUE #356: [Feature Request] Serve custom page on auth success
**Is your feature request related to a problem? Please describe.**
Currently the local server that's spawned for `AcquireTokenInteractive` always returns a static success page that's hardcoded at https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/e0d0f06e28b4211cd972f53ed9f611f9157cedc5/apps/internal/local/server.go#L17-L28

It'd be nice to be able to customise this on a per request basis in order to be able to communicate any other information that might be relevant to the user, or to theme the page as appropriate.

**Describe the solution you'd like**
An additional field in `AcquireInteractiveOption` to provide custom HTML to be used as the success page, with the default hardcoded one used as a fallback if that is not specified.

Phuurl:
Allows the success page displayed in browser after a successful `AcquireTokenInteractive` flow to be customised by the application.

Closes #356

### DIFF
--- a/apps/internal/local/server.go
+++ b/apps/internal/local/server.go
@@ -59,7 +59,7 @@ type Server struct {
 }
 
 // New creates a local HTTP server and starts it.
-func New(reqState string, port int) (*Server, error) {
+func New(reqState string, port int, okHtml []byte) (*Server, error) {
 	var l net.Listener
 	var err error
 	var portStr string
@@ -81,6 +81,10 @@ func New(reqState string, port int) (*Server, error) {
 	}
 	if err != nil {
 		return nil, err
+	}
+
+	if okHtml != nil {
+		okPage = okHtml
 	}
 
 	serv := &Server{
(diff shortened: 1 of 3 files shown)


Phuurl:
@chlowell @element-of-surprise @rayluo (sorry, just picked you from the most recent top contributors)

It's been a while since this PR was opened, and it's therefore drifted.

I'm happy to refactor it to align with the current state, but equally I don't particularly want to do that if it's never going to get merged 😅 

Given that there's a fair few 👍s on here it seems like it's something that users would like. 

I'm more than happy to refactor it and get a new working version for it (because hey, it's a feature that I'd really like, as well as seemingly others)), but again if there's no interaction / chance of it being merged then I won't bother 😞.

chlowell (commented):
Thanks for doing this. I think it's a good start on a useful feature.

PR #359, opened 2022-11-11
Use authority url port if it is specified to support private cloud sc…

sarathys:
Use authority url port if it is specified to support private cloud scenarios


sarathys:
https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/362

PR #360, opened 2022-11-13
Add support for disabling authority validation if needed to support private cloud scenarios

sarathys:


apps/confidential/confidential.go
```diff
@@ -283,6 +283,8 @@ type Options struct {
 
 	// Instructs MSAL Go to use an Azure regional token service with sepcified AzureRegion.
 	AzureRegion string
+	//Validate Authority
+	ValidateAuthority bool
```
bgavrilMS, line 287: We should not use the term `ValidateAuthority`, this is something that we are trying to deprecate. Consider using `IsInstanceDiscoveryEnabled`

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -213,9 +214,23 @@ func NewInfoFromAuthorityURI(authorityURI string, validateAuthority bool) (Info,
 		return Info{}, err
 	}
 
+	port := u.Port()
```
bgavrilMS, line 217: Does this not always come up as 443 ? (a unit test should cover it)

bgavrilMS, line 217: I would add a comment around this functionality, i.e. preserve the port number. This is a bug that can be fixes separately from this proposal.

apps/internal/oauth/resolvers.go
```diff
@@ -47,7 +47,9 @@ func newAuthorityEndpoint(rest *ops.REST) *authorityEndpoint {
 // ResolveEndpoints gets the authorization and token endpoints and creates an AuthorityEndpoints instance
 func (m *authorityEndpoint) ResolveEndpoints(ctx context.Context, authorityInfo authority.Info, userPrincipalName string) (authority.Endpoints, error) {
 	if authorityInfo.AuthorityType == ADFS && len(userPrincipalName) == 0 {
-		return authority.Endpoints{}, errors.New("UPN required for authority validation for ADFS")
+		if authorityInfo.ValidateAuthority {
+			return authority.Endpoints{}, errors.New("UPN required for authority validation for ADFS")
```
bgavrilMS, line 51: Authority validation for ADFS is a broken concept and should be disabled completely.

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -213,9 +214,23 @@ func NewInfoFromAuthorityURI(authorityURI string, validateAuthority bool) (Info,
...
+	hostname := u.Hostname()
+	cannonicalAuthorityURI = fmt.Sprintf("https://%v/%v/", u.Hostname(), tenant)
+	if port != "" {
+		hostname = fmt.Sprintf("%v:%v", hostname, port)
+		cannonicalAuthorityURI = fmt.Sprintf("https://%v/%v/", hostname, tenant)
+	}
+	// Note: temporary workaround to support private cloud scenarios while the following GitHub issues are addressed
+	// https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/301
+	// https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/289
+	disableAuthorityValidation, ok := os.LookupEnv("DISABLE_AUTHORITY_VALIDATION")
```
bgavrilMS, line 227: I'm not opposed to this, but not as a temporary solution. We should review and implement this in all MSALs if it makes sense.

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -213,9 +214,23 @@ func NewInfoFromAuthorityURI(authorityURI string, validateAuthority bool) (Info,
...
+	if ok {
+		validateAuthority = !(disableAuthorityValidation == "1")
```
bgavrilMS, line 229: If the user has already set `WithInstanceDiscovery(false)` won't this option override it?

apps/internal/oauth/resolvers.go
```diff
@@ -47,7 +47,9 @@ func newAuthorityEndpoint(rest *ops.REST) *authorityEndpoint {
 // ResolveEndpoints gets the authorization and token endpoints and creates an AuthorityEndpoints instance
 func (m *authorityEndpoint) ResolveEndpoints(ctx context.Context, authorityInfo authority.Info, userPrincipalName string) (authority.Endpoints, error) {
 	if authorityInfo.AuthorityType == ADFS && len(userPrincipalName) == 0 {
-		return authority.Endpoints{}, errors.New("UPN required for authority validation for ADFS")
+		if authorityInfo.ValidateAuthority {
```
bgavrilMS, line 50: Authority validation and instance discovery are different things, but on AAD we use the same mechanism to implement both. 

Authority validation - make a call to instance discovery endpoint to check that the authority host is valid. Only makes sense for AAD. For ADFS the mechanism was deemed unworkable. For B2C it doesn't exist.

Instance discovery - is also used to find authority aliases. E.g. "login.microsoft.com" is the same as "login.windows.net". This helps MSAL with caching. 


@rayluo can advise here, but the recommendation is to have a flag `instanceDIscoveryEnabled` that disables both concepts. In the future, we should deprecate / remove the concept of authority validation.

apps/internal/oauth/ops/authority/authority.go
sarathys: for private cloud STS comes up at port 3001

apps/internal/oauth/resolvers.go
sarathys: removing the code to validate authority for ADFS

apps/confidential/confidential.go
```diff
@@ -283,6 +283,9 @@ type Options struct {
 
 	// Instructs MSAL Go to use an Azure regional token service with sepcified AzureRegion.
 	AzureRegion string
+
+	// Disable instance discovery
+	IsInstanceDiscoveryEnabled bool
```
chlowell, line 288: When a field isn't explicitly assigned, its value is its type's default ("zero") value. The zero value of `bool` is `false`. This implementation therefore makes `New(clientID, cred)` the same as `New(clientID, cred, WithInstanceDiscovery(false))` i.e., it disables instance discovery by default.

So, I think `WithInstanceDiscovery` is the right user-facing API, but the internal option must be inverted. Personally, I would do
```go
func WithInstanceDiscovery(enabled bool) Option {
	return func(o *Options) {
		o.DisableInstanceDiscovery = !enabled
	}
}
```

rayluo: I agree with @chlowell 's point on "`WithInstanceDiscovery` is the right user-facing API" and the Instance Discovery behavior shall be enabled by default (to match the historical MSAL behavior).

I'll leave the implementation details to you Go experts to examine `o.DisableInstanceDiscovery = !enabled`. I am having a hard time to understand that implementation, and I would personally do an explicit pseudo code as `if enabled is a true or the default zero; then enable Instance Discovery`.

sarathys:
changes are here: https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/362

Issue #351, opened 2022-09-28
Prevent redundant discovery requests


chlowell:
Public and confidential clients send redundant discovery requests because silent and non-silent (audible? 😛) authentication request and cache discovery metadata independently. For example, `AcquireTokenOnBehalfOf`
first attempts silent authentication, which requests instance metadata and caches it with `storage.Manager`. When silent auth fails, it then requests a token from the STS via `oauth.Client`, which requests instance and tenant metadata and caches it on an instance of `oauth.authorityEndpoint`.

Issue #31, opened 2020-10-05
Add support for ADFS2019


sangonzal:
[Microsoft Docs ADFS documentation](https://docs.microsoft.com/en-us/windows-server/identity/ad-fs/overview/whats-new-active-directory-federation-services-windows-server)

bgavrilMS:
This is complete via https://github.com/AzureAD/microsoft-authentication-library-for-go/commit/dac2c13522a7c85a08d7f31e52477b5e357834de.

 App developers must explicitly disable instance discovery using `WithInstanceDiscovery(false)`

bgavrilMS:
Fixed in 0.8.1

PR #362, opened 2022-11-23
Add support for disabling instance discovery for AzureStack scenarios

sarathys:
Add support for disabling instance discovery for AzureStack and Custom Cloud scenarios

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -287,7 +287,7 @@ type Options struct {
 
 	capabilities []string
 
-	// Disable instance discovery
+	// To enable/disable instance discovery for AzureStack support
 	IsInstanceDiscoveryEnabled bool
 }
 
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -286,6 +286,9 @@ type Options struct {
 	AzureRegion string
 
 	capabilities []string
+
+	// Disable instance discovery
+	IsInstanceDiscoveryEnabled bool
 }
 
 func (o Options) validate() error {
@@ -340,6 +343,13 @@ func WithX5C() Option {
 	}
 }
 
+// WithInstanceDiscovery set to false to disable authority validation (to support private cloud scenarios)
+func WithInstanceDiscovery(enabled bool) Option {
+	return func(o *Options) {
+		o.IsInstanceDiscoveryEnabled = !enabled
+	}
+}
+
 // WithAzureRegion sets the region(preferred) or Confidential.AutoDetectRegion() for auto detecting region.
 // Region names as per https://azure.microsoft.com/en-ca/global-infrastructure/geographies/.
 // See https://aka.ms/region-map for more details on region names.
@@ -383,6 +393,7 @@ func New(clientID string, cred Credential, options ...Option) (Client, error) {
 		base.WithClientCapabilities(opts.capabilities),
 		base.WithRegionDetection(opts.AzureRegion),
 		base.WithX5C(opts.SendX5C),
+		base.WithInstanceDiscovery(opts.IsInstanceDiscoveryEnabled),
 	}
 	if cred.tokenProvider != nil {
 		// The caller will handle all details of authentication, using Client only as a token cache.
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -290,6 +290,10 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	authParams.AuthorizationType = silent.AuthorizationType
 	authParams.Claims = silent.Claims
 	authParams.UserAssertion = silent.UserAssertion
+	//if validate discovery is false, add authority to known hosts, so that Alias Lookup is not done
+	if(!authParams.AuthorityInfo.ValidateAuthority){
+		authParams.KnownAuthorityHosts = append(authParams.KnownAuthorityHosts, authParams.AuthorityInfo.Host)
+	}
 
 	var storageTokenResponse storage.TokenResponse
 	if authParams.AuthorizationType == authority.ATOnBehalfOf {
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -194,8 +194,9 @@ func WithRegionDetection(region string) Option {
 }
 
 func WithInstanceDiscovery(enabled bool) Option {
-	return func(c *Client) {
+	return func(c *Client) error {
 		c.AuthParams.AuthorityInfo.ValidateAuthority = !enabled
+		return nil
 	}
 }
 
@@ -291,7 +292,7 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	authParams.Claims = silent.Claims
 	authParams.UserAssertion = silent.UserAssertion
 	//if validate discovery is false, add authority to known hosts, so that Alias Lookup is not done
-	if(!authParams.AuthorityInfo.ValidateAuthority){
+	if !authParams.AuthorityInfo.ValidateAuthority {
 		authParams.KnownAuthorityHosts = append(authParams.KnownAuthorityHosts, authParams.AuthorityInfo.Host)
 	}
 
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -193,6 +193,12 @@ func WithRegionDetection(region string) Option {
 	}
 }
 
+func WithInstanceDiscovery(enabled bool) Option {
+	return func(c *Client) {
+		c.AuthParams.AuthorityInfo.ValidateAuthority = !enabled
+	}
+}
+
 // New is the constructor for Base.
 func New(clientID string, authorityURI string, token *oauth.Client, options ...Option) (Client, error) {
 	authInfo, err := authority.NewInfoFromAuthorityURI(authorityURI, true)
--- a/apps/internal/oauth/ops/authority/authority.go
+++ b/apps/internal/oauth/ops/authority/authority.go
@@ -312,6 +312,7 @@ func firstPathSegment(u *url.URL) (string, error) {
 func NewInfoFromAuthorityURI(authorityURI string, validateAuthority bool) (Info, error) {
 	authorityURI = strings.ToLower(authorityURI)
 	var authorityType string
+	var cannonicalAuthorityURI string
 	u, err := url.Parse(authorityURI)
 	if err != nil {
 		return Info{}, fmt.Errorf("authorityURI passed could not be parsed: %w", err)
@@ -331,9 +332,19 @@ func NewInfoFromAuthorityURI(authorityURI string, validateAuthority bool) (Info,
 		return Info{}, err
 	}
 
+	port := u.Port()
+	hostname := u.Hostname()
+	cannonicalAuthorityURI = fmt.Sprintf("https://%v/%v/", u.Hostname(), tenant)
+
+	// In private cloud deployments, STS can be deployed and running on a non 443 port, e.g. https://login.azs:3001
+	if port != "" {
+		hostname = fmt.Sprintf("%v:%v", hostname, port)
+		cannonicalAuthorityURI = fmt.Sprintf("https://%v/%v/", hostname, tenant)
+	}
+
 	return Info{
-		Host:                  u.Hostname(),
-		CanonicalAuthorityURI: fmt.Sprintf("https://%v/%v/", u.Hostname(), tenant),
+		Host:                  hostname,
+		CanonicalAuthorityURI: cannonicalAuthorityURI,
 		AuthorityType:         authorityType,
 		UserRealmURIPrefix:    fmt.Sprintf("https://%v/common/userrealm/", u.Hostname()),
 		ValidateAuthority:     validateAuthority,
--- a/apps/internal/oauth/resolvers.go
+++ b/apps/internal/oauth/resolvers.go
@@ -46,9 +46,6 @@ func newAuthorityEndpoint(rest *ops.REST) *authorityEndpoint {
 
 // ResolveEndpoints gets the authorization and token endpoints and creates an AuthorityEndpoints instance
 func (m *authorityEndpoint) ResolveEndpoints(ctx context.Context, authorityInfo authority.Info, userPrincipalName string) (authority.Endpoints, error) {
-	if authorityInfo.AuthorityType == ADFS && len(userPrincipalName) == 0 {
-		return authority.Endpoints{}, errors.New("UPN required for authority validation for ADFS")
-	}
 
 	if endpoints, found := m.cachedEndpoints(authorityInfo, userPrincipalName); found {
 		return endpoints, nil
--- a/apps/public/public.go
+++ b/apps/public/public.go
@@ -62,6 +62,9 @@ type Options struct {
 	HTTPClient ops.HTTPClient
 
 	capabilities []string
+
+	// To enable/disable instance discovery for AzureStack support
+	IsInstanceDiscoveryDisabled bool
 }
 
 func (p *Options) validate() error {
@@ -108,6 +111,13 @@ func WithHTTPClient(httpClient ops.HTTPClient) Option {
 	}
 }
 
+// WithInstanceDiscovery set to false to disable authority validation (to support private cloud scenarios)
+func WithInstanceDiscovery(instanceDiscoveryEnabled bool) Option {
+	return func(o *Options) {
+		o.IsInstanceDiscoveryDisabled = !instanceDiscoveryEnabled
+	}
+}
+
 // Client is a representation of authentication client for public applications as defined in the
 // package doc. For more information, visit https://docs.microsoft.com/azure/active-directory/develop/msal-client-applications.
 type Client struct {
@@ -128,7 +138,7 @@ func New(clientID string, options ...Option) (Client, error) {
 		return Client{}, err
 	}
 
-	base, err := base.New(clientID, opts.Authority, oauth.New(opts.HTTPClient), base.WithCacheAccessor(opts.Accessor), base.WithClientCapabilities(opts.capabilities))
+	base, err := base.New(clientID, opts.Authority, oauth.New(opts.HTTPClient), base.WithCacheAccessor(opts.Accessor), base.WithClientCapabilities(opts.capabilities), base.WithInstanceDiscovery(!opts.IsInstanceDiscoveryDisabled))
 	if err != nil {
 		return Client{}, err
 	}
--- a/apps/public/public.go
+++ b/apps/public/public.go
@@ -112,9 +112,9 @@ func WithHTTPClient(httpClient ops.HTTPClient) Option {
 }
 
 // WithInstanceDiscovery set to false to disable authority validation (to support private cloud scenarios)
-func WithInstanceDiscovery(instanceDiscoveryEnabled bool) Option {
+func WithInstanceDiscovery(enabled bool) Option {
 	return func(o *Options) {
-		o.IsInstanceDiscoveryDisabled = !instanceDiscoveryEnabled
+		o.IsInstanceDiscoveryDisabled = !enabled
 	}
 }
 
@@ -138,7 +138,7 @@ func New(clientID string, options ...Option) (Client, error) {
 		return Client{}, err
 	}
 
-	base, err := base.New(clientID, opts.Authority, oauth.New(opts.HTTPClient), base.WithCacheAccessor(opts.Accessor), base.WithClientCapabilities(opts.capabilities), base.WithInstanceDiscovery(!opts.IsInstanceDiscoveryDisabled))
+	base, err := base.New(clientID, opts.Authority, oauth.New(opts.HTTPClient), base.WithCacheAccessor(opts.Accessor), base.WithClientCapabilities(opts.capabilities), base.WithInstanceDiscovery(opts.IsInstanceDiscoveryDisabled))
 	if err != nil {
 		return Client{}, err
 	}


apps/public/public.go
```diff
@@ -62,6 +62,9 @@ type Options struct {
 	HTTPClient ops.HTTPClient
 
 	capabilities []string
+
+	// Default is Instance discovery is enabled, unless disabled by setting DisableInstanceDiscovery = true
+	DisableInstanceDiscovery bool
```
chlowell, line 67: Recommend not exporting this field, because applications have no reason to set it
```suggestion
	disableInstanceDiscovery bool
```

apps/confidential/confidential.go
```diff
@@ -286,6 +286,9 @@ type Options struct {
 	AzureRegion string
 
 	capabilities []string
+
+	// Default is Instance discovery is enabled, unless disabled by setting DisableInstanceDiscovery = true
+	DisableInstanceDiscovery bool
```
chlowell, line 291: Same thought here: applications have no reason to set this field, so we have no reason to export it
```suggestion
	disableInstanceDiscovery bool
```

apps/internal/base/base.go
```diff
@@ -284,6 +291,10 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	authParams.AuthorizationType = silent.AuthorizationType
 	authParams.Claims = silent.Claims
 	authParams.UserAssertion = silent.UserAssertion
+	//if validate discovery is false, add authority to known hosts, so that Alias Lookup is not done
+	if !authParams.AuthorityInfo.ValidateAuthority {
```
bgavrilMS, line 295: I think there is a logical error here: 

- if `instance_discovery` is disabled, then `validate_authority` is also disabled (correct, since we cannot valdiate the authority)
- but not the other way around - if `validate_authority` is disabled, then it doesn't mean that the authority is known. In particular, MSALs should still perform instance discovery to find out authority aliases.

We should disable / hide / deprecate the validate_authority flag altogether, but this can be done separately. 

For now, I recommend changing this to `if !authParams.disableInstanceDiscovery == true { `

apps/public/public.go
```diff
@@ -108,6 +111,13 @@ func WithHTTPClient(httpClient ops.HTTPClient) Option {
 	}
 }
 
+// WithInstanceDiscovery set to false to disable authority validation (to support private cloud scenarios)
+func WithInstanceDiscovery(enabled bool) Option {
```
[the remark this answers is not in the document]
chlowell: [TestWithTenantID](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/4abfea36f99928ad52523e017b804616993470f5/apps/confidential/confidential_test.go#L713) is a good starting point. A test for this option would be much simpler. It can just run through the happy path for each AcquireToken method and assert that it got the expected token. If the client doesn't send the expected sequence of requests set on the mock client, you would see an error from the AcquireToken method or a panic.

apps/confidential/confidential_test.go
```diff
@@ -107,6 +107,25 @@ func fakeClient(tk accesstokens.TokenResponse, credential Credential, options ..
...
+	options = append(options, WithAuthority("https://fake_authority/fake"))
+	options = append(options, WithInstanceDiscovery(false))
+	client, err := New("fake_client_id", credential, options...)
+	if err != nil {
+		return Client{}, err
+	}
+	client.base.Token.AccessTokens = &fake.AccessTokens{
+		AccessToken: tk,
+	}
+	client.base.Token.Authority = &fake.Authority{}
```
chlowell, line 120: `Authority` is responsible for instance discovery, so this fake is replacing code we want to test. I recommend instead using the `mock` package to test an unmodified client end-to-end, through its public API, as in [TestWithTenantID](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/4abfea36f99928ad52523e017b804616993470f5/apps/confidential/confidential_test.go#L713).

apps/tests/devapps/instance_discovery_disabled_sample.go
```diff
@@ -0,0 +1,105 @@
+package main
```
chlowell, line 1: This file needs a copyright header. Taking a step back though, what's this file for? The user API seems simple enough to not benefit from samples, and the functions here seem to require additional configuration to serve as tests.

jonnguyen1: I think previously sarathy and I thought that this was meant to be the test code, so we added this sample to run. Can remove if you want.

bganapa (approved):

apps/tests/devapps/instance_discovery_disabled_sample.go
chlowell: Sounds good to me; the new option has unit tests and better documentation elsewhere.

apps/internal/base/base.go
```diff
@@ -380,6 +393,12 @@ func (b Client) AcquireTokenOnBehalfOf(ctx context.Context, onBehalfOfParams Acq
 		return ar, err
 	}
 	authParams, err := b.AuthParams.WithTenant(onBehalfOfParams.TenantID)
+	//if instance discovery is disabled, add authority to known hosts, so that Alias Lookup is not done
+	//withTenant seems to reset the previous AuthParams, so have to add this knownAuthorityHosts field again
+	if authParams.AuthorityInfo.InstanceDiscoveryDisabled {
+		authParams.KnownAuthorityHosts = append(authParams.KnownAuthorityHosts, authParams.AuthorityInfo.Host)
```
chlowell, line 399: If this is necessary, it conceals a bug. `b.AuthParams.WithTenant(...)` returns a copy of `b.AuthParams` having the specified tenant; it doesn't modify `KnownAuthorityHosts`. If the copy has an empty or otherwise unexpected value for `KnownAuthorityHosts`, I expect `b.AuthParams` to have the same unexpected value.

apps/confidential/confidential_test.go
```diff
@@ -812,6 +964,105 @@ func TestWithTenantID(t *testing.T) {
...
+	uuid1 := "00000000-0000-0000-0000-000000000000"
+	//uuid2 := strings.ReplaceAll(uuid1, "0", "1")
+	lmo := "login.microsoftonline.com"
+	host := fmt.Sprintf("https://%s/", lmo)
+	for _, test := range []struct {
+		authority, expectedAuthority, tenant string
+		expectError                          bool
+	}{
+		{authority: host + "common", tenant: uuid1, expectedAuthority: host + uuid1},
+		/*{authority: host + "organizations", tenant: uuid1, expectedAuthority: host + uuid1},
```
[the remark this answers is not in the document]
jonnguyen1: Ah, I had commented these out while debugging, thought I uncommented it. Fixing.

apps/confidential/confidential_test.go
```diff
@@ -812,6 +964,105 @@ func TestWithTenantID(t *testing.T) {
...
+		{authority: host + uuid1, tenant: uuid2, expectedAuthority: host + uuid2},
+		{authority: host + uuid1, tenant: "common", expectError: true},
+		{authority: host + uuid1, tenant: "organizations", expectError: true},
+		{authority: host + "consumers", tenant: uuid1, expectError: true},*/
+	} {
+		for _, method := range []string{ /*"authcode", "authcodeURL", "credential",*/ "obo"} {
```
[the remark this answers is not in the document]
jonnguyen1: Same as above

apps/internal/base/base.go
jonnguyen1: What you said is correct. I think just InstanceDiscoveryDisabled was being defaulted to false. Removing this line ran the tests fine.

apps/confidential/confidential_test.go
```diff
@@ -710,6 +710,158 @@ func TestWithClaims(t *testing.T) {
 	}
 }
 
+func TestWithClaimsNoInstanceDiscovery(t *testing.T) {
```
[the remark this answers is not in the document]
jonnguyen1: Removed

apps/confidential/confidential_test.go
```diff
@@ -812,6 +964,105 @@ func TestWithTenantID(t *testing.T) {
 	}
 }
 
+func TestWithTenantIDNoInstanceDiscovery(t *testing.T) {
```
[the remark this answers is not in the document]
jonnguyen1: TenantID creates a new authInfo in the authparam, which previously reset instanceDiscovery flag. I think it's worth keeping this test.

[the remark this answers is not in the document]
chlowell: That's a bug in the `authority.Info` constructor surfacing in `AuthParams.WithTenant()`. It's unrelated to combining these options because the copy returned by `WithTenant()` gets used whether the options are set or not. It would be better to catch this bug in a test of either `WithTenant` or the constructor ([here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/4abfea36f99928ad52523e017b804616993470f5/apps/internal/oauth/ops/authority/authority_test.go#L322) or [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/oauth/ops/authority/authority_test.go#L301)). I lean toward the former because `WithTenant` is responsible for returning a complete copy whereas the constructor doesn't have, and isn't necessarily required to have, a parameter for each of the struct's fields, though perhaps it should (`authority.Info` is awkward about this; problem for a future PR).

apps/tests/devapps/instance_discovery_disabled_sample.go
chlowell: I meant I'm in favor of deleting this file because it lacks purpose

apps/confidential/confidential_test.go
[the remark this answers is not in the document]
jonnguyen1: That makes sense, I can remove the withTenantIdNoInstance test. I think inherently the "bug" is only showing up because I added the InstanceDiscovery field to the authInfo struct. In that regard, having it defaulted to false wouldn't affect anything else since that flag is only important to cases where instanceDiscovery is disabled.

apps/tests/devapps/instance_discovery_disabled_sample.go
jonnguyen1: Deleted

apps/tests/devapps/sample_utils.go
```diff
@@ -23,6 +23,7 @@ type Config struct {
 	ClientSecret        string   `json:"client_secret"`
 	Thumbprint          string   `json:"thumbprint"`
 	PemData             string   `json:"pem_file"`
+	InstanceDiscovery   string   `json:"instance_discovery"`
```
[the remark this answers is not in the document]
jonnguyen1: Yep. Deleting

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -331,13 +332,15 @@ func NewInfoFromAuthorityURI(authorityURI string, validateAuthority bool) (Info,
...
+	// u.Host includes the port, if any, which is required for private cloud deployments
 	return Info{
-		Host:                  u.Hostname(),
-		CanonicalAuthorityURI: fmt.Sprintf("https://%v/%v/", u.Hostname(), tenant),
-		AuthorityType:         authorityType,
-		UserRealmURIPrefix:    fmt.Sprintf("https://%v/common/userrealm/", u.Hostname()),
-		ValidateAuthority:     validateAuthority,
-		Tenant:                tenant,
+		Host:                      u.Host,
+		CanonicalAuthorityURI:     fmt.Sprintf("https://%v/%v/", u.Host, tenant),
```
chlowell, line 338: Sorry for not noticing this earlier but we should test that clients behave correctly when the given authority includes a port. I think it would be enough to test the simplest AcquireToken method plus silent auth (the authority is used as a cache key), verifying URLs include the port.

apps/public/public_test.go
```diff
@@ -237,6 +237,80 @@ func TestAcquireTokenWithTenantID(t *testing.T) {
...
+				}
+				if err != nil {
+					t.Fatal(err)
+				}
+				if method == "devicecode" {
+					if ar, err = dc.AuthenticationResult(ctx); err != nil {
+						t.Fatal(err)
+					}
+				}
+				if ar.AccessToken != accessToken {
```
bgavrilMS, line 300: I think 2 extra things need to be checked: 

1. HomeAccountId property of Account objects from should be in format "oid.oid" (for AAD it's "oid.tid" but we don't have "tid" in ADFS). This is pretty important, otherwise other MSALs might not be able to read from the cache.
2. Account object (which is populated from the id token) should have the username set. I believe ADFS will expose this in the "upn" claim and AAD in the "preferred_username" claim. IMO, I would set Username to `idToken.preferred_username ?? idToken.upn`

bgavrilMS: CC @chlowell @rayluo

jonnguyen1: We are setting HomeAccountID like this currently. Do we have to make changes to this? And is this something that should go through in this pr? 

// HomeAccountID creates the home account ID.
func (c ClientInfo) HomeAccountID() string {
	if c.UID == "" || c.UTID == "" {
		return ""
	}
	return fmt.Sprintf("%s.%s", c.UID, c.UTID)
}

bgavrilMS: I think so, otherwise you risk introducing a security issue or a cache miss issue. If home account Id is "", MSAL might not be able to retrieve the correct token from the cache (or it might not be able to retrieve one at all). 

if think the logic is: 

- if both are emtpy, homeAccountId is empty (this is probably for client_credentials)
- if c.Uitd == "", then use `fmt.Sprintf("%s.%s", c.UID, c.UID)`

rayluo: > 2 extra things need to be checked:
>
> * HomeAccountId property of Account objects from should be in format "oid.oid" (for AAD it's "oid.tid" but we don't have "tid" in ADFS). This is pretty important, otherwise other MSALs might not be able to read from the cache.

AFAIK, this is an undefined corner case in the original Token Cache Schema. Existing MSALs probably already have inconsistent behaviors here, ranging from "oid"-only to "oid.adfs". The effect is indeed different MSALs would not hit the same account in the cache. Not sure whether it is prevalent in the wild. We did not hear report on this. It is ADFS, anyway.

But if MSAL Go currently returns a completely empty string as HomeAccountID, that sounds too extreme. Better still return something, at least the oid.

> * Account object (which is populated from the id token) should have the username set. I believe ADFS will expose this in the "upn" claim and AAD in the "preferred_username" claim. IMO, I would set Username to `idToken.preferred_username ?? idToken.upn`

Agreed. [MSAL Python did it the same way](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/1.20.0/msal/token_cache.py#L15-L18), too.

> And is this something that should go through in this pr?

Not necessarily in this PR, but if it would break Azure Stack scenarios, we may need to fix it sooner or later.

jonnguyen1: Then there would be cases where we would have UID.UID in AAD, when we want "" right? Does this pose some type of security issue for AAD? If so, sounds like we should distinguish between ADFS and AAD in the function/make a function for each auth type.

apps/internal/oauth/resolvers.go
```diff
@@ -46,9 +46,6 @@ func newAuthorityEndpoint(rest *ops.REST) *authorityEndpoint {
 
 // ResolveEndpoints gets the authorization and token endpoints and creates an AuthorityEndpoints instance
 func (m *authorityEndpoint) ResolveEndpoints(ctx context.Context, authorityInfo authority.Info, userPrincipalName string) (authority.Endpoints, error) {
-	if authorityInfo.AuthorityType == ADFS && len(userPrincipalName) == 0 {
-		return authority.Endpoints{}, errors.New("UPN required for authority validation for ADFS")
-	}
```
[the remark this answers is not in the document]
jonnguyen1: CC: @sarathys since these are his changes

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -290,13 +290,14 @@ func NewClientCapabilities(capabilities []string) (ClientCapabilities, error) {
...
-	Tenant                string
-	Region                string
+	Host                      string
+	CanonicalAuthorityURI     string
+	AuthorityType             string
+	UserRealmURIPrefix        string
+	ValidateAuthority         bool
+	Tenant                    string
+	Region                    string
+	InstanceDiscoveryDisabled bool
```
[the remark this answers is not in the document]
jonnguyen1: Seems like a gofmt thing, so I would assume it is somewhat built into the styling.

apps/public/public_test.go
bgavrilMS: In AAD, the client info has `uid` and `utid` claims. In ADFS, it only has `uid` claim. You don't need to know if authority is ADFS to see that the client_info does not have `utid` claim. 

Just avoid setting HomeAccountId to "" if the `uid` claim is present. This is a potential security issue, because the cache logic will not be able to differentiate between users.

apps/confidential/confidential_test.go
```diff
@@ -812,6 +812,119 @@ func TestWithTenantID(t *testing.T) {
...
+	authority := fmt.Sprintf("https://%s%s/%s", lmo, port, tenant)
+	cred, err := NewCredFromSecret("secret")
+	if err != nil {
+		t.Fatal(err)
+	}
+	idToken, refreshToken, URL := "", "", ""
+	mockClient := mock.Client{}
+	//2 calls to instance discovery are made because Host is not trusted
+	mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(host, tenant)))
+	mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(host, tenant)))
```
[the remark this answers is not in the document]
jonnguyen1: One request is sent out during the silentToken request, and the other is sent during the AuthCode request. Since the host isn't trusted, two requests are sent.

apps/confidential/confidential_test.go
```diff
@@ -812,6 +812,119 @@ func TestWithTenantID(t *testing.T) {
 	}
 }
 
+func TestWithInstanceDiscovery(t *testing.T) {
+	accessToken := "*"
+	stackurl := "stack.local"
+	host := fmt.Sprintf("https://%s/", stackurl)
```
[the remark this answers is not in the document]
jonnguyen1: Fixed

apps/internal/base/internal/storage/partitioned_storage.go
```diff
@@ -41,35 +41,40 @@ func (m *PartitionedManager) Read(ctx context.Context, authParameters authority.
 	clientID := authParameters.ClientID
 	scopes := authParameters.Scopes
 
-	metadata, err := m.getMetadataEntry(ctx, authParameters.AuthorityInfo)
-	if err != nil {
-		return TokenResponse{}, err
+	// fetch metadata if and only if the authority isn't explicitly trusted
+	aliases := authParameters.KnownAuthorityHosts
+	if len(aliases) == 0 {
```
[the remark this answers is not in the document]
jonnguyen1: Yeah. Sounds like we can do this as the conditional: 
`if authParameters.Host is not in KnownAuthorityHosts || instanceDiscoveryEnabled ` do lookup

apps/internal/base/internal/storage/partitioned_storage.go
```diff
@@ -148,13 +159,20 @@ func (m *PartitionedManager) Write(authParameters authority.AuthParams, tokenRes
 		localAccountID := idTokenJwt.LocalAccountID()
 		authorityType := authParameters.AuthorityInfo.AuthorityType
 
+		var preferredUsername string
+		if idTokenJwt.PreferredUsername != "" {
+			preferredUsername = idTokenJwt.PreferredUsername
+		} else {
+			preferredUsername = idTokenJwt.UPN
+		}
```
chlowell, line 167: Since there's only one condition, you can simply initialize with the value the `else` branch would assign:
```suggestion
		preferredUsername := idTokenJwt.UPN
		if idTokenJwt.PreferredUsername != "" {
			preferredUsername = idTokenJwt.PreferredUsername
		}
```

apps/internal/base/base.go
```diff
@@ -293,10 +293,6 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	authParams.AuthorizationType = silent.AuthorizationType
 	authParams.Claims = silent.Claims
 	authParams.UserAssertion = silent.UserAssertion
-	//if instance discovery is disabled, add authority to known hosts, so that Alias Lookup is not done
-	if authParams.AuthorityInfo.InstanceDiscoveryDisabled {
```
chlowell: What's the benefit of moving this into the storage implementations?

apps/confidential/confidential_test.go
```diff
@@ -246,6 +246,76 @@ func TestAcquireTokenByAuthCode(t *testing.T) {
...
+		t.Fatalf("unexpected access token %s", tk.AccessToken)
+	}
+}
+
+func TestAcquireTokenByAuthCodeADFS(t *testing.T) {
+	cred, err := NewCredFromSecret("fake_secret")
+	if err != nil {
+		t.Fatal(err)
+	}
+	client, err := fakeClient(accesstokens.TokenResponse{
```
chlowell, line 270: What quirks of the ADFS auth code flow should this test cover? This fake replaces the real code for authority validation, authentication, caching, etc., so I wonder whether this test would catch a production bug related to those quirks.

apps/confidential/confidential_test.go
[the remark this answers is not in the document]
chlowell: I opened #351 for this kind of thing

apps/confidential/confidential_test.go
jonnguyen1: This test just checks if homeAccountId is being properly set to oid.oid or oid.uid, as suggested here 
>HomeAccountId property of Account objects from should be in format "oid.oid" (for AAD it's "oid.tid" but we don't have "tid" in ADFS). This is pretty important, otherwise other MSALs might not be able to read from the cache.

apps/internal/base/base.go
jonnguyen1: I figured it'd be clearer if the alias was added during the flow that required the alias to be in the KnownAuthorityHosts.

apps/internal/base/internal/storage/partitioned_storage.go
```diff
@@ -41,35 +41,47 @@ func (m *PartitionedManager) Read(ctx context.Context, authParameters authority.
...
+			knownAuthority = true
+			break
+		}
+	}
+	if !knownAuthority && !authParameters.AuthorityInfo.InstanceDiscoveryDisabled {
+		metadata, err := m.getMetadataEntry(ctx, authParameters.AuthorityInfo)
+		if err != nil {
+			return TokenResponse{}, err
+		}
+		aliases = metadata.Aliases
```
chlowell, line 58: I believe this could be much simpler. Functionally, `aliases` are the hosts considered to match this token request. The requested host, `authParameters.AuthorityInfo.Host`, is of course always a match. We do need to know any aliases of that host to prevent unnecessary cache misses. The client should discover these from instance metadata, but only if instance discovery is enabled. How about this then:
```go
// aliases are the hosts matching this auth attempt. The requested host always matches. If
// instance discovery is permitted, request instance metadata to discover any other aliases.
aliases := []string{authParameters.AuthorityInfo.Host}
if !authParameters.AuthorityInfo.InstanceDiscoveryDisabled {
	metadata, err := m.getMetadataEntry(ctx, authParameters.AuthorityInfo)
	if err != nil {
		return TokenResponse{}, err
	}
	aliases = metadata.Aliases
}
```
With that, the caller doesn't need to set `KnownAuthorities`. If we also replace `WithKnownAuthorityHosts` with `WithInstanceDiscovery(false)` [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/16c062c187ac55067ece02de9fe65f2f7b4033cd/apps/confidential/confidential.go#L396), we no longer need the field or its setter, which is a win because that API is a vestige of a rejected design.

apps/confidential/confidential.go
```diff
@@ -399,11 +399,11 @@ func New(clientID string, cred Credential, options ...Option) (Client, error) {
 		// Declaring the authority host known prevents unnecessary metadata discovery requests. (The
 		// authority is irrelevant to Client and friends because the token provider is responsible
 		// for authentication.)
-		parsed, err := url.Parse(opts.Authority)
+		_, err := url.Parse(opts.Authority)
```
chlowell, line 402: This change makes parsing the authority and most of the comment vestigial. This block can simply be
```go
// The caller will handle all details of authentication, using Client only as a token cache.
baseOpts = append(baseOpts, base.WithInstanceDiscovery(false))
```

apps/internal/oauth/resolvers.go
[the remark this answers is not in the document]
chlowell: Looking at the current implementation, I think we have bugs here and need more test coverage. The UPN is apparently supposed to serve two purposes but in practice does not:
1. Determining the OID config endpoint. The helper that gets the endpoint takes a UPN argument but doesn't use it; see [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/16c062c187ac55067ece02de9fe65f2f7b4033cd/apps/internal/oauth/resolvers.go#L125).
2. Retrieving cached endpoints. Again, the UPN is irrelevant; see [here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/16c062c187ac55067ece02de9fe65f2f7b4033cd/apps/internal/oauth/resolvers.go#L88-L97), and note the function returns the first cache entry matching the authority whether it matches the UPN's domain or not.

That said, I don't know how this is supposed to work: is a UPN required for authority validation on ADFS, as the current error states? How, if at all, should the client cache ADFS endpoints?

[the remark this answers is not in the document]
bgavrilMS: @chlowell - ADFS authority validation is different than AAD authority validation because there is no centralized endpoint which knows all the authroities like in AAD world. In practical terms there is no authority validation possible on ADFS. Other MSALs do not implement any such validation. 

There are 2 choices: 

- if MSAL knows the authority is ADFS (because it ends with "adfs") then it ignores the flag
- if MSAL is authority type agnostic (JS and Py are), then app developer must set the flag to false by themsevels.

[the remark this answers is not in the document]
chlowell: Then I think we can consider cleaning up this API and adding test coverage here part of #31.

Do you prefer one choice over the other? MSAL Go isn't authority type agnostic, but it might be possible to make it so before it officially supports ADFS.

apps/confidential/confidential_test.go
```diff
@@ -204,55 +204,79 @@ func TestAcquireTokenByAuthCode(t *testing.T) {
...
+				t.Fatal(err)
+			}
+			if tk.AccessToken != token {
+				t.Fatalf("unexpected access token %s", tk.AccessToken)
+			}
+			account := client.Account(tk.Account.HomeAccountID)
+			if params.utid == "" {
+				if account.HomeAccountID != "123-456.123-456" {
+					t.Fatal("Unexpected Account.HomeAccountId")
+				}
```
chlowell, line 262: Just a minor nitpick--including the expected and actual values in failure messages helps future debuggers figure out what went wrong:
```suggestion
				if actual := account.HomeAccountID; actual != "123-456.123-456" {
					t.Fatalf("expected %q, got %q", "123-456.123-456", actual)
				}
```
(tidier if the expected value comes from a variable instead of a literal)

chlowell (approved):

rayluo (approved):

PR #363, opened 2022-12-02
Add Domain Hint

eopeter:
This PR Seeks to enable adding Domain Hint option to be used to accelerate the user to their federated IdP sign-in page. Or they can be used by a multi-tenant application to accelerate the user straight to the branded Azure AD sign-in page for their tenant.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -459,7 +459,7 @@ func WithLoginHint(username string) interface {
 	}
 }
 
-// WithDomainHint pre-populates the login prompt with the IdP domain.
+// WithDomainHint adds the IdP domain as domain_hint query parameter in the auth url.
 func WithDomainHint(domain string) interface {
 	AuthCodeURLOption
 	options.CallOption
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -420,7 +420,7 @@ type AuthCodeURLOption interface {
 
 // AuthCodeURL creates a URL used to acquire an authorization code. Users need to call CreateAuthorizationCodeURLParameters and pass it in.
 //
-// Options: [WithClaims], [WithLoginHint], [WithDomainHint], [WithTenantID]
+// Options: [WithClaims], [WithDomainHint], [WithLoginHint], [WithTenantID]
 func (cca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, scopes []string, opts ...AuthCodeURLOption) (string, error) {
 	o := authCodeURLOptions{}
 	if err := options.ApplyOptions(&o, opts); err != nil {
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -410,7 +410,7 @@ func (cca Client) UserID() string {
 
 // authCodeURLOptions contains options for AuthCodeURL
 type authCodeURLOptions struct {
-	claims, loginHint, tenantID string
+	claims, loginHint, tenantID, domainHint string
 }
 
 // AuthCodeURLOption is implemented by options for AuthCodeURL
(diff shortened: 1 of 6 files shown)


apps/confidential/confidential.go
```diff
@@ -458,6 +459,29 @@ func WithLoginHint(username string) interface {
 	}
 }
 
+// WithDomainHint pre-populates the login prompt with the IdP domain.
```
rayluo, line 462: Will there be a situation that the app developer wants to use the default "without domain hint" behavior? If we pre-populate the domain hint with the IdP domain, how does the app developer opt out?

apps/public/public.go
```diff
@@ -531,6 +532,33 @@ func WithLoginHint(username string) interface {
 	}
 }
 
+// WithDomainHint pre-populates the login prompt with the IdP domain.
+func WithDomainHint(domain string) interface {
```
rayluo, line 536: Nit: Is it possible to refactor our code base so that we wouldn't need to duplicate this code snippet twice?

CC: @chlowell

rayluo (commented):
CC: @chlowell @bgavrilMS

apps/confidential/confidential.go
eopeter: this comment was wrong and has been fixed. Should have said 
`// WithDomainHint adds the IdP domain as domain_hint query parameter in the auth url.`

It is not prepopulated and the user has to use the `client.WithDomainHint("contoso.com")` to ask to use domain_hint in the auth url

apps/public/public.go
eopeter: I refactored to avoid the code duplication

rayluo: Oh, thanks for consolidating the new WithDomainHint() and existing WithLoginHint() with a common underlying helper. My original comment was about the WET between confidential client and public client, though. All the implementation and their corresponding test case are currently duplicated in the two clients, while the behavior could ideally be shared among a common base of both confidential client and public client.

Not sure how much refactoring would be needed. I'll defer this topic to @chlowell .

chlowell: I think `withHint()` and its supporting type make the code more complex for no benefit--this approach doesn't lower the maintenance burden. It's better for `WithLoginHint()` and `WithDomainHint()` to simply set their respective fields directly.

Sharing one implementation of `WithDomainHint()` between confidential and public clients might be possible, but the constraints imposed by Go and the current implementation make it difficult and much more complex than parallel options in each package. I don't think it's worth pursuing.

Regarding a shared base for both client types, we do have one, but testing it isn't a substitute for testing the public API.

eopeter: Ok, I reverted the `WithHint()` code and now have separate `WithLoginHint()` and `WithDomainHint()`

apps/public/public.go
```diff
@@ -147,7 +147,7 @@ type CreateAuthCodeURLOption interface {
 
 // CreateAuthCodeURL creates a URL used to acquire an authorization code.
 //
-// Options: [WithClaims], [WithLoginHint], [WithTenantID]
+// Options: [WithClaims], [WithLoginHint], [WithDomainHint], [WithTenantID]
```
chlowell, line 150:
```suggestion
// Options: [WithClaims], [WithDomainHint], [WithLoginHint], [WithTenantID]
```

apps/public/public.go
```diff
@@ -557,7 +585,7 @@ func WithRedirectURI(redirectURI string) interface {
 // AcquireTokenInteractive acquires a security token from the authority using the default web browser to select the account.
 // https://docs.microsoft.com/en-us/azure/active-directory/develop/msal-authentication-flows#interactive-and-non-interactive-authentication
 //
-// Options: [WithLoginHint], [WithRedirectURI], [WithTenantID]
+// Options: [WithLoginHint], [WithDomainHint], [WithRedirectURI], [WithTenantID]
```
chlowell, line 588:
```suggestion
// Options: [WithDomainHint], [WithLoginHint], [WithRedirectURI], [WithTenantID]
```

chlowell (approved):
Looks good to me as Go. I'm not familiar with the AAD feature, so defer to @rayluo for MSAL perspective.

rayluo (approved):

PR #366, opened 2022-12-16
Multitenant silent authentication fixes

chlowell:
This fixes a couple bugs:
1. Silent auth defaults to the given account's tenant instead of the client's configured tenant.
1. Silent auth fails when the cache lacks an ID token from the requested tenant even though the cache contains a valid refresh token. My fix here is to have the cache return all the matching data it has instead of returning an error on the first miss. The caller is responsible for deciding whether the returned data is sufficient for its purpose.

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -271,11 +271,8 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
 }
 
 func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilentParameters) (AuthResult, error) {
-	tenant := silent.TenantID
-	if tenant == "" {
-		tenant = silent.Account.Realm
-	}
-	authParams, err := b.AuthParams.WithTenant(tenant)
+	// when silent.TenantID == "", this is a no-op and we use the client application's configured tenant
+	authParams, err := b.AuthParams.WithTenant(silent.TenantID)
 	if err != nil {
 		return AuthResult{}, err
 	}
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -271,8 +271,27 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
 }
 
 func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilentParameters) (AuthResult, error) {
-	// when silent.TenantID == "", this is a no-op and we use the client application's configured tenant
-	authParams, err := b.AuthParams.WithTenant(silent.TenantID)
+	tenant := silent.TenantID
+	if tenant == "" {
+		// the caller didn't specify a tenant, so we'll use the client's configured tenant or the given account's home tenant
+		switch tenant = b.AuthParams.AuthorityInfo.Tenant; tenant {
+		case "common", "organizations":
+			if _, homeTenant, found := strings.Cut(silent.Account.HomeAccountID, "."); found {
+				// note that both public and confidential clients allow specifying an account for silent auth
+				tenant = homeTenant
+			} else if !b.AuthParams.IsConfidentialClient {
+				// public client requires the caller to identify a specific user for silent authentication
+				return AuthResult{}, errors.New("use the WithSilentAccount option to specify an account")
+			}
+			// else we have a confidential client and no account specified. We can't return an error here because
+			// the caller may have configured the client with a custom token provider, in which case the client
+			// handles caching and the token provider is responsible for everything else, including determining
+			// the correct tenant.
+		default:
+			// use the client's configured tenant
+		}
+	}
+	authParams, err := b.AuthParams.WithTenant(tenant)
 	if err != nil {
 		return AuthResult{}, err
 	}
--- a/apps/internal/base/internal/storage/storage.go
+++ b/apps/internal/base/internal/storage/storage.go
@@ -84,6 +84,7 @@ func isMatchingScopes(scopesOne []string, scopesTwo string) bool {
 
 // Read reads a storage token from the cache if it exists.
 func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams, account shared.Account) (TokenResponse, error) {
+	tr := TokenResponse{}
 	homeAccountID := authParameters.HomeAccountID
 	realm := authParameters.AuthorityInfo.Tenant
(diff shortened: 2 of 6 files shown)


apps/internal/base/internal/storage/storage.go
```diff
@@ -84,6 +84,7 @@ func isMatchingScopes(scopesOne []string, scopesTwo string) bool {
 
 // Read reads a storage token from the cache if it exists.
 func (m *Manager) Read(ctx context.Context, authParameters authority.AuthParams, account shared.Account) (TokenResponse, error) {
+	tr := TokenResponse{}
```
ellismg, line 87: Do you think there is value in still returning an error in the case where we fail to read some parts of `TokenResponse` to communicate to our caller that we were only able to do a partial read?  In general, I know you shouldn't touch any of the other returned values when error is non nil, but I've seen some go APIs that document what values are correct when an error is returned - I imagine that we could say we could trust an non-zero part of the structure.

ellismg (approved):
Thanks for fixing this @chlowell.  I verified that it fixes my larger end to end scenario and thanks for taking my repo and turning it into a test.

FWIW - It looks like some of the option types have been renamed (and the signature of some public methods changed) which meant I had to do a little surgery both in the `azidentity` package and my code (which uses MSAL directly) to adopt the breaking changes.  Not sure if you intend to roll the back on not.

Thanks again for your help here!

apps/internal/base/internal/storage/storage.go
ellismg: I'm just worried that some consumer is going to call `Read()` in this state, fail to find an ID token, but since an error is not returned, they go ahead and party on the zero value of the ID Token and expect it to be something sensible.

bgavrilMS:
> This fixes a couple bugs:
> 
> 1. Silent auth defaults to the given account's tenant instead of the client's configured tenant.
> 2. Silent auth fails when the cache lacks an ID token from the requested tenant even though the cache contains a valid refresh token. My fix here is to have the cache return all the matching data it has instead of returning an error on the first miss. The caller is responsible for deciding whether the returned data is sufficient for its purpose.

1. If the app developers configures a tenanted authority (i.e. lmo/GUID), then AcquireTokenSilent should use that GUID. 
2. When does this happen, i.e. when do you not find an IdTOken in the cache?

apps/internal/base/base.go
```diff
@@ -271,11 +271,8 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
 }
 
 func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilentParameters) (AuthResult, error) {
-	tenant := silent.TenantID
-	if tenant == "" {
-		tenant = silent.Account.Realm
-	}
-	authParams, err := b.AuthParams.WithTenant(tenant)
+	// when silent.TenantID == "", this is a no-op and we use the client application's configured tenant
```
bgavrilMS, line 274: Hmm. If the `tenant == ""`, are you going to use the `tenantId` of the PCA/CCA authority? This should only be the case if that authority's tenant is a GUID or `consumers`, but not `common` or `organizations`. If it's `common` or `organiations` then `Account.HomeTenantId` is the correct tenant to use. 

Note: HomeTenantId is not the same as Realm.

chlowell: Is this correct?
```mermaid
graph TD
    A{caller set WithTenantID?}
    A-->|yes|B
    A-->|no|C
    C[check authority]-->D
    D(GUID or consumers)-->B
    C-->E(common or organizations)
    E-->|PCA|F
    E-->|CCA|G{{error: need authority}}
    B[use that tenant]
    F{have home tenant?}
    F-->|yes|B
    F-->|no|H{{error: need account}}
```

chlowell:
> when do you not find an IdTOken in the cache?

Whenever you silently authenticate a user in a new tenant e.g., authenticate a user in tenant A then silently authenticate that user in tenant B. The cache doesn't return an ID token in this scenario because it searches for one whose realm equals the requested tenant ([here](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/apps/internal/base/internal/storage/storage.go#L337); `realm` argument is the requested tenant). I suppose that could be a bug? I left it as-is because it didn't seem obviously wrong--silent auth doesn't need the ID token.

apps/internal/base/internal/storage/storage.go
chlowell: Sure, that's possible in principle, though I think zero values here are obviously invalid. Because this is an internal API, I don't mind trusting callers to continue following the convention of validating cache data before using it (see [`AuthResultFromStorage()`](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/16c062c187ac55067ece02de9fe65f2f7b4033cd/apps/internal/base/base.go#L99)). If that proves unworkable, I would rather decompose `Read()` into `ReadAccessToken()`, `ReadIDToken()`, etc. than ask callers to ignore irrelevant errors.

apps/internal/base/base.go
chlowell: I've updated the tenant selection code for silent auth to follow the above flow except for returning an error when a confidential client is configured for the "common" authority. That's the default for confidential clients and although it's incorrect for token requests, silent auth is nonetheless possible with that configuration should the client also be configured to use a custom token provider, because the client won't send a token request in that case.

bgavrilMS (approved):

apps/internal/base/base.go
bgavrilMS: For OBO, the customer should be using `incommingAssertion.tid` and not `common`
For client_credentials, we should never use `common`, that's only for user tokens.
For AcquireTokenSilent, you should always have an `Account` and use `acc.HomeTenantId` in some cases.

But what about `AcquireTokenByAuthorizationCode`? That's part of the confidential client, but it should allow usage of `common`

chlowell: This PR does doesn't change the auth code flow. A confidential client can still have `lmo/common` as its authority and try to authenticate there.

rayluo: > I've updated the tenant selection code for silent auth to follow the above flow **except for** returning an error when a confidential client is configured for the "common" authority

I would recommend updating the diagram, because it will become the much-easier-to-consume medium in the future when someone is reviewing this change.

Other than that, this PR looks good to me. Approving.

PR #375, opened 2023-01-24
Remove silent auth special case for home tenant aliases

chlowell:
#366 made access token cache reads prefer a user's home tenant over an alias for that tenant ("common", "organizations"). However, it didn't make a corresponding change to cache writes, which use the alias for the realm. The result is an unnecessary token request during silent auth because the cache wrote a token for e.g. "common" and the client searches for one matching the home tenant. 

This PR prevents that unnecessary request by making the cache tenant agnostic according to the rationale here: https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/341#issuecomment-1400936207. When the client requests a token from "common", for example, it will cache that token with realm "common". I first took the other approach, having cache writes follow reads in replacing aliases with the user's home tenant ID, but found making that work might require a separate implementation of `AcquireTokenSilent` for OBO authentication. The "tenant agnostic" approach is much simpler.

### DIFF
--- a/apps/confidential/confidential_test.go
+++ b/apps/confidential/confidential_test.go
@@ -159,6 +159,51 @@ func TestAcquireTokenByCredential(t *testing.T) {
 	}
 }
 
+func TestAcquireTokenOnBehalfOf(t *testing.T) {
+	// this test is an offline version of TestOnBehalfOf in integration_test.go
+	cred, err := NewCredFromSecret("secret")
+	if err != nil {
+		t.Fatal(err)
+	}
+	lmo := "login.microsoftonline.com"
+	assertion := "assertion"
+	mockClient := mock.Client{}
+	// TODO: OBO does instance discovery twice before first token request https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/351
+	mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, "common")))
+	mockClient.AppendResponse(mock.WithBody(mock.GetTenantDiscoveryBody(lmo, "common")))
+	mockClient.AppendResponse(mock.WithBody(mock.GetAccessTokenBody(token, "", "rt", "", 3600)))
+
+	client, err := New("clientID", cred, WithHTTPClient(&mockClient))
+	if err != nil {
+		t.Fatal(err)
+	}
+	tk, err := client.AcquireTokenOnBehalfOf(context.Background(), assertion, tokenScope)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if tk.AccessToken != token {
+		t.Fatalf("wanted %q, got %q", token, tk.AccessToken)
+	}
+	// should return the cached access token
+	tk, err = client.AcquireTokenOnBehalfOf(context.Background(), assertion, tokenScope)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if tk.AccessToken != token {
+		t.Fatalf("wanted %q, got %q", token, tk.AccessToken)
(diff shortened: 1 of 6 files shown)


apps/confidential/confidential_test.go
```diff
@@ -31,6 +31,8 @@ import (
 	"github.com/kylelemons/godebug/pretty"
 )
 
+const localhost = "https://localhost"
```
rayluo, line 34: Nit: It is probably irrelevant to this PR, but since you bother refactor all the preexisting `https://localhost` magic strings anyway, do we consider change it to `http://localhost` i.e. without httpS? Most times, localhost test would not have a cert provisioned.

apps/internal/base/base.go
```diff
@@ -280,26 +280,8 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
 }
 
 func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilentParameters) (AuthResult, error) {
+	// when tenant == "", the caller didn't specify a tenant and WithTenant will use the client's configured tenant
 	tenant := silent.TenantID
-	if tenant == "" {
-		// the caller didn't specify a tenant, so we'll use the client's configured tenant or the given account's home tenant
-		switch tenant = b.AuthParams.AuthorityInfo.Tenant; tenant {
```
rayluo, line 286: The PR looks surprisingly simple. Is it because the MSAL Go was indeed largely written in the tenant-agnostic way previously, so that this new PR merely needs to revert those logic that we introduced in the recent PR #366?

One question here. If caller did not specify a tenant, don't we still need `tenant = b.AuthParams.AuthorityInfo.Tenant`?

chlowell: `tenant` is used only as the argument to `WithTenant()`. When it's `""`, that method sets the tenant for this authentication to the client's default, which is carried by `b.AuthParams`:

https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/8531f217b6e983184cab9bba88b80600ea879d2b/apps/internal/oauth/ops/authority/authority.go#L190-L194

It does appear that MSAL Go was tenant agnostic before #366; I suppose the relevant parts of the implementation were inspired by MSAL Python? The dynamic tenant feature added the first special handling of "common" and "organizations".

rayluo (approved):

apps/internal/base/base.go
```diff
@@ -280,26 +280,8 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
 }
 
 func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilentParameters) (AuthResult, error) {
+	// when tenant == "", the caller didn't specify a tenant and WithTenant will use the client's configured tenant
 	tenant := silent.TenantID
-	if tenant == "" {
```
bgavrilMS, line 284: Sure looks simpler :)

bgavrilMS (approved):
Signing off on this provided that you've tested that MSAL continues to serve tokens from a cache after the upgrade.

apps/internal/base/base.go
chlowell: 😆 it's definitely that. For reference, here's [`AuthParams.WithTenant()`](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/8531f217b6e983184cab9bba88b80600ea879d2b/apps/internal/oauth/ops/authority/authority.go#L183), which sets the tenant for each authentication and uses the client's configured tenant when the caller doesn't specify an override.

PR #378, opened 2023-01-30
Adding ExportReplace to support the Context object [breaking change for 1.0]

element-of-surprise:
This PR adds support for the Context object in caches that are implementing a cache.

We cannot update ExportReplace without breaking implementors, so we are introducing an ExportReplaceCtx as a superset.

Calls that use methods from ExportReplace will now check for the existence of ExportReplaceCtx and use those superset methods instead if available.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -740,12 +740,12 @@ func (cca Client) AcquireTokenOnBehalfOf(ctx context.Context, userAssertion stri
 
 // Account gets the account in the token cache with the specified homeAccountID.
 func (cca Client) AccountCtx(ctx context.Context, homeAccountID string) Account {
-	return cca.base.Account(ctx, homeAccountID)
+	return cca.base.AccountCtx(ctx, homeAccountID)
 }
 
 // RemoveAccountCtx signs the account out and forgets account from token cache.
 func (cca Client) RemoveAccountCtx(ctx context.Context, account Account) error {
-	cca.base.RemoveAccount(ctx, account)
+	cca.base.RemoveAccountCtx(ctx, account)
 	return nil
 }
 
@@ -753,13 +753,13 @@ func (cca Client) RemoveAccountCtx(ctx context.Context, account Account) error {
 //
 // Deprecated: This function is replaced with AccountCtx().
 func (cca Client) Account(homeAccountID string) Account {
-	return cca.base.Account(context.Background(), homeAccountID)
+	return cca.base.AccountCtx(context.Background(), homeAccountID)
 }
 
 // RemoveAccount signs the account out and forgets account from token cache.
 //
 // Deprecated: This function is replaced with AccountCtx().
 func (cca Client) RemoveAccount(account Account) error {
-	cca.base.RemoveAccount(context.Background(), account)
+	cca.base.RemoveAccountCtx(context.Background(), account)
 	return nil
 }
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -50,8 +50,7 @@ duplication.
 .Net People, Take note on X509:
 This uses x509.Certificates and private keys. x509 does not store private keys. .Net
 has some x509.Certificate2 thing that has private keys, but that is just some bullcrap that .Net
-added, it doesn't exist in real life.  Seriously, "x509.Certificate2", bahahahaha.  As such I've
-put a PEM decoder into here.
+added, it doesn't exist in real life. As such I've put a PEM decoder into here.
 */
 
 // TODO(msal): This should have example code for each method on client using Go's example doc framework.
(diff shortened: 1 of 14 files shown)


chlowell:
I added some changes to help get this to the finish line:
- retarget to `dev` branch
- add `Context` parameters and error returns to existing methods--breaking compatibility--instead of adding new methods (closes #379)
- methods accessing the cache fail on cache I/O errors and propagate those errors
    - I changed some functions to use named return params so they can return errors from deferred functions. I'm not fond of that syntax, feel free to suggest better ideas

apps/confidential/confidential.go
```diff
@@ -740,12 +739,11 @@ func (cca Client) AcquireTokenOnBehalfOf(ctx context.Context, userAssertion stri
 }
 
 // Account gets the account in the token cache with the specified homeAccountID.
-func (cca Client) Account(homeAccountID string) Account {
-	return cca.base.Account(homeAccountID)
+func (cca Client) Account(ctx context.Context, homeAccountID string) (Account, error) {
```
bgavrilMS, line 742: Consider naming the id `accountId` instead of `homeAccountId`. The concept of "home" implies tenancy, which is only in AAD. ADFS doesn't have this. And as we expand to other IdPs, they likely won't have this either.

bgavrilMS (approved):

apps/internal/base/base.go
```diff
@@ -279,12 +283,12 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
...
 }
 
-func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilentParameters) (AuthResult, error) {
+func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilentParameters) (ar AuthResult, err error) {
 	// when tenant == "", the caller didn't specify a tenant and WithTenant will use the client's configured tenant
 	tenant := silent.TenantID
 	authParams, err := b.AuthParams.WithTenant(tenant)
 	if err != nil {
-		return AuthResult{}, err
+		return
```
element-of-surprise, line 291: You should still do "return AuthResult{}, err" instead of a naked return.

Naked returns are considered harmful.

apps/internal/base/base.go
```diff
@@ -296,37 +300,48 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+			if err != nil {
+				return
+			}
+			defer func() {
+				err = b.export(ctx, s, suggestedCacheKey, err)
+			}()
 		}
 		storageTokenResponse, err = b.pmanager.Read(ctx, authParams)
 		if err != nil {
-			return AuthResult{}, err
```
element-of-surprise, line 304: same

apps/internal/base/base.go
```diff
@@ -296,37 +300,48 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+				return
+			}
+			defer func() {
+				err = b.export(ctx, s, suggestedCacheKey, err)
+			}()
 		}
 		authParams.AuthorizationType = authority.ATRefreshToken
 		storageTokenResponse, err = b.manager.Read(ctx, authParams, silent.Account)
 		if err != nil {
-			return AuthResult{}, err
```
element-of-surprise, line 315: same

apps/internal/base/base.go
```diff
@@ -296,37 +300,48 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
 		}
 	}
 
 	// ignore cached access tokens when given claims
 	if silent.Claims == "" {
-		result, err := AuthResultFromStorage(storageTokenResponse)
+		ar, err = AuthResultFromStorage(storageTokenResponse)
 		if err == nil {
-			return result, nil
+			return
```
element-of-surprise, line 337: same

apps/internal/base/base.go
```diff
@@ -296,37 +300,48 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
 		}
 	}
 
 	// redeem a cached refresh token, if available
 	if reflect.ValueOf(storageTokenResponse.RefreshToken).IsZero() {
-		return AuthResult{}, errors.New("no token found")
```
element-of-surprise, line 329: same

apps/internal/base/base.go
```diff
@@ -335,10 +350,11 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 
 	token, err := b.Token.Refresh(ctx, silent.RequestType, authParams, cc, storageTokenResponse.RefreshToken)
 	if err != nil {
-		return AuthResult{}, err
```
element-of-surprise, line 338: same

apps/internal/base/base.go
```diff
@@ -335,10 +350,11 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+		return
 	}
 
-	return b.AuthResultFromToken(ctx, authParams, token, true)
```
element-of-surprise, line 341: same

element-of-surprise: And I'm going to stop noting this, but here and the rest of the file.

apps/internal/base/base.go
```diff
@@ -279,12 +283,12 @@ func (b Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, s
 	return baseURL.String(), nil
 }
 
-func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilentParameters) (AuthResult, error) {
+func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilentParameters) (ar AuthResult, err error) {
```
element-of-surprise, line 286: I'm seeing a lot of places where we added named returns, but we don't need them for a defer statement. 

The original Go team frowned on any use of named returns if defer wasn't involved.  Pike said they should never have made them.  Naked returns, where you say `return` on named returns was verboten inside Google, as it was considered harmful.  There are Go2 requests to remove them, though we all know there will never be a Go2.

I've seen a use or two of named returns outside this that were ok, but I struggle to remember what they were.

element-of-surprise (commented):
As discussed, clothe the returns and ignore my comment on the named returns, I apparently was too tired to be doing reviews, I now see the defer with assignments.

This is approved, however since I created this thing and you are finishing it, I can't approve it.

Issue #28, opened 2020-08-12
Review logging


hchittanuru3:
Make sure to review what is being logged

bgavrilMS:
I have some guidance around logging here: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/7325?_a=files&path=/PII/pii%20handling.md

I'm happy to review this one.

chlowell:
Closed by #382

PR #382, opened 2023-02-06
Logging audit and logging docs

bgavrilMS:
Fix for https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/28

1. Review all logs to ensure there is no PII (the library would need to expose a PII flag to be able to log PII)
2. Provide some clarity in the docs about what PII is and what the library logs.

### DIFF
--- a/apps/design/design.md
+++ b/apps/design/design.md
@@ -5,6 +5,7 @@ Contributors:
 - Keegan Caruso(Keegan.Caruso@microsoft.com)
 - Joel Hendrix(jhendrix@microsoft.com)
 - Santiago Gonzalez(Santiago.Gonzalez@microsoft.com)
+- Bogdan Gavril (bogavril@microsoft.com)
 
 ## History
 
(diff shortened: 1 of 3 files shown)


apps/design/design.md
```diff
@@ -140,6 +141,10 @@ Since we have representation from the Go SDK team, we might have them go bridge
 the current implementation using some of that code so its possible for our users to store the
 cert in Keyvault.
 
-## Notes:
+## Logging
 
-Do we need: AcquireTokenSilent()??  Seems like we could just bake this into other acquire calls automatically????
\ No newline at end of file
+All logging is done using the `fmt` package. For errors, see [error design](../errors/error_design.md).
```
chlowell, line 146: "Logging by `fmt`" implies printing to stdout, which clients don't do.
```suggestion
For errors, see [error design](../errors/error_design.md).
```
Actually, clients communicate by return values instead of logging, so although I like the rest of this content, it's a little awkward. However, it's good information and vacuously true today, and it would be concretely true of any future logging, which wouldn't include PII.

chlowell (approved):

apps/design/design.md
```diff
@@ -140,6 +141,10 @@ Since we have representation from the Go SDK team, we might have them go bridge
...
 cert in Keyvault.
 
-## Notes:
+## Logging
 
-Do we need: AcquireTokenSilent()??  Seems like we could just bake this into other acquire calls automatically????
\ No newline at end of file
+For errors, see [error design](../errors/error_design.md).
+
+This library does not log personal identifiable information (PII). For a definition of PII, see https://www.microsoft.com/en-us/trust-center/privacy/customer-data-definitions. MSAL Go does not log any of the 3 data categories listed there.
```
bgavrilMS, line 148: But there are `fmt` calls all throughout the library, no? App developers can redirect stdout to their logs. For all intents and purposes, the `fmt` calls are like logging.

chlowell: There are, however not all `fmt` functions write to stdout. MSAL clients call `fmt.Errorf` and `fmt.Sprintf`, which format strings without printing them.

Issue #380, opened 2023-02-03
API changes for GA release


chlowell:
# API changes since v0.8.0

Upgrading from v0.8.* may require code changes as described below.

## `confidential.New` requires an authority [#394](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/394)

`confidential.WithAuthority` has been deleted, and the URL it set is now a parameter of `confidential.New` :

```go
// before
client, err := confidential.New(clientID, cred, confidential.WithAuthority(authority))

// after
client, err := confidential.New(authority, clientID, cred)
```

## `confidential.NewCredFromCertChain` replaces `NewCredFromCert` [#391](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/391)

If your application calls `NewCredFromCert`, update arguments and return values:

### before

```go
certs, key, err := confidential.CertFromPEM(certData, password)
if err != nil {
    // TODO: handle error
}
cred := confidential.NewCredFromCert(certs[0], key)
```

### after

```go
certs, key, err := confidential.CertFromPEM(certData, password)
if err != nil {
    // TODO: handle error
}
cred, err := confidential.NewCredFromCert(certs, key)
if err != nil {
    // TODO: handle error
}
```

If your application calls `confidential.NewCredFromCertChain`, call `NewCredFromCert` instead.

## All methods accessing cache data have a `Context` parameter [#378](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/378)

For example, `RemoveAccount` and `Accounts`:

```go
// before
accounts, err := client.Accounts()

// after
accounts, err := client.Accounts(context.TODO())
```

## Renamed functions

These functions are renamed but otherwise unchanged:

- `confidential.NewCredFromCertChain` -> `NewCredFromCert` [#391](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/391)
- `confidential.WithAccessor` -> `WithCache` [#386](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/386)
- `public/Client.CreateAuthCodeURL` -> `AuthCodeURL` [#384](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/384)

## Deleted API

- `confidential.NewCredFromAssertion`. This function was deprecated in v0.6.0; use `confidential.NewCredFromAssertionCallback` instead [#385](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/385)
[...]

PR #385, opened 2023-02-14
Delete deprecated confidential.NewCredFromAssertion

chlowell:
Part of #380

### DIFF
--- a/apps/internal/version/version.go
+++ b/apps/internal/version/version.go
@@ -5,4 +5,4 @@
 package version
 
 // Version is the version of this client package that is communicated to the server.
-const Version = "0.8.1"
+const Version = "0.9.0"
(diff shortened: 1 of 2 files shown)


apps/internal/version/version.go
```diff
@@ -5,4 +5,4 @@
 package version
 
 // Version is the version of this client package that is communicated to the server.
-const Version = "0.8.1"
+const Version = "0.9.0"
```
bgavrilMS, line 8: I assume `0.9.0` will be the "beta" release where we aim to stabilize the public API?

chlowell: Yes, that's the idea, basically an RC. `0.9.0` would be the last `0.x` and would have the stable API.

PR #386, opened 2023-02-14
Rename confidential.WithAccessor to WithCache

chlowell:
Part of #380. Confidential client has `WithAccessor`, public client has `WithCache`, but both should have the same name because they do same thing. `WithCache` looks better to me because it indicates more clearly what the option does.


bgavrilMS (approved):

chlowell:
SonarCloud is red because aligning the option name and documentation for both clients created too much duplication in this PR. I merged anyway because that duplication is by design and not a quality problem.

Issue #32, opened 2020-10-05
Add support for B2C authorities


sangonzal:

siredmar:
I just spent two days trying to understand what i am doing wrong. Now i searched the library for any reference of `b2c` or `B2C`. Seems that nothing is in there or commented out. 
Please add support for B2C!

Roronoa-Yu:
I have same scenarios, please add support for azure b2c.
<img width="1417" alt="image" src="https://user-images.githubusercontent.com/27270110/120584836-426a4800-c463-11eb-836c-542ac2b86e4b.png">

satarr:
Same here. This library seems to have been left out in the cold. No updates for B2C and "work in progress" code is present within package source code.

aeastw:
Hi, just wondering if there are any plans to schedule support for B2C?

bgavrilMS:
There are no plans to support B2C in MSAL GO.

PR #388, opened 2023-02-16
Quick fix for Azure AD B2C

szkiba:
Fix CanonicalAuthorityURI handling and add default Azure AD B2C domain (.b2clogin.com) as trusted authority host domain.
Theese two changes makes the library works with Azure AD B2C (and user flows).

Maybe not this is the best way to fix the problem, but other solutions would require significant refactor in code.

### DIFF
--- a/apps/internal/oauth/ops/authority/authority.go
+++ b/apps/internal/oauth/ops/authority/authority.go
@@ -51,6 +51,12 @@ func TrustedHost(host string) bool {
 	if _, ok := aadTrustedHostList[host]; ok {
 		return true
 	}
+
+	// default Azure AD B2C authority domain
+	if strings.HasSuffix(host, ".b2clogin.com") {
+		return true
+	}
+
 	return false
 }
 
--- a/apps/internal/oauth/ops/authority/authority.go
+++ b/apps/internal/oauth/ops/authority/authority.go
@@ -309,13 +309,17 @@ type Info struct {
 	InstanceDiscoveryDisabled bool
 }
 
-func firstPathSegment(u *url.URL) (string, error) {
+func firstTwoPathSegment(u *url.URL) (string, string, error) {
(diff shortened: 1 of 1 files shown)


szkiba:

mrueg:
@chlowell any chance to get this or a similar fix in? Thanks!

chlowell:
B2C isn't supported yet (see #32), so I imagine we need more than a quick fix. However, I don't know what all is required to support B2C, so I'm not equipped to evaluate this PR from that perspective. Setting that aside, we'd certainly need tests before taking a change like this. I'm afraid I don't have much to contribute there either because I also don't know what testing is required for B2C. FYI @bgavrilMS

mrueg:
It looks like the "about" description part of the repository is misleading and could be improved:

```
It uses industry standard OAuth2 and OpenID Connect. The library also supports Azure AD B2C.
```

bgavrilMS:
Yeah, we'd also need to disable instance discovery similar to ADFS authorities, probably have an authority_type property that can be used elsewhere and have tests. 

I updated the "About".

bgavrilMS:
I am going to reject this. We will not be adding B2C support to MSAL GO.

mrueg:
Would you mind documenting that as a limitation of this library? 

From an external point of view, I would assume the target is to have feature parity on a framework level for all supported languages.
https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/104
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/8c1c63df4be8e057eb3e08e0ce086570b0fd9ac0/lib/msal-browser/docs/working-with-b2c.md

bgavrilMS:
Yes, but there is a new offering called CIAM that superseeds B2C. While MSAL GO doesn't support CIAM, we would want new apps to target B2C. 

https://github.com/Azure-Samples/ms-identity-ciam-dotnet-tutorial

PR #390, opened 2023-02-16
Remove unnecessary options structs

chlowell:
Part of #380. This removes the options bags for methods and constructors from the public API. These types are useless in application code because they aren't input to any function (applications set options by calling functions like `WithAuthority()` instead). They could conceivably be used in unusual test scenarios as part of a fake MSAL however, because these types distract from the real options API, I think we shouldn't export them before seeing a clear need for them; doing so would be non-breaking.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -243,74 +243,78 @@ func AutoDetectRegion() string {
 // package doc. A new Client should be created PER SERVICE USER.
 // For more information, visit https://docs.microsoft.com/azure/active-directory/develop/msal-client-applications
 type Client struct {
-	accessor                          cache.ExportReplace
-	authority, azureRegion            string
-	base                              base.Client
-	capabilities                      []string
-	cred                              *accesstokens.Credential
-	disableInstanceDiscovery, sendX5C bool
-	httpClient                        ops.HTTPClient
+	base base.Client
+	cred *accesstokens.Credential
 	// userID is some unique identifier for a user. It actually isn't used by us at all, it
 	// simply acts as another hint that a confidential.Client is for a single user.
 	userID string
 }
 
-func (c Client) validate() error {
-	u, err := url.Parse(c.authority)
+// clientOptions are optional settings for New(). These options are set using various functions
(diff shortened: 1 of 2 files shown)


apps/confidential/confidential.go
```diff
@@ -243,97 +243,74 @@ func AutoDetectRegion() string {
 // package doc. A new Client should be created PER SERVICE USER.
 // For more information, visit https://docs.microsoft.com/azure/active-directory/develop/msal-client-applications
 type Client struct {
-	base base.Client
-
-	cred *accesstokens.Credential
-
+	accessor                          cache.ExportReplace
```
bgavrilMS, line 246: should we call it more in the lines of `cache serialization` ? Maybe `cacheSerializer` ?

bgavrilMS: (not sure if this is part of public API. If not, ignore comment)

apps/public/public.go
```diff
@@ -47,101 +48,87 @@ type AuthResult = base.AuthResult
...
+	return func(c *Client) {
+		c.disableInstanceDiscovery = !enabled
 	}
 }
 
 // Client is a representation of authentication client for public applications as defined in the
 // package doc. For more information, visit https://docs.microsoft.com/azure/active-directory/develop/msal-client-applications.
 type Client struct {
-	base base.Client
+	accessor                 cache.ExportReplace
```
bgavrilMS, line 105: most of these are in common with confidential client.

bgavrilMS (approved):

apps/confidential/confidential.go
chlowell: Just FYI, names beginning with a lowercase letter are private

jhendrixMSFT (approved):

apps/public/public.go
chlowell: There are ways we could share these fields but I haven't thought of one that's really better than duplicating them. We could for example move the common fields onto a new type and have constructor options set them there. However, accommodating options not shared between client types would then be difficult and more complex than duplicating the common fields.

PR #392, opened 2023-02-24
Upgrade CI tools

chlowell:
Upgrading CI to use the latest version of Go and linter. The upgraded linter flagged a method call that has no side effect and whose return value is discarded i.e., the call has no effect. I deleted this call because it's unnecessary in any case--there's no reason to update `DeviceCodeResult.ExpiresOn` because it's the moment at which the code expires, not the amount of time remaining until the code expires.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        go: ["1.19", "1.18"]
+        go: ["1.19", "1.20"]
 
     steps:
       - name: Set up Go 1.x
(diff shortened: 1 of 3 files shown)


rayluo:
> a method call that has no side effect and whose return value is discarded i.e., the call has no effect. I deleted this call because it's unnecessary in any case--there's no reason to update `DeviceCodeResult.ExpiresOn` because it's the moment at which the code expires, not the amount of time remaining until the code expires.

Makes sense. Agree with its removal. FWIW, [that line was introduced in a 5000-line change PR](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/117/files#diff-b320ef63c36a1f830fd3413a42e88f1f33b0ebf225facc9024fbdc671851a27bR162). Perhaps it was meant to be a hack to simulate the expiration for testing.

rayluo (approved):

PR #397, opened 2023-03-14
fix: public/browserLogin add callback path support

RomanManz:
The current version only works with callbacks not containing URL paths.
Also added a sample using the public AcquireTokenInteractive flow.


RomanManz:

bgavrilMS:
We will not accept this.

Issue #400, opened 2023-03-27
[Feature Request] Gain optional claims from IDToken


jkroepke:
**Is your feature request related to a problem? Please describe.**
I have configured optional claims on my app registration. After calling `result, err := devCode.AuthenticationResult(ctx)`, the claim is part of the IDToken, but not accessible through the go struct.

**Describe the solution you'd like**
Make all optional claim available as struct Field. 

**Describe alternatives you've considered**
Use the [`AdditionalFields`](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/9a63cc219c3822f23f3a54c62c1261b729a50536/apps/internal/oauth/ops/accesstokens/tokens.go#L42) field for non-defined fields.

bgavrilMS:
Why not use the `AdditionalFields` ? I don't think we should be adding explicit support for a subset of optional claims.

Besides, I believe AAD already has Conditional Access policies to prevent or force MFA based on IP address..

PR #401, opened 2023-03-27
Added optional claim ipaddr to IDToken struct

jkroepke:
See #400 for reference. 

Since the optional claim `UPN` is already present, I thought about adding `ipaddr` as well here.

### DIFF
--- a/apps/internal/oauth/ops/accesstokens/tokens.go
+++ b/apps/internal/oauth/ops/accesstokens/tokens.go
@@ -30,6 +30,7 @@ type IDToken struct {
 	TenantID          string `json:"tid,omitempty"`
 	Subject           string `json:"sub,omitempty"`
 	UPN               string `json:"upn,omitempty"`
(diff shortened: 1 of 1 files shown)


jkroepke:
@bgavrilMS @rayluo Is there any additional process that I have to follow that this is getting merged?

apps/internal/oauth/ops/accesstokens/tokens.go
```diff
@@ -30,6 +30,7 @@ type IDToken struct {
 	TenantID          string `json:"tid,omitempty"`
 	Subject           string `json:"sub,omitempty"`
 	UPN               string `json:"upn,omitempty"`
+	IpAddress         string `json:"ipaddr,omitempty"`
```
chlowell, line 33:
```suggestion
	IPAddress         string `json:"ipaddr,omitempty"`
```
(following the Google [style guide](https://google.github.io/styleguide/go/decisions#initialisms))

PR #402, opened 2023-04-07
Prevent persistent cache data races

chlowell:
Concurrent authentication causes data loss when using a persistent cache because `base.Client` doesn't synchronize its access. The data flow of authentication looks like this:
1. load persistent cache into memory (clobber data in memory)
1. authenticate
1. store authentication data in memory
1. write in-memory data to persistent cache (clobber persisted data)

When multiple goroutines authenticate concurrently, one may execute step 1 while another is between steps 3 and 4, deleting new data in memory before it's persisted. My solution in this PR is simply to lock around accessing the persistent cache.

To prevent repetition, I simplified several methods by consolidating the (internal) interfaces for partitioned and flat caches. These interfaces had almost the same Read/Write API; the difference was an unnecessary parameter. While I was at it, I also eliminated unnecessary writes to persistence--methods that don't write the in-memory cache no longer export it to the persistent cache.

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -43,15 +43,6 @@ type accountManager interface {
 	RemoveAccount(account shared.Account, clientID string)
 }
 
-type noopCacheAccessor struct{}
-
-func (n noopCacheAccessor) Replace(ctx context.Context, u cache.Unmarshaler, h cache.ReplaceHints) error {
-	return nil
-}
-func (n noopCacheAccessor) Export(ctx context.Context, m cache.Marshaler, h cache.ExportHints) error {
-	return nil
-}
-
 // AcquireTokenSilentParameters contains the parameters to acquire a token silently (from cache).
 type AcquireTokenSilentParameters struct {
 	Scopes            []string
@@ -218,7 +209,6 @@ func New(clientID string, authorityURI string, token *oauth.Client, options ...O
 	client := Client{ // Note: Hey, don't even THINK about making Base into *Base. See "design notes" in public.go and confidential.go
 		Token:           token,
 		AuthParams:      authParams,
-		cacheAccessor:   noopCacheAccessor{},
 		cacheAccessorMu: &sync.RWMutex{},
 		manager:         storage.New(token),
 		pmanager:        storage.NewPartitionedManager(token),
@@ -305,10 +295,12 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 		authParams.AuthorizationType = authority.ATRefreshToken
 		m = b.manager
 	}
-	key := authParams.CacheKey(silent.IsAppCache)
-	b.cacheAccessorMu.RLock()
-	err = b.cacheAccessor.Replace(ctx, m, cache.ReplaceHints{PartitionKey: key})
-	b.cacheAccessorMu.RUnlock()
+	if b.cacheAccessor != nil {
+		key := authParams.CacheKey(silent.IsAppCache)
+		b.cacheAccessorMu.RLock()
+		err = b.cacheAccessor.Replace(ctx, m, cache.ReplaceHints{PartitionKey: key})
+		b.cacheAccessorMu.RUnlock()
+	}
 	if err != nil {
 		return ar, err
 	}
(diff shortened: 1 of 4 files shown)


apps/internal/base/base.go
```diff
@@ -296,38 +300,21 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+	if authParams.AuthorizationType != authority.ATOnBehalfOf {
 		authParams.AuthorizationType = authority.ATRefreshToken
-		storageTokenResponse, err = b.manager.Read(ctx, authParams, silent.Account)
-		if err != nil {
-			return ar, err
-		}
+		m = b.manager
+	}
+	key := authParams.CacheKey(silent.IsAppCache)
+	b.mu.RLock()
```
jhendrixMSFT, line 309: Shouldn't this be the write lock?

apps/internal/base/base.go
```diff
@@ -219,6 +221,7 @@ func New(clientID string, authorityURI string, token *oauth.Client, options ...O
 		cacheAccessor: noopCacheAccessor{},
 		manager:       storage.New(token),
 		pmanager:      storage.NewPartitionedManager(token),
+		mu:            &sync.RWMutex{},
```
jhendrixMSFT, line 224: Is the generic name `mu` to imply that this must be held when reading/writing any field?

apps/internal/base/base.go
```diff
@@ -417,103 +400,66 @@ func (b Client) AcquireTokenOnBehalfOf(ctx context.Context, onBehalfOfParams Acq
...
 	authParams.AuthorizationType = authority.AccountByID
 	authParams.HomeAccountID = homeAccountID
-	if s, ok := b.manager.(cache.Serializer); ok {
-		suggestedCacheKey := b.AuthParams.CacheKey(false)
-		err = b.cacheAccessor.Replace(ctx, s, cache.ReplaceHints{PartitionKey: suggestedCacheKey})
-		if err != nil {
-			return acct, err
-		}
-		defer func() {
-			err = b.export(ctx, s, suggestedCacheKey, err)
```
jhendrixMSFT, line 488: Does this call to `export()` need to be preserved?

chlowell: No; export writes the in-memory cache to the persistence implementation. This method copies persisted data into the in-memory cache and searches it for an account. It doesn't create any data, so it has nothing to add to the persisted data.

apps/internal/base/base.go
chlowell: No, this is only for locking around `cacheAccessor`. Would something like `cacheMu` be better?

jhendrixMSFT: In that case, the pattern is to name them `<var that's locked>Mu` so `cacheAccessorMu`

apps/internal/base/base.go
chlowell: It doesn't need to be, or at least I haven't come up with a scenario requiring an exclusive lock here :thinking:. Here's my reasoning: `Replace` unmarshals persisted data into the in-memory cache, so calling it repeatedly can cause the in-memory cache to lose only data that's been removed from the persistent cache. That could happen and may be a bug, but either way it's outside the responsibility of this type, which just needs to prevent its methods from concurrently calling `Replace` and `Export`.

I'd like to read lock here, provided that's safe, because this is a hot path. I haven't benchmarked this but I expect a significant perf advantage for clients without a persisted cache (i.e. almost all of them) because `Replace` is a no-op for them.

bgavrilMS:
Can you describe the actual user scenario that got broken? i.e. how does this race manifest? 

A few thoughts:

## MSAL.NET chose a slightly more granular approach to locking

1. LOCK. load persistent cache into memory (clobber data in memory). RELEASE LOCK
2. authenticate
3. LOCK. re-load persistent cache into memory 
4. store authentication data in memory
5. write in-memory data to persistent cache (clobber persisted data). RELEASSE LOCK

This ensures that we don't lock for the entire duration of the HTTP request. 

## Eventual consistency vs race issues

In MSAL.NET we experimented with disabling that lock. Unfortunately this seems to lead to a situation where a cache node (e.g. for Redis) now contains too many tokens. @pmaytak is looking at this.

I think the fix looks fine, but would be good to have a rough idea of the perf impact.

chlowell:
The race I described above causes unnecessary reauthentication because some tokens don't make it to the persistent cache and are lost the next time the client loads persisted data. Reauthentication could be acceptable if it were rare, but empirically I see data loss ~25% of the time with only 2 goroutines authenticating concurrently.

## benchmarks

The scenario is 10 goroutines concurrently authenticating and retrieving a cached token, with the cache persisted to a file. "Baseline" is the latest published version i.e. none of the locking added in this PR. "Read lock" is this PR as it is now, with a read lock in `AcquireTokenSilent`. "Exclusive lock" changes that read lock to an exclusive one.

image

It's interesting that this PR actually improves the benchmark despite adding locks. I believe this is because it also removes unnecessary writes; apparently those are more expensive in this scenario than the new locks.

bgavrilMS (approved):

bgavrilMS:
> 

@pmaytak - can we do something similar for MSAL.NET?

bgavrilMS:
@chlowell - 30% perf improvement for cached token acquisition is massive! Ship it :)

apps/internal/base/base.go
chlowell: The persisted cache can lock to prevent simultaneous Export and Replace but only for the duration of each call. That approach can't prevent the destructive interleaved calls I described above. Replace->authenticate->Export is a transaction only at a higher level.

There's another way to eliminate the locking overhead for clients without a persistent cache. Currently, the client has a default "persistent" cache with no-op Export/Replace and always locks around them. If the default were `nil` instead, the client could decide to lock only when it's not `nil` i.e. the user specified a persistent cache.

jhendrixMSFT (approved):

rayluo:
> It's interesting that this PR actually improves the benchmark despite adding locks. I believe this is because it also removes unnecessary writes; apparently those are more expensive in this scenario than the new locks.

Given that this PR improves the baseline, that hypothesis is reasonable, and I am also giving a Ship-It! :-)

The following comments are mostly FYIs for open discussion.

> > solution in this PR is simply to lock around accessing the persistent cache
>
> MSAL.NET chose a slightly more granular approach to locking
> ......
> This ensures that we don't lock for the entire duration of the HTTP request.

The MSAL .Net's approach were also adopted in MSAL Python and MSAL Java. The MSAL Go implementation in this PR looks also in the same granular manner, as opposed to locking up the entire token acquisition http request. So, we are all good here.

> "Read lock" is this PR as it is now, with a read lock in AcquireTokenSilent. "Exclusive lock" changes that read lock to an exclusive one.

Were the "read lock" and "exclusive lock" referred to the actual file lock characteristic? FWIW, MSAL Python (EX) implementation used to acquire a lock exclusively, during both token read and token write. Since then, an improvement kept only lock behavior during token write, and removed the lock attempt on token read. It just assumes a successful read (meaning the token cache file is in valid json format) would mean the data is clean. It also seems to work well.

> The race I described above causes unnecessary reauthentication because some tokens don't make it to the persistent cache and are lost the next time the client loads persisted data. Reauthentication could be acceptable if it were rare, but empirically I see data loss ~25% of the time with only 2 goroutines authenticating concurrently.

Out of curiosity, was that "empirical ~25% data loss with only 2 goroutines authenticating concurrently" a real-world "end user occasionally runs 2 instances of same app manually" scenario, or an in-lab "spinning up 2 instances and having them loop authentication repeatedly" scenario? The "~25% data loss" sounds too-high-than-we-thought for the former.

rayluo (approved):

PR #404, opened 2023-04-11
Add `state` to `authCodeURLOptions`

maxtwardowski:
The purpose of this PR is to make it possible to pass `state` when requesting an auth code URL


maxtwardowski:
`@microsoft-github-policy-service agree company="pyr.sh"`

PR #405, opened 2023-04-11
fix: empty cache key in client.Account

joelrose:
The `CacheKey` method in `Account` is called on the wrong receiver. The `AuthParams` struct is correctly copied at the beginning of the method but then not used to call the `CacheKey` method.

This leads to `CacheKey` always returning an empty string.


bgavrilMS:
We would require a unit test for this fix.

joelrose:
Hi @bgavrilMS, closing this as I've chosen an approach avoiding this library and won't be able to update the PR. I'm also not sure if this issue still persists, but it looks like the changed code still exists which would indicate to me that this is still a problem.

PR #408, opened 2023-04-12
Remove r.login.microsoftonline.com

element-of-surprise:
Issue detailed in #361

PR #403, opened 2023-04-10
Adding some better diagnostics for len(scopes) == 0

element-of-surprise:
For issue:

https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/304

Issue #361, opened 2022-11-14
[Feature Request] Switch regional on public client to use <region>.login.microsoft.com instead of <region>.r.login.microsoftonline.com


bgavrilMS:
See https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2165936 for details.

element-of-surprise:
Fixed with: https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/408

Issue #304, opened 2022-04-13
Return a friendlier error when no scopes are specified


chlowell:
Token acquisition methods require a `[]string` of scopes but don't require it to contain values. If it's empty, the method will request a token for default scopes "openid offline_access profile" and return an error from AAD like
```
{error":"invalid_scope","error_description":"AADSTS1002012: The provided value for scope
openid offline_access profile is not valid. Client credential flows must have a scope value 
with /.default suffixed to the resource identifier (application ID URI)...}
```
Is there a scenario in which requesting these default scopes gets a useful token? If not, please consider returning a friendlier error to callers who don't specify a scope.

element-of-surprise:
https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/403

element-of-surprise:
Change merged, closing this out.

PR #412, opened 2023-04-19
MSAL Go 1.0.0

rayluo:
The following content will become the release notes for this release. Please proofread or suggest some blurb.

* Better diagnostics message (#304, #403)
* Prevent persistent cache data races (#402)
* Switch to new regional endpoints (#361, #408)

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -10,6 +10,7 @@ import (
 	"net/url"
 	"reflect"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/cache"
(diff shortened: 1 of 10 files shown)


chlowell (approved):

apps/internal/base/base.go
```diff
@@ -417,103 +392,76 @@ func (b Client) AcquireTokenOnBehalfOf(ctx context.Context, onBehalfOfParams Acq
...
-		err = b.cacheAccessor.Replace(ctx, s, cache.ReplaceHints{PartitionKey: suggestedCacheKey})
+func (b Client) AllAccounts(ctx context.Context) ([]shared.Account, error) {
+	if b.cacheAccessor != nil {
+		b.cacheAccessorMu.RLock()
+		defer b.cacheAccessorMu.RUnlock()
+		key := b.AuthParams.CacheKey(false)
+		err := b.cacheAccessor.Replace(ctx, b.manager, cache.ReplaceHints{PartitionKey: key})
 		if err != nil {
-			return accts, err
+			return nil, err
```
RickWinter, line 430: Should this be an empty array instead of nil?  Similar to how Account returns a default Acount{}

apps/internal/oauth/oauth.go
```diff
@@ -107,6 +112,9 @@ func (t *Client) Credential(ctx context.Context, authParams authority.AuthParams
 		}
 		tr, err := cred.TokenProvider(ctx, params)
 		if err != nil {
+			if len(scopes) == 0 {
+				return accesstokens.TokenResponse{}, fmt.Errorf("token request had an empty authority.AuthParams.Scopes, which may cause the following error: %w", err)
```
RickWinter, line 116: Should we fmt the err before returning an error that uses the format?
  see line 207 below

apps/internal/oauth/oauth.go
```diff
@@ -178,15 +204,24 @@ func (t *Client) UsernamePassword(ctx context.Context, authParams authority.Auth
...
 		}
-		return t.AccessTokens.FromSamlGrant(ctx, authParams, saml)
+		tr, err := t.AccessTokens.FromSamlGrant(ctx, authParams, saml)
+		if err != nil {
+			return accesstokens.TokenResponse{}, err
+		}
+		return tr, nil
 	case authority.Managed:
+		if len(authParams.Scopes) == 0 {
+			return accesstokens.TokenResponse{}, fmt.Errorf("token request had an empty authority.AuthParams.Scopes, which may cause the following error: %w", err)
```
RickWinter, line 223: Ditto on comment about formating the error in a seperate line

apps/internal/base/base.go
chlowell: Callers can iterate over a `nil` slice, check its length, etc. so we don't break calling code by returning `nil` here. However, in this case callers should ignore the slice because the function returns a non-nil error.

apps/internal/oauth/oauth.go
chlowell: You mean, to shorten this line?

localden (approved):

apps/internal/oauth/oauth.go
RickWinter: Consistency and coding patterns was the feedback.  We inline the format in some cases and create a temp to immediately return in others.  No difference in the end result but it does at a glance imply something is different

rayluo:
@chlowell , you added the `test` label here. Are you running some test with this release candidate? I will wait for your GO, then. :-)

bgavrilMS (approved):

chlowell:
I added the label to trigger the Action that runs tests (any label would have worked). I did also run the azidentity test suite--need an update there for the regional authority change but nothing is really broken. I think we're ready to go 🚀

PR #419, opened 2023-05-11
Update readme

chlowell:
Removing preview references and links to outdated wiki and incomplete samples; updating code for v1.0. The readme is part of the package documentation, so this is worth publishing as a patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,33 +56,26 @@ Acquiring tokens with MSAL Go follows this general pattern. There might be some 
     * Public clients should specify a user account, if one is available:
 
     ```go
-    accounts, err := publicClient.Accounts(context.TODO())
+    // If your application previously authenticated a user, call AcquireTokenSilent with that user's account
+    // to use cached authentication data. This example shows choosing an account from the cache, however this
+    // isn't always necessary because the AuthResult returned by authentication methods includes user account
+    // information.
+    accounts, err := client.Accounts(context.TODO())
     if err != nil {
         // TODO: handle error
     }
-    var userAccount public.Account
     if len(accounts) > 0 {
-        // there may be more accounts; here we assume the first one is wanted
-        userAccount = accounts[0]
+        // There may be more accounts; here we assume the first one is wanted.
+        // AcquireTokenSilent returns a non-nil error when it can't provide a token.
+        result, err = client.AcquireTokenSilent(context.TODO(), scopes, public.WithSilentAccount(accounts[0]))
     }
-
-    scopes := []string{"scope"}
-    var result public.AuthResult
-    if !userAccount.IsZero() {
-        // search the cache for an access token
-        result, err = publicClient.AcquireTokenSilent(context.TODO(), scopes, public.WithSilentAccount(userAccount))
-    }
-    if err != nil {
-        // cache miss, authenticate a user with another AcquireToken* method
-        result, err = publicClient.AcquireTokenInteractive(context.TODO(), scopes)
+    if err != nil || len(accounts) == 0 {
+            // cache miss, authenticate a user with another AcquireToken* method
+        result, err = client.AcquireTokenInteractive(context.TODO(), scopes)
         if err != nil {
             // TODO: handle error
         }
-        // this account can be used in a future AcquireTokenSilent call
-        userAccount = result.Account
     }
-	// TODO: use access token
-    _ = result.AccessToken
     ```
 
     * Confidential clients can simply call `AcquireTokenSilent()`:
--- a/README.md
+++ b/README.md
@@ -56,27 +56,33 @@ Acquiring tokens with MSAL Go follows this general pattern. There might be some 
     * Public clients should specify a user account, if one is available:
 
     ```go
-    var userAccount public.Account
     accounts, err := publicClient.Accounts(context.TODO())
-	if err != nil {
-		// TODO: handle error
-	}
+    if err != nil {
+        // TODO: handle error
+    }
+    var userAccount public.Account
     if len(accounts) > 0 {
         // there may be more accounts; here we assume the first one is wanted
         userAccount = accounts[0]
     }
+
     scopes := []string{"scope"}
-    result, err := publicClient.AcquireTokenSilent(context.TODO(), scopes, public.WithSilentAccount(userAccount))
+    var result public.AuthResult
+    if !userAccount.IsZero() {
+        // search the cache for an access token
+        result, err = publicClient.AcquireTokenSilent(context.TODO(), scopes, public.WithSilentAccount(userAccount))
+    }
     if err != nil {
-        // cache miss, authenticate a user with another AcquireToken... method
+        // cache miss, authenticate a user with another AcquireToken* method
         result, err = publicClient.AcquireTokenInteractive(context.TODO(), scopes)
         if err != nil {
             // TODO: handle error
         }
+        // this account can be used in a future AcquireTokenSilent call
+        userAccount = result.Account
     }
-    if err == nil {
-        accessToken := result.AccessToken
-    }
+	// TODO: use access token
+    _ = result.AccessToken
     ```
 
     * Confidential clients can simply call `AcquireTokenSilent()`:
--- a/README.md
+++ b/README.md
@@ -70,12 +70,15 @@ Acquiring tokens with MSAL Go follows this general pattern. There might be some 
         result, err = client.AcquireTokenSilent(context.TODO(), scopes, public.WithSilentAccount(accounts[0]))
     }
     if err != nil || len(accounts) == 0 {
-            // cache miss, authenticate a user with another AcquireToken* method
+        // cache miss, authenticate a user with another AcquireToken* method
         result, err = client.AcquireTokenInteractive(context.TODO(), scopes)
         if err != nil {
             // TODO: handle error
         }
     }
+    // TODO: save the authenticated user's account, use the access token
+    userAccount := result.Account
+    accessToken := result.AccessToken
     ```
 
     * Confidential clients can simply call `AcquireTokenSilent()`:
@@ -90,9 +93,7 @@ Acquiring tokens with MSAL Go follows this general pattern. There might be some 
             // TODO: handle error
         }
     }
-    if err == nil {
-        accessToken := result.AccessToken
-    }
+    accessToken := result.AccessToken
     ```
 
 ## Community Help and Support
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Acquiring tokens with MSAL Go follows this general pattern. There might be some 
     ```go
     import "github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
 
-    publicClient, err := public.New("client_id", public.WithAuthority("https://login.microsoft.com/your_tenant"))
+    publicClient, err := public.New("client_id", public.WithAuthority("https://login.microsoftonline.com/your_tenant"))
     ```
 
    * Initializing a confidential client:
@@ -48,7 +48,7 @@ Acquiring tokens with MSAL Go follows this general pattern. There might be some 
     if err != nil {
         // TODO: handle error
     }
-    confidentialClient, err := confidential.New("https://login.microsoft.com/your_tenant", "client_id", cred)
+    confidentialClient, err := confidential.New("https://login.microsoftonline.com/your_tenant", "client_id", cred)
     ```
 
 1. Call `AcquireTokenSilent()` to look for a cached token. If `AcquireTokenSilent()` returns an error, call another `AcquireToken...` method to authenticate.
--- a/README.md
+++ b/README.md
@@ -1,14 +1,12 @@
 # Microsoft Authentication Library (MSAL) for Go
 
-**MSAL for Go is a new addition to the MSAL family of libraries, has been made available in production ready preview to gauge customer interest and to gather feedback from the community. We welcome all contributors (see [CONTRIBUTING.md](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/CONTRIBUTING.md)) to help us grow our list of supported MSAL SDKs.**
-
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -26,70 +24,78 @@ To install Go, visit [this link](https://golang.org/dl/).
...
+1. Create a client. MSAL separates [public and confidential client applications](https://tools.ietf.org/html/rfc6749#section-2.1), so call `public.New()` or `confidential.New()` to create the appropriate client for your application.
 
-1. MSAL separates [public and confidential client applications](https://tools.ietf.org/html/rfc6749#section-2.1). So, you would create an instance of a `PublicClientApplication` and `ConfidentialClientApplication` and use this throughout the lifetime of your application.
    * Initializing a public client:
 
     ```go
-    publicClientApp, err := public.New("client_id", public.WithAuthority("https://login.microsoft.com/Enter_The_Tenant_Name_Here"))
+    import "github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
+
+    publicClient, err := public.New("client_id", public.WithAuthority("https://login.microsoft.com/your_tenant"))
```
rayluo, line 38:
```suggestion
    publicClient, err := public.New("client_id", public.WithAuthority("https://login.microsoftonline.com/your_tenant"))
```

README.md
```diff
@@ -26,70 +24,78 @@ To install Go, visit [this link](https://golang.org/dl/).
...
+    import "github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
+
+    // confidential clients have a credential, such as a secret or a certificate
     cred, err := confidential.NewCredFromSecret("client_secret")
     if err != nil {
-        return nil, fmt.Errorf("could not create a cred from a secret: %w", err)
+        // TODO: handle error
     }
-    confidentialClientApp, err := confidential.New("client_id", cred, confidential.WithAuthority("https://login.microsoft.com/Enter_The_Tenant_Name_Here"))
+    confidentialClient, err := confidential.New("https://login.microsoft.com/your_tenant", "client_id", cred)
```
rayluo, line 51: Same as above

README.md
```diff
@@ -26,70 +24,78 @@ To install Go, visit [this link](https://golang.org/dl/).
...
+    import "github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
+
+    // confidential clients have a credential, such as a secret or a certificate
     cred, err := confidential.NewCredFromSecret("client_secret")
     if err != nil {
-        return nil, fmt.Errorf("could not create a cred from a secret: %w", err)
+        // TODO: handle error
     }
-    confidentialClientApp, err := confidential.New("client_id", cred, confidential.WithAuthority("https://login.microsoft.com/Enter_The_Tenant_Name_Here"))
+    confidentialClient, err := confidential.New("https://login.microsoftonline.com/your_tenant", "client_id", cred)
```
rayluo, line 51: Out of curiosity, the confidential.New(...) accepts the authority from the first param which is a string, but the publicNew(...) accepts authority from the second param which is an output of WithAuthority helper. Is such an asymmetry indeed the case?

README.md
```diff
@@ -26,70 +24,78 @@ To install Go, visit [this link](https://golang.org/dl/).
...
+	}
     if len(accounts) > 0 {
-        // Assuming the user wanted the first account
+        // there may be more accounts; here we assume the first one is wanted
         userAccount = accounts[0]
-        // found a cached account, now see if an applicable token has been cached
-        result, err := publicClientApp.AcquireTokenSilent(context.Background(), []string{"your_scope"}, public.WithSilentAccount(userAccount))
+    }
+    scopes := []string{"scope"}
+    result, err := publicClient.AcquireTokenSilent(context.TODO(), scopes, public.WithSilentAccount(userAccount))
```
rayluo, line 69: Since this line was moved out from the `if len(accounts) > 0 {...}` block, wouldn't `userAccount` potentially be undefined? This may not be an issue, though, if `AcquireTokenSilent()` would gracefully return an error. Is that the case here?

README.md
chlowell: Yes. Public client has a reasonable default authority but confidential client doesn't, so only the confidential client constructor requires an authority.

README.md
chlowell: `userAccount` is defined in any case (line 59) but has default ("zero") values when the cache doesn't contain an account. `AcquireTokenSilent()` returns the same error in both cases because a zero-value account is functionally the same as no account. Well, unless the cache has a token for the client itself; `AcquireTokenSilent()` would return that token when no account or a zero-value account is specified (like the `account=None` case in MSAL Python).

README.md
rayluo: Precisely speaking, only confidential client's `AcquireTokenByCredential()` requires a tenanted authority. Auth code flow can still use ".../common" or ".../organizations". That being said, the current API decision is also reasonable.

README.md
rayluo: > `userAccount` is defined in any case (line 59) but has default ("zero") values when the cache doesn't contain an account.

Thanks, now I get this part.

> `AcquireTokenSilent()` returns the same error in both cases because a zero-value account is functionally the same as no account. Well, unless the cache has a token for the client itself; `AcquireTokenSilent()` would return that token when no account or a zero-value account is specified (like the `account=None` case in MSAL Python)

Wait a minute. In MSAL Python, for better or worse, the 2 similar calling snippets `acquire_token_silent(..., account=some_account)` and `acquire_token_silent(..., account-None)` are for two conceptually different scenarios: the former is acquiring a token for a user, the latter is for a client. And we do not expect the former scenario to fall back to the latter scenario. That is also why MSAL Python's sample snippet put the `acquire_token_silent(..., account=some_account)` inside the `if accounts:` clause.

Now this MSAL Go Readme and the example apps seem to encourage that approach of "if we find no user account then fall back to get a token for this client". Is that intentional?

Aside, from Azure Identity's perspective, there is the ["chaining credentials" concept](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.chainedtokencredential?view=azure-python), which may be used to chain different identities, but is it supposed to mix a user identity and a client identity?

chlowell: > Wait a minute. In MSAL Python, for better or worse, the 2 similar calling snippets `acquire_token_silent(..., account=some_account)` and `acquire_token_silent(..., account=None)` are for two conceptually different scenarios: the former is acquiring a token for a user, the latter is for a client. And we do not expect the former scenario to fall back to the latter scenario. That is also why MSAL Python's sample snippet put the acquire_token_silent(..., account=some_account) inside the if accounts: clause.

It's similar here: public and confidential clients have separate APIs and documentation but share a silent auth implementation.

> Now this MSAL Go Readme and the example apps seem to encourage that approach of "if we find no user account then fall back to get a token for this client". Is that intentional?

The docs don't encourage that approach. I wouldn't even call it an approach because I don't think it's a behavior users would expect or rely on--it's a subtle implication of the implementation that affects a narrow, unusual scenario. Let's get concrete, here's a repro:

```go
package main

import (
	"context"

	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/cache"
	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
)

const (
	clientID = "..."
	secret   = "..."
	tenantID = "..."
)

var scopes = []string{"https://management.azure.com/.default"}

// flatCache isn't partitioned
type flatCache []byte

func (c *flatCache) Export(ctx context.Context, m cache.Marshaler, _ cache.ExportHints) error {
	v, err := m.Marshal()
	if err == nil {
		*c = v
	}
	return err
}

func (c *flatCache) Replace(ctx context.Context, u cache.Unmarshaler, _ cache.ReplaceHints) error {
	return u.Unmarshal(*c)
}

func main() {
	// create public and confidential clients for the same application, sharing the same unpartitioned cache
	authority := "https://login.microsoftonline.com/"+tenantID
	c := flatCache{}
	cred, _ := confidential.NewCredFromSecret(secret)
	cca, _ := confidential.New(authority, clientID, cred, confidential.WithCache(&c))
	pca, _ := public.New(clientID, public.WithAuthority(authority), public.WithCache(&c))

	// confidential client caches a token for the application itself
	ar, err := cca.AcquireTokenByCredential(context.Background(), scopes)
	if err != nil {
		panic(err)
	}

	// public client attempts silent authentication without specifying an account
	// (same result if the specified account is zero-valued)
	ar2, err := pca.AcquireTokenSilent(context.Background(), scopes)
	if err == nil && ar.AccessToken == ar2.AccessToken {
		// we land here, meaning the public client returns the application token
		panic("🤔 how bad is this")
	}
}
```

I think a developer would have to set this up deliberately, which implies understanding the consequences, so maybe we don't need to prevent this in MSAL. What do you think? I suppose we could have the public client's `AcquireTokenSilent` return an error when the caller doesn't provide a home account ID.

> Aside, from Azure Identity's perspective, there is the ["chaining credentials" concept](https://learn.microsoft.com/en-us/python/api/azure-identity/azure.identity.chainedtokencredential?view=azure-python), which may be used to chain different identities, but is it supposed to mix a user identity and a client identity?

A chain may include user and client identities, it's up to the application, but we don't want a user credential type to return a service principal's token out of a shared cache--we'll have to investigate whether that's possible.

rayluo: Thanks for the detailed explanation. Now that we know we are indeed talking about the same topic, let's dive deep into each aspect of this topic.

```go
 	// ... client attempts silent authentication without specifying an account
	// (same result if the specified account is zero-valued)
	ar2, err := pca.AcquireTokenSilent(context.Background(), scopes)
	if err == nil && ar.AccessToken == ar2.AccessToken {
		// we land here, meaning the public client returns the application token
		panic("🤔 how bad is this")
	}
```

Although we haven't asked a Thread Model reviewer to comment on this, the gut feeling is it would be really bad when intending to get a token for one entity (user `JohnDoe@example.com`) but the app ends up providing a token for another entity (app PasswordManager). In such an (albeit contrived) example, the PasswordManager app intended to sync the local password database into the user John Doe's cloud storage but ended up syncing it into the app's own cloud storage just because John Doe already signed out.

> The docs don't encourage that approach.
> ...
> I think a developer would have to set this up deliberately, which implies understanding the consequences.

I beg to disagree. If we wrote our docs and [sample this way](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/419/files#diff-56851c379cdbf396ea9fcfe74780951c108045cecb1991183e7ec3f6c60e6792R19-R29), the innocent developer would just copy and paste and then unknowingly run into the nasty surprise when it is too late.

> maybe we don't need to prevent this in MSAL. What do you think?

We don't have to change MSAL (more on this below). After all, it is the (only?) canonical calling pattern at this point. We just need to change our doc and sample to only run AcquireTokenSilent() inside the `if len(accounts) > 0 {...}` code path. 

> ... to prevent this in MSAL. ... I suppose we could have the public client's AcquireTokenSilent return an error when the caller doesn't provide a home account ID.

That approach alone is insufficient, because public client has no client tokens anyway, and confidential client would still run AcquireTokenSilent thus not benefit from the fix.

However, thanks for your inspiration, I think we can have a better solution which contains two parts to work together:
* let AcquireTokenSilent() always return "token not found" if the input account is absent or zero-valued,
* let AcquiretokenByCredential() automatically look up client token from the cache, and then fall back to get a new token from the wire

So that the user-flow and app-flow will have different calling pattern. It won't mix up. 

The beauty of this fix is that it is effectively still backward-compatible.

chlowell: > If we wrote our docs and sample this way, the innocent developer would just copy and paste and then unknowingly run into the nasty surprise when it is too late.

To be clear, the developer would also have to authenticate the application with a confidential client and share that client's cache with a public client, so this isn't an easy scenario to stumble into. But if there's no public client scenario for silently authenticating without a specific account, I'm happy to make that impossible with #426.

> We just need to change our doc and sample to only run AcquireTokenSilent() inside the if len(accounts) > 0 {...} code path.

#426 makes the simpler calling pattern--try silent auth first--safe.

> let AcquiretokenByCredential() automatically look up client token from the cache, and then fall back to get a new token from the wire

Can you give some more context on this? I see how it might be helpful but not the need for it.

apps/public/example_test.go
```diff
@@ -0,0 +1,46 @@
...
+	accounts, err := client.Accounts(context.TODO())
+	if err != nil {
+		// TODO: handle error
+	}
+	if len(accounts) > 0 {
+		// There may be more accounts; here we assume the first one is wanted.
+		// AcquireTokenSilent returns a non-nil error when it can't provide a token.
+		result, err = client.AcquireTokenSilent(context.TODO(), scopes, public.WithSilentAccount(accounts[0]))
+	}
+	if err != nil || len(accounts) == 0 {
```
rayluo, line 35: Thanks, @chlowell, for engaging in the recent discussion and making corresponding changes!

Nit: Was this line of now-complicated logic part of the reason that you did not want to go this route? Perhaps we can simply it like this.
```suggestion
	if result == nil {
```

Besides, the code snippet in README seems not as polished as this examples_test.go file.

Those issues are minor. I'll leave it to you to decide whether further polish it. Meanwhile, I'm approving this PR.

rayluo (approved):

README.md
rayluo: Adding a summary of an offline discussion.

The PR #426 does not harm. But it does not make much difference, either.

It is not about a PCA and a CCA sharing their cache. That is a rare - if not invalid - scenario. Hopefully, the app dev knows what they are doing if they go that route.

It is more about a CCA would always have a legit need to sometimes acquire a token for an end user, sometimes for the app itself. And the former shall not fallback to the latter. PR #426 does not help here.

> > let AcquiretokenByCredential() automatically look up client token from the cache, and then fall back to get a new token from the wire
>
> Can you give some more context on this? I see how it might be helpful but not the need for it.

I propose to change acquire_token_for_client() to always query cache automatically, AND acquire_token_silent(..., account=None) will always return TokenNotFound (which eliminates the unintended fallback-to-client-token). So that the two functions in the code snippet above will have no chance to be misused. I wrote a [PR in MSAL Python](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/577), perhaps that can help your understanding.

apps/public/example_test.go
chlowell: Yes, I don't like conditionally calling `AcquireTokenSilent` because it makes the silent auth failure case more awkward. It would be nice to examine `result` instead of `err` and `accounts` (again!) but `result` can't be compared to `nil` because it isn't a pointer. It is zero-valued when either condition is true, however we'd have to reflect to establish that.

Thanks for reminding me to copy this code to the readme 😊

Issue #393, opened 2023-02-24
Improve pull request CI


chlowell:
As configured, CI runs only when a PR is labeled or merged to `dev` or `main` (see the [linting action](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/.github/workflows/golangci-lint.yml), for example). This prevents non-maintainers from executing malicious code in CI, which is great. However, it also stops CI from running automatically when a PR is updated with new commits--a maintainer must add or remove a label to trigger CI. There may be ways to improve this with Action configuration. We could also separate the "safe" tests with no access to live resources or secrets and run those on PR updates.

bgavrilMS:
> As configured, CI runs only when a PR is labeled or merged to `dev` or `main` (see the [linting action](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/dev/.github/workflows/golangci-lint.yml), for example). This prevents non-maintainers from executing malicious code in CI, which is great. However, it also stops CI from running automatically when a PR is updated with new commits--a maintainer must add or remove a label to trigger CI. There may be ways to improve this with Action configuration. We could also separate the "safe" tests with no access to live resources or secrets and run those on PR updates.

It should be sufficient to disable PR builds for PRs originating from forks.

chlowell:
I want to run CI on PRs by default though, and run it again on each push, to prevent merging broken code. I assume the motivation for the current setup was to avoid exposing secrets to everyone who opens a PR, however it's unnecessary because GitHub doesn't pass secrets to workflows triggered by PRs from a fork anyway.

There's more work here than reconfiguring the Action triggers though--we also need to skip the integration tests for external PRs because those tests need the secrets and will just fail without them ([for example](https://github.com/AzureAD/microsoft-authentication-library-for-go/actions/runs/5248941471/jobs/9481156504?pr=422)). The simple solution is to skip these tests when the secrets are absent, however that would enable a misconfigured run to succeed without trying all tests; it would be best to make the misconfiguration obvious by failing the run.

bgavrilMS:
In my experience, the number of contributors slowly drops as the complexity of the library increases and rarely do contributors write tests anyway. So might as well optimize for your daily flow.

PR #429, opened 2023-06-25
skip tests when necessary environment variables not set

### LINKED ISSUE #431: Tests fail
If you clone this repo and run the tests `go test ./...` they fail. This is unexpected and unwelcoming to contributors.

On investigation this happens because  special environment variables are not set.

This is also an obstacle to packaging on Linux.

hickford:
Unit tests should succeed without any setup

Also skip very long-running performance test unless explicit environment variable is tested

Fixes #431

### DIFF
--- a/apps/tests/integration/integration_test.go
+++ b/apps/tests/integration/integration_test.go
@@ -88,9 +88,16 @@ type secret struct {
 	Value string `json:"value"`
 }
 
-func newLabClient() (*labClient, error) {
+func newLabClient(t *testing.T) (*labClient, error) {
 	clientID := os.Getenv("clientId")
 	secret := os.Getenv("clientSecret")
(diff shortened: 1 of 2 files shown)


chlowell (changes requested):
Thanks for doing this. It would be great to have `go test ./...` just do the right thing, however I'm wary of this change by itself because it would enable misconfigured CI runs to succeed without trying these tests. I think we should address this as part of #393

hickford:
> I'm wary of this change by itself because it would enable misconfigured CI runs to succeed without trying these tests.

@chlowell Good point. Addressed in latest commit by failing test instead of skipping when on CI.

bgavrilMS (approved):

hickford:
@chlowell @bgavrilMS Who can merge?

PR #432, opened 2023-06-25
refactor performance test to be standard golang benchmarks

hickford:
Example output

```
goos: linux
goarch: amd64
pkg: github.com/AzureAD/microsoft-authentication-library-for-go/apps/tests/performance
cpu: 12th Gen Intel(R) Core(TM) i7-1265U
Benchmark1User10000Tokens-12                2476            414902 ns/op

Benchmark1User100000Tokens-12      
     100          16092443 ns/op
Benchmark10Users10000Tokens-12               716           1592947 ns/op
Benchmark100Users10000Tokens-12              654           1681811 ns/op
Benchmark1000Users1000Tokens-12             8030            142265 ns/op
Benchmark10000Users100Tokens-12            74020             18107 ns/op
PASS
ok      github.com/AzureAD/microsoft-authentication-library-for-go/apps/tests/performance       47.578s
```


chlowell:
Thanks for doing this! This is nitpicking, but it appears these benchmarks could be consolidated as sub-benchmarks parameterized by user and token counts i.e., something like this:
```go
func BenchmarkCache(b *testing.B) {
	for _, bm := range []struct {
		users, tokens int
	}{
		{100, 10_000},
	} {
		b.Run(fmt.Sprintf("%d users %d tokens", bm.users, bm.tokens), func(b *testing.B) {
			// inline benchmark() and perhaps also queryCache()
		})
	}
}
```
...which would be tidier. I haven't tried it though, maybe there's a catch. What do you think?

hickford:
Great idea, I agree that’s better

bgavrilMS:
Would be nice to integrate with https://github.com/benchmark-action/github-action-benchmark

bgavrilMS:
Do we still want to merge this in?

chlowell:
I wouldn't block anything on it but I think we should merge this (after tidying it up into sub-benchmarks as I described above). It's weird to benchmark with a unit test printing to console instead of go's standard benchmarking framework.

hickford:
> after tidying it up into sub-benchmarks

@chlowell Done, ready for review

Issue #439, opened 2023-06-30
[Feature Request] Extensibility for POP with SHR


bgavrilMS:
Request for MSAL to enable extensibility for POP + SHR. This is a request from Kubernetes team.

Proposed interface for extensibility: 

```csharp
    interface IAuthenticationScheme
    {
        /// <summary>
        /// Extra parameters that are added to the request to the /token endpoint. 
        /// For POP this would be req_cnf=(public key) and token_type=pop
        /// </summary>
        IReadOnlyDictionary<string, string> GetTokenRequestParams();

        /// <summary>
        /// Key ID to which to associate the ESTS token response with. Used by MSAL for caching puroposes.
        /// </summary>
        string KeyId { get; }

        /// <summary>
        /// Creates the access token that goes into an Authorization HTTP header. For POP this means adding the SHR claims and signing 
        /// the token with the client key. 
        /// </summary>
        string FormatAccessToken(string accessToken); 
    }
```
See [POP implementation](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AuthScheme/PoP/PoPAuthenticationScheme.cs) for this interface in MSAL.NET, which does in-memory key management as well.

MSAL changes: 

- [ ] Add new API to all requests in MSAL GO, e.g. `AcquireTokenForClient(...).WithAuthenticationScheme(popScheme)`
- [ ] Add the parameters from `GetTokenRequestParams` to the /token request as POST data
- [ ] When token response comes back from ESTS, associate the access token to the `KeyId`. Make sure this association is reflected in the cache (see below)
- [ ] Before constructing the `AuthenticationResult`, format the access token by invoking `FormatAccessToken` (this will create the actual POP token. Remember that MSAL doesn't cache the formatted POP token!)
- [ ] Update token cache lookup logic to take into account `keyId`
- [ ] If token is found in cache and matches `keyId`, make sure to call `FormatAccessToken` before constructing `AuthenticationResult`

Changes to the cache

- [ ] Add -{authschme} to the access token key. This is to differentiate between bearer and pop tokens. Only 1 pop token will be supported. If this part of the key is not present, assume the token is "Bearer". 
- [ ] Add key to the payload to associate the token with the kid.
- [ ] Ensure token type is "pop" (AAD returns the token_type)

Example pop access token in cache:

```json
{
  "AccessToken": {
[...]

PR #442, opened 2023-07-12
Draft: Support AuthenticationScheme interface to enable PoP bound access tokens

manojampalam:
PR attempting to realize proposal - https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/439

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -314,8 +314,11 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	// ignore cached access tokens when given claims
 	if silent.Claims == "" {
 		ar, err = AuthResultFromStorage(storageTokenResponse)
-		if authParams.AuthenticationScheme != nil {
-			ar.AccessToken, _ = authParams.AuthenticationScheme.FormatAccessToken(ar.AccessToken)
+		if err == nil && authParams.AuthenticationScheme != nil {
+			ar.AccessToken, err = authParams.AuthenticationScheme.FormatAccessToken(ar.AccessToken)
+			if err != nil {
+				return ar, err
+			}
 		}
 		if err == nil {
 			return ar, err
@@ -423,7 +426,10 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 		err = b.cacheAccessor.Export(ctx, b.manager, cache.ExportHints{PartitionKey: key})
 	}
 	if authParams.AuthenticationScheme != nil {
-		ar.AccessToken, _ = authParams.AuthenticationScheme.FormatAccessToken(ar.AccessToken)
+		ar.AccessToken, err = authParams.AuthenticationScheme.FormatAccessToken(ar.AccessToken)
+		if err != nil {
+			return AuthResult{}, err
+		}
 	}
 	return ar, err
 }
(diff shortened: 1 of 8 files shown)


apps/internal/oauth/ops/authority/authority.go
```diff
@@ -180,6 +187,8 @@ type AuthParams struct {
 	LoginHint string
 	// DomainHint is a directive that can be used to accelerate the user to their federated IdP sign-in page
 	DomainHint string
+	// Authn Scheme
+	AuthenticationScheme
```
julienstroheker, line 191: we might want this to be more explicit

apps/public/public.go
```diff
@@ -210,6 +212,33 @@ func WithClaims(claims string) interface {
 	}
 }
 
+// WithAuthenticationScheme
+func WithAuthenticationScheme(authnScheme AuthenticationScheme) interface {
```
julienstroheker, line 216: UT ?

julienstroheker (commented):
we should also add an example apps/tests/devapps on how to consume this AuthenticationScheme

bgavrilMS:
@manojampalam - I gave you write access to this repo if you want to move away from a fork.

apps/internal/base/base.go
```diff
@@ -312,6 +314,12 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	// ignore cached access tokens when given claims
 	if silent.Claims == "" {
 		ar, err = AuthResultFromStorage(storageTokenResponse)
+		if err == nil && authParams.AuthenticationScheme != nil {
```
bgavrilMS, line 317: We need to make sure that the token cache schema is the same across MSALs. The simplest is to serialize the cache to a plaintext file and observe it's structure. Note that each access token has a "cache key" and a "cache payload".

For POP tokens, we don't want the `kid` to be part of the access token cache key. This is to prevent the token cache becoming too large, as keys can rotate frequently. 

The changes needed are: 

- add the "schema name" (i.e. "pop") in the key. You'll have to add it to `IAuthenticationScheme` as well. 
- add the `kid` to the payload payload
- ensure `token_type` is in the payload as well

```json
"AccessToken": {
    "-login.windows.net-accesstoken_with_authscheme-16dab2ba-145d-4b1b-8569-bf4b9aed4dc8-72f988bf-86f1-41af-91ab-2d7cd011db47-https://vault.azure.net/.default-pop": {
      "environment": "login.windows.net",
      "client_id": "16dab2ba-145d-4b1b-8569-bf4b9aed4dc8",
      "secret": "actual_token",
      "credential_type": "AccessToken",
      "realm": "72f988bf-86f1-41af-91ab-2d7cd011db47",
      "target": "https://vault.azure.net/.default",
      "cached_at": "1688149697",
      "expires_on": "1688236096",
      "extended_expires_on": "1688236096",
      "kid": "TJtUe-3Wa8X6cOxKzmeTxjHDO1-gXBKtazWh0xx3SA0",    # new!
      "token_type": "pop",   # not Bearer!
      "refresh_on": "1688192896",
      "ext_expires_on": "1688236096"
    }
```

manojampalam:
Moving changes and comments to PR in main repo https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/443

PR #443, opened 2023-07-18
Draft: Support AuthenticationScheme interface to enable PoP bound access tokens

manojampalam:
PR attempting to realize proposal - https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/439 

This is to gather initial feedback. Comprehensive unit and integration tests will be added once code changes are agreed/acked.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -669,7 +669,7 @@ func (cca Client) AcquireTokenByCredential(ctx context.Context, scopes []string,
 	authParams.Scopes = scopes
 	authParams.AuthorizationType = authority.ATClientCredentials
 	authParams.Claims = o.claims
-	authParams.AuthenticationScheme = o.authnScheme
+	authParams.AuthnScheme = o.authnScheme
 	token, err := cca.base.Token.Credential(ctx, authParams, cca.cred)
 	if err != nil {
 		return AuthResult{}, err
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -59,6 +59,8 @@ added, it doesn't exist in real life. As such I've put a PEM decoder into here.
 // For details see https://aka.ms/msal-net-authenticationresult
 type AuthResult = base.AuthResult
 
+type AuthenticationScheme = authority.AuthenticationScheme
+
 type Account = shared.Account
 
 // CertFromPEM converts a PEM file (.pem or .key) for use with [NewCredFromCert]. The file
@@ -454,6 +456,32 @@ func WithClaims(claims string) interface {
 	}
 }
 
+func WithAuthenticationScheme(authScheme AuthenticationScheme) interface {
+	AcquireSilentOption
+	AcquireByCredentialOption
+	options.CallOption
+} {
+	return struct {
+		AcquireSilentOption
+		AcquireByCredentialOption
+		options.CallOption
+	}{
+		CallOption: options.NewCallOption(
+			func(a any) error {
+				switch t := a.(type) {
+				case *acquireTokenSilentOptions:
+					t.authScheme = authScheme
+				case *acquireTokenByCredentialOptions:
+					t.authScheme = authScheme
+				default:
+					return fmt.Errorf("unexpected options type %T", a)
+				}
+				return nil
+			},
+		),
+	}
+}
+
 // WithTenantID specifies a tenant for a single authentication. It may be different than the tenant set in [New].
 // This option is valid for any token acquisition method.
 func WithTenantID(tenantID string) interface {
@@ -499,6 +527,7 @@ func WithTenantID(tenantID string) interface {
 type acquireTokenSilentOptions struct {
 	account          Account
 	claims, tenantID string
+	authScheme       AuthenticationScheme
 }
 
 // AcquireSilentOption is implemented by options for AcquireTokenSilent
@@ -549,6 +578,7 @@ func (cca Client) AcquireTokenSilent(ctx context.Context, scopes []string, opts 
 		Credential:  cca.cred,
 		IsAppCache:  o.account.IsZero(),
 		TenantID:    o.tenantID,
+		AuthnScheme: o.authScheme,
 	}
 
 	return cca.base.AcquireTokenSilent(ctx, silentParameters)
@@ -614,6 +644,7 @@ func (cca Client) AcquireTokenByAuthCode(ctx context.Context, code string, redir
 // acquireTokenByCredentialOptions contains optional configuration for AcquireTokenByCredential
 type acquireTokenByCredentialOptions struct {
 	claims, tenantID string
+	authScheme       AuthenticationScheme
 }
 
 // AcquireByCredentialOption is implemented by options for AcquireTokenByCredential
@@ -637,7 +668,7 @@ func (cca Client) AcquireTokenByCredential(ctx context.Context, scopes []string,
 	authParams.Scopes = scopes
 	authParams.AuthorizationType = authority.ATClientCredentials
 	authParams.Claims = o.claims
-
+	authParams.AuthenticationScheme = o.authScheme
 	token, err := cca.base.Token.Credential(ctx, authParams, cca.cred)
 	if err != nil {
 		return AuthResult{}, err
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -456,7 +456,8 @@ func WithClaims(claims string) interface {
 	}
 }
 
-func WithAuthenticationScheme(authScheme AuthenticationScheme) interface {
+// WithAuthenticationScheme
+func WithAuthenticationScheme(authnScheme AuthenticationScheme) interface {
 	AcquireSilentOption
 	AcquireByCredentialOption
 	options.CallOption
@@ -470,9 +471,9 @@ func WithAuthenticationScheme(authScheme AuthenticationScheme) interface {
 			func(a any) error {
 				switch t := a.(type) {
 				case *acquireTokenSilentOptions:
-					t.authScheme = authScheme
+					t.authnScheme = authnScheme
 				case *acquireTokenByCredentialOptions:
-					t.authScheme = authScheme
+					t.authnScheme = authnScheme
 				default:
 					return fmt.Errorf("unexpected options type %T", a)
 				}
@@ -527,7 +528,7 @@ func WithTenantID(tenantID string) interface {
 type acquireTokenSilentOptions struct {
 	account          Account
 	claims, tenantID string
-	authScheme       AuthenticationScheme
+	authnScheme      AuthenticationScheme
 }
 
 // AcquireSilentOption is implemented by options for AcquireTokenSilent
@@ -578,7 +579,7 @@ func (cca Client) AcquireTokenSilent(ctx context.Context, scopes []string, opts 
 		Credential:  cca.cred,
 		IsAppCache:  o.account.IsZero(),
 		TenantID:    o.tenantID,
-		AuthnScheme: o.authScheme,
+		AuthnScheme: o.authnScheme,
 	}
 
 	return cca.base.AcquireTokenSilent(ctx, silentParameters)
@@ -644,7 +645,7 @@ func (cca Client) AcquireTokenByAuthCode(ctx context.Context, code string, redir
 // acquireTokenByCredentialOptions contains optional configuration for AcquireTokenByCredential
 type acquireTokenByCredentialOptions struct {
 	claims, tenantID string
-	authScheme       AuthenticationScheme
+	authnScheme      AuthenticationScheme
 }
 
 // AcquireByCredentialOption is implemented by options for AcquireTokenByCredential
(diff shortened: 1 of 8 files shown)


apps/internal/base/base.go
```diff
@@ -313,7 +327,7 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	if silent.Claims == "" {
 		ar, err = AuthResultFromStorage(storageTokenResponse)
 		if err == nil {
-			return ar, err
+			return ar.ApplyAuthnScheme(&authParams)
 		}
```
manojampalam, line 331: From @bgavrilMS 

We need to make sure that the token cache schema is the same across MSALs. The simplest is to serialize the cache to a plaintext file and observe it's structure. Note that each access token has a "cache key" and a "cache payload".

For POP tokens, we don't want the `kid` to be part of the access token cache key. This is to prevent the token cache becoming too large, as keys can rotate frequently. 

The changes needed are: 

- add the "schema name" (i.e. "pop") in the key. You'll have to add it to `IAuthenticationScheme` as well. 
- add the `kid` to the payload payload
- ensure `token_type` is in the payload as well

```json
"AccessToken": {
    "-login.windows.net-accesstoken_with_authscheme-16dab2ba-145d-4b1b-8569-bf4b9aed4dc8-72f988bf-86f1-41af-91ab-2d7cd011db47-https://vault.azure.net/.default-pop": {
      "environment": "login.windows.net",
      "client_id": "16dab2ba-145d-4b1b-8569-bf4b9aed4dc8",
      "secret": "actual_token",
      "credential_type": "AccessToken",
      "realm": "72f988bf-86f1-41af-91ab-2d7cd011db47",
      "target": "https://vault.azure.net/.default",
      "cached_at": "1688149697",
      "expires_on": "1688236096",
      "extended_expires_on": "1688236096",
      "kid": "TJtUe-3Wa8X6cOxKzmeTxjHDO1-gXBKtazWh0xx3SA0",    # new!
      "token_type": "pop",   # not Bearer!
      "refresh_on": "1688192896",
      "ext_expires_on": "1688236096"
    }
```

bgavrilMS: In the current design in MSALs, access token bound to key1 overwrites access token bound to key2. Is this a problem?

manojampalam: I think so. There can be races around pop key rotation. Ex. around a time pop key rotates from pop_old to pop_new, and  Aq*Token(pop_old) and Aq*Token(pop_new) come in together, MSAL wouldn't know which token to cache and may end up re-writing cache entry w/ pop_old based entry.

bgavrilMS: These cache keys are for MSAL internal use only, not for distributed caches. The race would resolve itself on a single machine pretty quickly. 

A cache key for the distributed cache should include the kid, I agree. MSAL Go doesn't have explicit support for this scenario, yet.

manojampalam: @bgavrilMS still working to incorporate your feedback and assess impact on clients due to possible races.

manojampalam: @bgavrilMS pushed changes inline with your asks -
 - replaced  `keyid` from accesstoken cache key, replaced with `token_type`
 - added `keyid` to `token_type` to payload. 
 - added `AccessTokenType()` to AuthenticationScheme interface (inline with similar function name in msal .net)

apps/confidential/confidential_test.go
```diff
@@ -1225,3 +1227,33 @@ func TestWithDomainHint(t *testing.T) {
...
+		ExtExpiresOn:  internalTime.DurationTime{T: time.Now().Add(1 * time.Hour)},
+		GrantedScopes: accesstokens.Scopes{Slice: tokenScope},
+		TokenType:     "TokenType",
+	}, cred)
+	if err != nil {
+		t.Fatal(err)
+	}
+	result, err := client.AcquireTokenByCredential(ctx, tokenScope, WithAuthenticationScheme(authScheme))
+	if err != nil {
+		t.Fatal("silent authentication should fail because the cache is empty")
```
chlowell, line 1250:
```suggestion
		t.Fatal(err)
```

apps/confidential/confidential_test.go
```diff
@@ -1225,3 +1227,33 @@ func TestWithDomainHint(t *testing.T) {
...
+	}
+	result, err = client.AcquireTokenSilent(ctx, tokenScope, WithAuthenticationScheme(authScheme))
+	if err != nil {
+		t.Fatal(err)
+	}
+	if result.AccessToken != token+"-formated" {
```
chlowell, line 1256: Should we check this after `AcquireTokenByCredential` as well?

apps/confidential/confidential_test.go
```diff
@@ -108,6 +108,7 @@ func fakeClient(tk accesstokens.TokenResponse, credential Credential, options ..
 			fakeTokenEndpoint, "https://fake_authority/fake/jwt", "fake_authority"),
 	}
 	client.base.Token.WSTrust = &fake.WSTrust{}
+	client.base.AuthParams.AuthnScheme = &authority.BearerAuthenticationScheme{}
```
chlowell, line 111: This is redundant, no?
```suggestion
```

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -138,6 +138,38 @@ const (
 	ADFS = "ADFS"
 )
 
+type AuthenticationScheme interface {
```
chlowell, line 141: Just a reminder that all interface changes are breaking--if there's a possibility we'll want to add, remove, or tweak some aspect of this in the future, it shouldn't be an interface

apps/tests/integration/integration_test.go
```diff
@@ -409,3 +410,37 @@ func TestRemoveAccount(t *testing.T) {
 	}
 
 }
+
+func TestAuthnScheme(t *testing.T) {
```
chlowell, line 414: Consider deleting this test--it doesn't add any interesting coverage because it just explicitly specifies the default scheme. Many other tests implicitly test the default scheme and we don't need a live test to validate that custom schemes are used correctly

apps/confidential/confidential.go
```diff
@@ -454,6 +456,33 @@ func WithClaims(claims string) interface {
 	}
 }
 
+// WithAuthenticationScheme
```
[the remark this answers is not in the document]
bgavrilMS: Can we add a remark to the docs stating: "This is an extensibility mechanism designed to be used only by Azure Arc for proof of possession tokens"


We have a similar contract with Azure SDK as well in MSALs, see for example https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Extensibility/ConfidentialClientApplicationBuilderExtensions.cs#L25

Eventually MSALs should support POP out of the box and we'll want to deprecate this.

apps/internal/base/base.go
```diff
@@ -90,6 +91,13 @@ type AuthResult struct {
 	DeclinedScopes []string
 }
 
+func (ar *AuthResult) ApplyAuthnScheme(params *authority.AuthParams) (AuthResult, error) {
```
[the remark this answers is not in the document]
julienstroheker: @manojampalam ?

apps/tests/integration/integration_test.go
julienstroheker: fair enough I removed it.

apps/confidential/confidential.go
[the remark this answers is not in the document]
julienstroheker: I added a comment on top of the function is that enough or you want me to do more ?

apps/tests/integration/integration_test.go
julienstroheker: Should I also remove the sample then under `test/devapps` ?

apps/internal/oauth/ops/authority/authority.go
julienstroheker: @manojampalam

apps/tests/integration/integration_test.go
chlowell: If customers shouldn't use this feature, I think we shouldn't provide sample code for it.

julienstroheker: fair enough. I'll remove it.

apps/internal/base/base.go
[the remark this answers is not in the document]
manojampalam: Agree, prior iteration of FormatAccessToken had more logic within, hence made sense then. Will get rid of it.

apps/internal/oauth/ops/authority/authority.go
manojampalam: good point. @bgavrilMS your thoughts?  Equivalent interface in .Net is [internal ](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/c89dd200d3c1e4f6f171106403509d685dec01e0/src/client/Microsoft.Identity.Client/AuthScheme/IAuthenticationScheme.cs#L12) at this point, as PoP support is built in via its SHR feature. 

Otherwise, I feel, this interface as such is sufficient for the purpose of client implemented PoP enablement. If at all we need more functionality in future, we could have it via a new option based on AuthenticationSchemeV2 ?

apps/confidential/confidential.go
```diff
@@ -454,6 +456,33 @@ func WithClaims(claims string) interface {
 	}
 }
 
+// WithAuthenticationScheme is an extensibility mechanism designed to be used only by Azure Arc for proof of possession tokens.
```
bgavrilMS, line 459:
```suggestion
// WithAuthenticationScheme is an extensibility mechanism designed to be used only by Azure Arc for proof of possession access tokens.
```

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -138,6 +138,38 @@ const (
...
+	// Extra parameters that are added to the request to the /token endpoint.
+	TokenRequestParams() map[string]string
+	// Key ID of the public / private key pair used by the encryption algorithm, if any.
+	// Tokens obtained by authentication schemes that use this are bound to the KeyId, i.e.
+	// if a different kid is presented, the access token cannot be used.
+	KeyID() string
+	// Creates the access token that goes into an Authorization HTTP header.
+	FormatAccessToken(accessToken string) (string, error)
```
bgavrilMS, line 149: @manojampalam - I liked your idea of being able to modify the entire AuthenticaitonResult object, not just the access token. But this can work as well, not a blocking comment.

apps/internal/oauth/ops/authority/authority.go
bgavrilMS: This is one of the reasons why we put in the disclaimer that this API is an extensbility point created for a very specific scenario. In the off-chance we need to modify this interface, we could either go via the `AuthenticationScheme2` route or  introduce a breaking change in an API that isn't meant for wide consumption. We have done this in the past in MSAL.NET and it worked ok. We can discuss in more detail and decide if a breaking change arises.

apps/internal/base/internal/storage/items_test.go
```diff
@@ -54,14 +55,16 @@ func TestCreateAccessToken(t *testing.T) {
...
+		"keyID",
 	)
 	if !extExpiresOn.Equal(actualAt.ExtendedExpiresOn.T) {
 		t.Errorf("Actual ext expires on %s differs from expected ext expires on %s", actualAt.ExtendedExpiresOn, extExpiresOn)
 	}
 }
 
 func TestKeyForAccessToken(t *testing.T) {
-	const want = "testhid-env-accesstoken-clientid-realm-user.read"
+	const want = "testhid-env-accesstoken-clientid-realm-user.read-bearer"
```
bgavrilMS, line 67: Can we make sure (via a test) that the old format is usable? I.e. if I have a token cache item without "-bearer" in the key, MSAL can read it and know that this is a bearer token.

bgavrilMS, line 67: I am concerned that this will break backwards  and forwards compatibility with the token cache. It's ok if older MSALs can't read POP tokens, but they should continue to read bearer tokens. As such, when the auth scheme is beaerr, we should leave the key alone. 

Here's the equivalent logic in MSAL.NET https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AuthScheme/AuthSchemeHelper.cs

We should have some tests on this: 

- old cache format + new MSAL
- new cache format (i.e. with POP and Bearer tokens )+ old MSAL (I guess this will have to be manual test)

apps/internal/base/base.go
bgavrilMS: Reviewed and left a comment. The access token cache key should not include "-Bearer" (for backwards compat). Only chnage the key for pop tokens. See other comment for more details.

apps/internal/base/base.go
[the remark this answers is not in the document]
manojampalam: removed

apps/internal/base/internal/storage/items_test.go
manojampalam: Thanks. Fixed how key is generated in items.go. Updated test.

manojampalam: Where can we find steps to to simulate
   _new cache format (i.e. with POP and Bearer tokens )+ old MSAL (I guess this will have to be manual test)_

apps/internal/base/base.go
manojampalam: Done.

chlowell (approved):

julienstroheker:
We were able to validate this changes e2e on our side using [Kubelogin](https://github.com/Azure/kubelogin) to generate the token and [Guard](https://github.com/kubeguard/guard) to validates it.

[Kubelogin PR In Progress](https://github.com/Azure/kubelogin/pull/319)

Token generated by kubelogin via interactive mode : 
image


Token validated by Guard :
image

Token expired rejected by Guard : 
image

`u` claim mismatch rejected by Guard : 
image

PR #446, opened 2023-08-05
Fix AcquireTokenSilent for ADFS users

chlowell:
#426 broke silent auth for ADFS users by requiring a home account ID, which ADFS accounts don't have. This PR restores the v1.0.0 behavior for ADFS by relaxing that requirement, allowing any nonzero account for silent auth. However, I'm uncertain the prior behavior was correct. Because the home account ID for ADFS accounts and tokens is always `""`, the cache matches access tokens on realm, client ID and scope alone, effectively ignoring the user's identity. If the cache has tokens for more than one ADFS user, it will return the first token it considers regardless of the owner. @rayluo @bgavrilMS should the cache match something else in this case such as the local account ID?

### DIFF
--- a/apps/public/public.go
+++ b/apps/public/public.go
@@ -27,6 +27,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"reflect"
 	"strconv"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/cache"
@@ -297,8 +298,8 @@ func (pca Client) AcquireTokenSilent(ctx context.Context, scopes []string, opts 
 	if err := options.ApplyOptions(&o, opts); err != nil {
 		return AuthResult{}, err
 	}
-	// a home account ID is required to find user tokens in the cache
-	if o.account.HomeAccountID == "" {
+	// an account is required to find user tokens in the cache
+	if reflect.ValueOf(o.account).IsZero() {
 		return AuthResult{}, errNoAccount
 	}
 
--- a/apps/confidential/confidential_test.go
+++ b/apps/confidential/confidential_test.go
@@ -370,6 +370,73 @@ func TestAcquireTokenSilentTenants(t *testing.T) {
 	}
 }
 
+func TestADFSTokenCaching(t *testing.T) {
+	cred, err := NewCredFromSecret(fakeSecret)
+	if err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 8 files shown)


bgavrilMS:
> #426 broke silent auth for ADFS users by requiring a home account ID, which ADFS accounts don't have. This PR restores the v1.0.0 behavior for ADFS by relaxing that requirement, allowing any nonzero account for silent auth. However, I'm uncertain the prior behavior was correct. Because the home account ID for ADFS accounts and tokens is always `""`, the cache matches access tokens on realm, client ID and scope alone, effectively ignoring the user's identity. If the cache has tokens for more than one ADFS user, it will return the first token it considers regardless of the owner. @rayluo @bgavrilMS should the cache match something else in this case such as the local account ID?

How can you tell between 2 users then?!

bgavrilMS:
No, this is not the correct approach. The id of an ADFS account it based on the `sub` claim in the id token. THis is a mandatory OpenIdConnect claim, guaranteed to be there. 

See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/TokenResponseHelper.cs#L48

```csharp
 string homeAccountId = clientInfo?.ToAccountIdentifier() ?? idToken?.Subject;
```

@rayluo  - please confirm if this is the logic that MSAL Py uses.

bgavrilMS (changes requested):
Use `sub` claim as home account id

bgavrilMS (approved):

apps/public/public.go
```diff
@@ -297,8 +298,8 @@ func (pca Client) AcquireTokenSilent(ctx context.Context, scopes []string, opts
 	if err := options.ApplyOptions(&o, opts); err != nil {
 		return AuthResult{}, err
 	}
-	// a home account ID is required to find user tokens in the cache
-	if o.account.HomeAccountID == "" {
+	// an account is required to find user tokens in the cache
+	if reflect.ValueOf(o.account).IsZero() {
 		return AuthResult{}, errNoAccount
```
bdwyertech, line 303: @chlowell would you guys consider exposing this errNoAccount as ErrNoAccount so us devs can use errors.Is()?  I'm doing a string comparison currently and it feels messy.

https://github.com/bdwyertech/go-az/blob/529fe3c0feca36fc1cab1e74774222f589c9ecc3/pkg/az/auth.go#L117

PR #456, opened 2023-09-06
Enable unmarshaling partitioned caches

### LINKED ISSUE #455: Can't unmarshal a partitioned cache
Easy to repro with a unit test in `partitioned_storage_test.go`:
```go
func TestMarshal(t *testing.T) {
	mgr := newPartitionedManagerForTest(nil)
	b, err := mgr.Marshal()
	if err != nil {
		t.Fatal(err)
	}
	err = mgr.Unmarshal(b)
	if err != nil {
		t.Fatal(err)
	}
}
```
`Unmarshal` returns the error `Unmarshal(*storage.InMemoryContract) only supports structs that have the field AdditionalFields or implements json.Unmarshaler`. And that's spot on, [`storage.InMemoryContract`](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/0b07e1c7710e0f3512543260f8608583e3ec7818/apps/internal/base/internal/storage/items.go#L35) doesn't have an `AdditionalFields` field.

chlowell:
Closes #455 by adding `AdditionalFields` and JSON tags.

### DIFF
--- a/apps/internal/base/internal/storage/items.go
+++ b/apps/internal/base/internal/storage/items.go
@@ -33,11 +33,13 @@ type Contract struct {
 // the internal cache. This design is shared between MSAL versions in many languages.
 // This cannot be changed without design that includes other SDKs.
 type InMemoryContract struct {
-	AccessTokensPartition  map[string]map[string]AccessToken
-	RefreshTokensPartition map[string]map[string]accesstokens.RefreshToken
-	IDTokensPartition      map[string]map[string]IDToken
-	AccountsPartition      map[string]map[string]shared.Account
-	AppMetaData            map[string]AppMetaData
+	AccessTokensPartition  map[string]map[string]AccessToken               `json:"AccessTokensPartition,omitempty"`
+	RefreshTokensPartition map[string]map[string]accesstokens.RefreshToken `json:"RefreshTokensPartition,omitempty"`
+	IDTokensPartition      map[string]map[string]IDToken                   `json:"IdTokensPartition,omitempty"`
+	AccountsPartition      map[string]map[string]shared.Account            `json:"AccountsPartition,omitempty"`
+	AppMetaData            map[string]AppMetaData                          `json:"AppMetadata,omitempty"`
+
+	AdditionalFields map[string]interface{}
 }
 
 // NewContract is the constructor for Contract.
--- a/apps/internal/base/internal/storage/partitioned_storage_test.go
+++ b/apps/internal/base/internal/storage/partitioned_storage_test.go
@@ -6,6 +6,7 @@ package storage
 import (
 	"context"
 	"fmt"
+	"os"
 	"reflect"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


apps/internal/base/internal/storage/items.go
```diff
@@ -33,11 +33,13 @@ type Contract struct {
 // the internal cache. This design is shared between MSAL versions in many languages.
 // This cannot be changed without design that includes other SDKs.
 type InMemoryContract struct {
-	AccessTokensPartition  map[string]map[string]AccessToken
-	RefreshTokensPartition map[string]map[string]accesstokens.RefreshToken
-	IDTokensPartition      map[string]map[string]IDToken
-	AccountsPartition      map[string]map[string]shared.Account
-	AppMetaData            map[string]AppMetaData
+	AccessTokensPartition  map[string]map[string]AccessToken               `json:"AccessTokensPartition,omitempty"`
```
chlowell, line 36: @bgavrilMS is there a schema for partitioned caches or an example of how one should look serialized? I didn't find any, so I don't know the correct keys here.

bgavrilMS: The serialized partitioned cache should look **exactly** like the non-partitioned cache. The partitioned cache is just an implementation detail in MSAL, so that MSAL is able to perform fast look-up. 

You do this by going through the entire double dictionary and fetching the tokens. The first "key" is the partition key (ignore it) and the second key is the access token key - identical to the non-partitioned cache.

E.g. for an app access token you can have: 

partition key: clientId + tenantId1  
access token key: clientId + tenantId1 + resource1 and clientId + tenantId1 + resource2

partition key: clientId + tenantId2
access token key: clientId + tenantId2 + resource1 and clientId + tenantId2 + resource2

But in the serialized cache you will have only the 4 tokens with their access token keys.

@pmaytak can help here as well.

chlowell: You mean that MSAL has a collection of partitions, each partition is indistinguishable from a non-partitioned cache, and MSAL should de/serialize only one partition at a time? That is to say, the partition key is an input to de/serialize?

pmaytak: Yes, partition key is a cache entry key to de/serialize.

Something like this:
image

See [app cache implementation in MSAL.NET](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/90ae6b9c6d33655a66cbc6413e8aadee487f052a/src/client/Microsoft.Identity.Client/PlatformsCommon/Shared/InMemoryPartitionedAppTokenCacheAccessor.cs#L25)

chlowell: MSAL Go partitions data in memory as expected however it always de/serializes all partitions. There's no way to de/serialize a single partition. Let's not proceed with this PR then, much more is required to make persistent caching work correctly with partitions.

pmaytak: MSAL.NET also de/serializes all partitions (so full internal cache). However, an L2 read only ever loads one partition. So as long as CCA is created per request, there's only ever 1 partition in the cache.

chlowell: I'm missing something here. MSAL Go gives the storage implementation opaque bytes and a suggested partition key, implying the bytes represent a single partition. If MSAL always de/serializes all partitions, a storage implementation that divided the cache across physical storage according to MSAL's suggested keys would just create redundant copies of the entire cache, no?

pmaytak: > If MSAL always de/serializes all partitions, a storage implementation that divided the cache across physical storage according to MSAL's suggested keys would just create redundant copies of the entire cache, no?

Yes, if MSAL instance (with the internal cache) is used as a singleton - then the external cache entries can have duplicate data. That's why we tell to create a new confidential client instance per request, so that the cache is empty initially and is deserialized with one partition.

> gives the storage implementation opaque bytes and a suggested partition key, implying the bytes represent a single partition

This should be how it works, but MSAL.NET's implementation was not ideal. Ideally the serializer will accept a cache key, and try to serialize only that.

PR #457, opened 2023-09-11
Correct option name in error message

chlowell:
Whoops, should be With**Silent**Account


chlowell:
@bgavril integration tests failed due to an invalid value of LAB_APP_CLIENT_SECRET (one of this repo's Actions secrets). The debug logs imply value is non-empty so I guess the secret has expired.

bgavrilMS (approved):

PR #459, opened 2023-10-06
Fix ADFS token caching

chlowell:
#451 assumed "**B**earer" is always the `token_type` of bearer tokens. It turns out some hosts return "**b**earer" instead, so that assumption broke token caching in some cases. I went with the easiest solution here--compare token types case-insensitively--but if that isn't safe in general I suppose we'll have to special-case [Bb]earer.


chlowell:
@bgavril integration tests fail due to an invalid value of LAB_APP_CLIENT_SECRET (one of this repo's Actions secrets). The debug logs imply value is non-empty so I guess the secret has expired.

bgavrilMS (approved):

bgavrilMS:
I'm going to merge this, as the failing test is not relevant for this fix.

PR #466, opened 2023-12-07
Clarify cache usage in client_creds sample

bgavrilMS:


apps/tests/devapps/client_secret_sample.go
```diff
@@ -11,23 +11,28 @@ import (
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
 )
 
+// Keep the ConfidentialClient application object around, because it maintains a token cache
+var _app *confidential.Client
```
rayluo, line 15: Shouldn't the other `client_certificate_sample.go` file also pick up such a change?

apps/tests/devapps/main.go
```diff
@@ -26,8 +26,20 @@ func main() {
...
 	} else if exampleType == "5" {
+		// This sample does not use a serialized cache - it relies on in-memory cache by reusing the app object
+		// This works great for app tokens, because there is only 1 token per resource, per tenant and most
+		// developers only require 1-2 tokens.
+		acquireTokenClientSecret()
+
+		// this time the token comes from the cache!
 		acquireTokenClientSecret()
 	} else if exampleType == "6" {
+		// This samples	uses a serialized cache in a file. This is for demonstration purposes only of the caching interface.
```
rayluo, line 37: This comes as a surprise. Was it because the client_cert_sample contains `confidential.WithCache(cacheAccessor)`, while the client_secret_sample does not? Personally, I would prefer the two samples behave the same, i.e., both use in-memory cache or both use serialized cache. Both can contain same comments to elaborate when to use serialized cache.

apps/tests/devapps/client_secret_sample.go
```diff
@@ -11,23 +11,28 @@ import (
...
-	}
-	app, err := confidential.New(config.Authority, config.ClientID, cred, confidential.WithCache(cacheAccessor))
-	if err != nil {
-		log.Fatal(err)
-	}
-	result, err := app.AcquireTokenSilent(context.Background(), config.Scopes)
-	if err != nil {
-		result, err = app.AcquireTokenByCredential(context.Background(), config.Scopes)
+
+	if _app == nil {
```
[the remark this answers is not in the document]
bgavrilMS: Ok, I had to make the _config as global variable too, but I think that's fine.

apps/tests/devapps/client_secret_sample.go
bgavrilMS: Done.

PR #478, opened 2024-04-09
Always close the IMDS response body

qmuntal:
The [http.Response](https://pkg.go.dev/net/http#Response) body should always be closed, even for non-200 status codes. Not doing so leaks the system resources associated to the body.

While here, associate the context passed in `detectRegion` to the http request. It was previously not being used.


epelc:
Can this get merged? It's a good fix. Adds context usage and fixes leaks.

bgavrilMS (approved):

PR #485, opened 2024-05-22
Enhancement -  Add option to add a state to an auth code url request

tomneurobell:
I have added another option to the AuthCodeURL function in the confidential package to allow for conveniently adding a state parameter to the request, this avoids the need to have to modify the URL output manually and improves DevEX.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -343,7 +343,7 @@ func New(authority, clientID string, cred Credential, options ...Option) (Client
 
 // authCodeURLOptions contains options for AuthCodeURL
 type authCodeURLOptions struct {
-	claims, loginHint, tenantID, domainHint string
+	claims, loginHint, tenantID, domainHint, state string
 }
 
 // AuthCodeURLOption is implemented by options for AuthCodeURL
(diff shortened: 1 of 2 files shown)


tomneurobell:

apps/confidential/confidential.go
```diff
@@ -363,12 +363,36 @@ func (cca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string,
...
 	}
+	ap.State = o.state
 	ap.Claims = o.claims
 	ap.LoginHint = o.loginHint
 	ap.DomainHint = o.domainHint
 	return cca.base.AuthCodeURL(ctx, clientID, redirectURI, scopes, ap)
 }
 
+// WithState adds a user-generated state to the request.
+func WithState(state string) interface {
```
wayneforrest, line 374: should this not be `any` or `interface{}`, allowing for any kind of object to be attached.

tomneurobell: No, the `ap.State` type was already a string type and this PR merely adds a way to set that value so any change to the underlying types is outside the scope of this change.
Additionally, `ap.State` needs to be added as a query param and therefore URL encoded.

As a workaround you can just stringify whatever type you wish to add and pass it that way.

apps/confidential/confidential.go
dcormier, line 1: It won't let me comment on the specific line because it hasn't changed, but add `, [WithState]` to the doc comment on the `AuthCodeURL` method (line 357).

tomneurobell: Nice catch, I have pushed the update to add the comment!
Thanks, @dcormier!

apps/confidential/confidential.go
```diff
@@ -363,12 +363,36 @@ func (cca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string,
...
 		return "", err
 	}
+	ap.State = o.state
 	ap.Claims = o.claims
 	ap.LoginHint = o.loginHint
 	ap.DomainHint = o.domainHint
 	return cca.base.AuthCodeURL(ctx, clientID, redirectURI, scopes, ap)
 }
 
+// WithState adds a user-generated state to the request.
```
bgavrilMS, line 373:
```suggestion
// WithState adds a user-generated state to the URL.
```

PR #490, opened 2024-06-12
Add option to set success and error pages for when calling AcquireInt…

wayneforrest:
Add option to set success and error web pages for when using `AcquireTokenInteractive` 

For example :

     var successPage = []byte(`<!DOCTYPE html>
         <html>
              <body>A Success Page</body>
         </html>`)
         
     var errorPage = []byte(`<!DOCTYPE html>
         <html>
              <body>A Error Page</body>
         </html>`)
         
        
	result, err = client.AcquireTokenInteractive(context.TODO(), scopes,
		public.WithSuccessPage(successPage),
		public.WithErrorPage(errorPage),
	)
	if err != nil {
		log.Fatalf("Failed to acquire token interactively: %v", err)
	}

### DIFF
--- a/apps/internal/local/server.go
+++ b/apps/internal/local/server.go
@@ -28,7 +28,7 @@ var okPage = []byte(`
 </html>
 `)
 
-const failPage = `
+var failPage = []byte(`
 <!DOCTYPE html>
 <html>
 <head>
@@ -40,7 +40,7 @@ const failPage = `
 	<p>Error details: error %s error_description: %s</p>
 </body>
 </html>
-`
+`)
 
 // Result is the result from the redirect.
 type Result struct {
@@ -60,7 +60,7 @@ type Server struct {
 }
 
 // New creates a local HTTP server and starts it.
-func New(reqState string, port int) (*Server, error) {
+func New(reqState string, port int, successPage []byte, errorPage []byte) (*Server, error) {
 	var l net.Listener
 	var err error
 	var portStr string
@@ -82,6 +82,14 @@ func New(reqState string, port int) (*Server, error) {
 	}
 	if err != nil {
 		return nil, err
+	}
+
+	if len(successPage) > 0 {
+		okPage = successPage
+	}
+
+	if len(errorPage) > 0 {
+		failPage = errorPage
 	}
 
 	serv := &Server{
@@ -145,7 +153,7 @@ func (s *Server) handler(w http.ResponseWriter, r *http.Request) {
 		desc := html.EscapeString(q.Get("error_description"))
 		// Note: It is a little weird we handle some errors by not going to the failPage. If they all should,
 		// change this to s.error() and make s.error() write the failPage instead of an error code.
-		_, _ = w.Write([]byte(fmt.Sprintf(failPage, headerErr, desc)))
+		_, _ = w.Write([]byte(fmt.Sprintf(string(failPage), headerErr, desc)))
 		s.putResult(Result{Err: fmt.Errorf(desc)})
 		return
 	}
--- a/apps/public/public.go
+++ b/apps/public/public.go
@@ -522,11 +522,57 @@ type interactiveAuthOptions struct {
 	claims, domainHint, loginHint, redirectURI, tenantID string
 	openURL                                              func(url string) error
 	authnScheme                                          AuthenticationScheme
+	successPage                                          []byte
+	errorPage                                            []byte
 }
 
 // AcquireInteractiveOption is implemented by options for AcquireTokenInteractive
 type AcquireInteractiveOption interface {
 	acquireInteractiveOption()
+}
+
+func WithSuccessPage(successPage []byte) interface {
+	AcquireInteractiveOption
+	options.CallOption
+} {
+	return struct {
+		AcquireInteractiveOption
+		options.CallOption
+	}{
+		CallOption: options.NewCallOption(
+			func(a any) error {
+				switch t := a.(type) {
+				case *interactiveAuthOptions:
+					t.successPage = successPage
+				default:
+					return fmt.Errorf("unexpected options type %T", a)
+				}
+				return nil
+			},
+		),
+	}
+}
+
+func WithErrorPage(errorPage []byte) interface {
+	AcquireInteractiveOption
+	options.CallOption
+} {
+	return struct {
+		AcquireInteractiveOption
+		options.CallOption
+	}{
+		CallOption: options.NewCallOption(
+			func(a any) error {
+				switch t := a.(type) {
+				case *interactiveAuthOptions:
+					t.errorPage = errorPage
+				default:
+					return fmt.Errorf("unexpected options type %T", a)
+				}
+				return nil
+			},
+		),
+	}
 }
 
 // WithLoginHint pre-populates the login prompt with a username.
@@ -671,7 +717,7 @@ func (pca Client) AcquireTokenInteractive(ctx context.Context, scopes []string, 
 	if o.authnScheme != nil {
 		authParams.AuthnScheme = o.authnScheme
 	}
-	res, err := pca.browserLogin(ctx, redirectURL, authParams, o.openURL)
+	res, err := pca.browserLogin(ctx, redirectURL, authParams, o.openURL, o.successPage, o.successPage)
 	if err != nil {
 		return AuthResult{}, err
 	}
@@ -709,13 +755,13 @@ func parsePort(u *url.URL) (int, error) {
 }
 
 // browserLogin calls openURL and waits for a user to log in
-func (pca Client) browserLogin(ctx context.Context, redirectURI *url.URL, params authority.AuthParams, openURL func(string) error) (interactiveAuthResult, error) {
+func (pca Client) browserLogin(ctx context.Context, redirectURI *url.URL, params authority.AuthParams, openURL func(string) error, successPage []byte, errorPage []byte) (interactiveAuthResult, error) {
 	// start local redirect server so login can call us back
 	port, err := parsePort(redirectURI)
 	if err != nil {
 		return interactiveAuthResult{}, err
 	}
-	srv, err := local.New(params.State, port)
+	srv, err := local.New(params.State, port, successPage, errorPage)
 	if err != nil {
 		return interactiveAuthResult{}, err
 	}


wayneforrest:
@microsoft-github-policy-service agree [company=“default"]

I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.

wayneforrest:

apps/public/public.go
```diff
@@ -522,13 +522,59 @@ type interactiveAuthOptions struct {
...
+	successPage                                          []byte
+	errorPage                                            []byte
 }
 
 // AcquireInteractiveOption is implemented by options for AcquireTokenInteractive
 type AcquireInteractiveOption interface {
 	acquireInteractiveOption()
 }
 
+func WithSuccessPage(successPage []byte) interface {
```
bgavrilMS, line 534: From the public API standpoint, I think we'd want an API more like `WithSystemBrowserOptions(succesPage, errorPage)`. This is because we may introduce support for broker, these options won't make sense .

CC @localden

localden: Agree - this needs to be mapped to the API concepts from other MSALs. Additionally, would be good to get @rayluo's input here - we _likely_ need to find a way to sanitize inputs here so that we don't end up in a scenario where someone can pass in malicious code into the rendered page.

bgavrilMS (approved):

apps/internal/local/server.go
```diff
@@ -145,7 +153,7 @@ func (s *Server) handler(w http.ResponseWriter, r *http.Request) {
 		desc := html.EscapeString(q.Get("error_description"))
 		// Note: It is a little weird we handle some errors by not going to the failPage. If they all should,
 		// change this to s.error() and make s.error() write the failPage instead of an error code.
-		_, _ = w.Write([]byte(fmt.Sprintf(failPage, headerErr, desc)))
+		_, _ = w.Write([]byte(fmt.Sprintf(string(failPage), headerErr, desc)))
```
chlowell, line 156: This tries to interpolate a couple strings into `failPage`, which may not contain format verbs. I guess someone could learn that by reading this code, or we could document it somewhere, but it raises the question: is this interpolation part of the API contract?

apps/internal/local/server.go
```diff
@@ -84,6 +84,14 @@ func New(reqState string, port int) (*Server, error) {
 		return nil, err
 	}
 
+	if len(successPage) > 0 {
+		okPage = successPage
+	}
+
+	if len(errorPage) > 0 {
+		failPage = errorPage
+	}
```
chlowell, line 93: `WithSystemBrowserOptions` is a per-call option but `failPage` and `okPage` are package-level variables. Setting them here therefore affects every `AcquireTokenInteractive` call, for all clients.

chlowell (changes requested):

apps/public/public.go
localden: @bgavrilMS @chlowell - just a sanity check on my part, do we need some sanitization logic here, similar to what we've updated in Python?

chlowell: I can imagine a scenario in which an attacker steals a token by modifying an app's success page but whether that's something MSAL must (try to) prevent depends on the threat model.

apps/internal/local/server.go
wayneforrest: @chlowell Thanks for checking this. I have fixed this, by removing the string interpolation, and also moved the scope of the success and error page to the lifetime of the server object.

apps/internal/local/server.go
wayneforrest: @chlowell I fixed this one, by moving the scope to the server struct;

apps/public/public.go
wayneforrest: Hi @chlowell , Can you please explain what to check for here?  (_my setup is not ideal yet, and I was not able to inspect the http headers to see if the token is present, when serving the error / success page_)

wayneforrest: Is there a preferred / accepted library for sanitising HTML ?

I found a few libraries ([this one ](https://pkg.go.dev/github.com/microcosm-cc/bluemonday#section-readme)stands out a bit more than others) that could help with sanitising HTML.

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/AzureAD/microsoft-authentication-library-for-go
+module github.com/wayneforrest/microsoft-authentication-library-for-go
```
rayluo, line 1: Was this change intended?

apps/public/public.go
rayluo: > I can imagine a scenario in which an attacker steals a token by modifying an app's success page but whether that's something MSAL must (try to) prevent depends on the threat model.

I thought I wrote a reply for this message but apparently it did not go through. What I tried to say was that, a success page typically contains only a hardcoded message such as "sign-in succeeded, you can close this tab now" so it does not have XSS risk. It is the error page that acts as a template containing variable(s) for error content, which we need sanitising before rendering.

go.mod
wayneforrest: ooh .. I see what happened here 😂 I was forking the repo. I will fix this.

localden:
@bgavrilMS @chlowell @rayluo overall, the PR looks good to me, but to confirm, we're OK with arbitrary `byte[]` payloads being passed for error and success pages?

wayneforrest:
> @bgavrilMS @chlowell @rayluo overall, the PR looks good to me, but to confirm, we're OK with arbitrary `byte[]` payloads being passed for error and success pages?

Are there any more changes that you would like me to add to this PR?

cc @chlowell you still have a change request on this PR. Has this been resolved for you?

localden (approved):

chlowell:
There are still a few things to address here:
1. we need tests
1. we should copy any `[]byte` from the application to make its content immutable
2. `Server` interpolates error details into `failPage`: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/004301ceccb1206100a19cae1a6a4e4e37be8b20/apps/internal/local/server.go#L148

A few options for addressing item 3:
1. don't interpolate when the application specified an error page
1. document the formatting, maybe simplify by interpolating only one string
1. consider a different design such as taking callbacks returning `[]byte` instead of plain old `[]byte`; then we can pass optional stuff like error details to the application and it can decide whether and how to present them

Option 2 looks good to me; @bgavrilMS @localden @rayluo do you have a preference?

localden:
> There are still a few things to address here:
> 
> 1. we need tests
> 2. we should copy any `[]byte` from the application to make its content immutable
> 3. `Server` interpolates error details into `failPage`: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/004301ceccb1206100a19cae1a6a4e4e37be8b20/apps/internal/local/server.go#L148
> 
> A few options for addressing item 3:
> 
> 1. don't interpolate when the application specified an error page
> 2. document the formatting, maybe simplify by interpolating only one string
> 3. consider a different design such as taking callbacks returning `[]byte` instead of plain old `[]byte`; then we can pass optional stuff like error details to the application and it can decide whether and how to present them
> 
> Option 2 looks good to me; @bgavrilMS @localden @rayluo do you have a preference?

Option 2 sounds reasonable to me - formatting to be documented and interpolation simplified. Thanks for reviewing @chlowell!

rayluo:
> 3\. `Server` interpolates error details into `failPage`: https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/004301ceccb1206100a19cae1a6a4e4e37be8b20/apps/internal/local/server.go#L148
> 
> A few options for addressing item 3: ...

What are those 3 options trying to address? I believe, regardless of whether a default `failPage` or an application-specified `s.optionErrorPage` is being used, [interpolation shall only accept escaped value (example here)](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/oauth2cli/authcode.py#L126).

chlowell:
> What are those 3 options trying to address? I believe, regardless of whether a default `failPage` or an application-specified `s.optionErrorPage` is being used, [interpolation shall only accept escaped value (example here)](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/oauth2cli/authcode.py#L126).

I'm not worrying about escaping--maybe I should--but developer confusion and unexpected results. If we (try to) augment an application-provided error page with an error message, we should document that behavior or present the message via some API so developers expect the additional content.

rayluo:
> > What are those 3 options trying to address? I believe, regardless of whether a default `failPage` or an application-specified `s.optionErrorPage` is being used, [interpolation shall only accept escaped value (example here)](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/oauth2cli/authcode.py#L126).
> 
> I'm not worrying about escaping--maybe I should--but developer confusion and unexpected results. If we (try to) augment an application-provided error page with an error message, we should document that behavior or present the message via some API so developers expect the additional content.

Great. Now we agree to escape the values that will be rendered on the error page. :-)

"Developer confusion and unexpected results" can be avoided by API documentation such as "the error page is treated as a template and its variable $Foo will be interpolated by blah blah". That is an intuitive concept and easily understandable.

wayneforrest:
> > > What are those 3 options trying to address? I believe, regardless of whether a default `failPage` or an application-specified `s.optionErrorPage` is being used, [interpolation shall only accept escaped value (example here)](https://github.com/AzureAD/microsoft-authentication-library-for-python/blob/dev/msal/oauth2cli/authcode.py#L126).
> > 
> > 
> > I'm not worrying about escaping--maybe I should--but developer confusion and unexpected results. If we (try to) augment an application-provided error page with an error message, we should document that behavior or present the message via some API so developers expect the additional content.
> 
> Great. Now we agree to escape the values that will be rendered on the error page. :-)
> 
> "Developer confusion and unexpected results" can be avoided by API documentation such as "the error page is treated as a template and its variable $Foo will be interpolated by blah blah". That is an intuitive concept and easily understandable.

Option 1 already solves my "personal" needs, for displaying end user friendly pages, where technical error codes / descriptions are not understood (unless you are a developer, useful for debugging). 

Also are the messages multi lingual ? If not then Option 2 might not be an ideal solution for all?

Option 3 would allow for error code interpretation; allowing one to suggest a course of action to end users rather than error messages whose audience is more developer orientated. I am not sure how to control the "escaping" as the control falls away (user application space). Perhaps on the return path to the SDK before the server renders the page, the page can  be sanitised using something like _bluemonday_ : https://pkg.go.dev/github.com/microcosm-cc/bluemonday ?

wayneforrest:
Hi @chlowell thanks for reviewing. Regarding the outstanding list of work:

1: I have added some [tests](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/490/files#diff-fb12097c66f2795fe9dc8738aa398098c86ab8397c11140f603d5306083e710eR162), can you please review. (it's a bit hard to visualise the changes in GitHub)
2: I have also made a "copy" of success/error pages [immutable](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/490/files#diff-c2412326515057bec0db0c5d2e9a23ebc4a9cdfd194d9675ca515e59f2b76ccbR546) once supplied to the SDK.
3: Interpolation - not done - I made some comments on this above.

rayluo:
> Option 1 already solves my "personal" needs, for displaying end user friendly pages, where technical error codes / descriptions are not understood (unless you are a developer, useful for debugging).
>
> ... not sure how to control the "escaping" as the control falls away (user application space) ...
>
> I have also made a "copy" of success/error pages [immutable](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/490/files#diff-c2412326515057bec0db0c5d2e9a23ebc4a9cdfd194d9675ca515e59f2b76ccbR546) once supplied to the SDK.

I'm not sure what the benefit is to make "a copy of success/error pages that is immutable". Is that an attack vector unique to Go?

But to be clear, my "escape/sanitize" suggestion is NOT for the app-provided success/error page. I was referring to the interpolation of error code and error description that already exist and is not fixed by this PR.

Technically, you could bypass the "not sure how to escape" issue, by removing that interpolation (even though it is NOT for app-provided error page). But that would risk a regression in the future. I would suggest we escape the error content properly this time.

wayneforrest:
> I'm not sure what the benefit is to make "a copy of success/error pages that is immutable". Is that an attack vector unique to Go?

I am not sure what the reasoning behind this is, it was suggested by @chlowell to make the `[]byte` immutable. I also checked how golang passes values, and it's by default "value" / "copies" - What I did makes no sense to me now, it's already passing a copy. I think I should revert what I did, but will wait to hear more.

chlowell:
What you reverted was correct--we should copy all slices provided by an application because slices are reference types. A slice value is a pointer to memory, not the content of that memory. We don't want to share a pointer with the application because the application can modify the data pointed to at any time ([for example](https://play.golang.com/p/Fr0oAeHL8ia)).

wayneforrest:
> What you reverted was correct--we should copy all slices provided by an application because slices are reference types. A slice value is a pointer to memory, not the content of that memory. We don't want to share a pointer with the application because the application can modify the data pointed to at any time ([for example](https://play.golang.com/p/Fr0oAeHL8ia)).

Thanks @chlowell ... And for the example! I can also confirm (dang GPT), I checked my [Go PL Book](https://www.gopl.io) page 51, and this commit on the [Go Spec](https://github.com/golang/go/commit/b34f0551387fcf043d65cd7d96a0214956578f94#diff-2ed8cb1b7a57009287e61a0cae50c7f1c50fd9eea78ac2a808a7b8c1cd560cfcR840). (a slice has a pointer to a backing array)

apps/internal/local/server.go
```diff
@@ -167,7 +175,11 @@ func (s *Server) handler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	_, _ = w.Write(okPage)
+	if len(s.optionSuccessPage) > 0 {
```
chlowell, line 178: You can refactor a little to remove this condition:
```go
func New(reqState string, port int, successPage []byte, errorPage []byte) (*Server, error) {
	if len(successPage) == 0 {
		successPage = okPage
	}
	serv := &Server{
		// len(successPage) != 0
		successPage: successPage,
	}
```
With that, you can unconditionally `w.Write(s.successPage)`

wayneforrest:
I have made some more changes, which might not be what's wanted yet.

1: I have used html templates, which is injection safe, and removed the Escape from earlier. https://pkg.go.dev/html/template
2: I have refactored the code as suggested by @chlowell (it makes it allot cleaner when checking lengths of optional pages)
3: I have added tests, to endure the parsing works for when the template variables might and might not be present.
    also removed the redundant `nil` initialisation of the error and success pages
4: I have altered default fail-page also to be a html template (using .Code and .Err template variables)
5: I have added doc comment to the public facing method `WithSystemBrowserOptions` - wording needs a review.

apps/internal/local/server.go
```diff
@@ -145,7 +149,11 @@ func (s *Server) handler(w http.ResponseWriter, r *http.Request) {
 		desc := html.EscapeString(q.Get("error_description"))
 		// Note: It is a little weird we handle some errors by not going to the failPage. If they all should,
 		// change this to s.error() and make s.error() write the failPage instead of an error code.
-		_, _ = w.Write([]byte(fmt.Sprintf(failPage, headerErr, desc)))
+		if len(s.optionErrorPage) > 0 {
+			_, _ = w.Write(s.optionErrorPage)
+		} else {
+			_, _ = w.Write([]byte(fmt.Sprintf(failPage, headerErr, desc)))
```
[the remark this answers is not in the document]
rayluo: @chlowell 
AFAIK, there was no official specs on this due to the "custom error page" feature was not mandatory across all MSALs. But there was a CVE mentioning a potential XSS risk. In short, an implementation shall escape the error code and error description so that a malicious `error_code=<script>alert("this code snippet was executed")</script>` input shall not be executed.

@wayneforrest , I am not sure whether your "4: I have altered default fail-page also to be a html template (using .Code and .Err template variables)" can address that XSS risk.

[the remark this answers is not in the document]
wayneforrest: Hi @rayluo I did a quick test by updating the tests as follows:

`url.Values{"error": []string{"<script>alert('this code snippet was executed')</script>"}, "error_description": []string{"error_description"}},`

This resulted in the following fail page content: 

`"\n<!DOCTYPE html>\n<html>\n<head>\n    <meta charset=\"utf-8\" />\n    <title>Authentication Failed</title>\n</head>\n<body>\n\t<p>Authentication failed. You can return to the application. Feel free to close this browser tab.</p>\n\t<p>Error details: error &lt;script&gt;alert(&#39;this code snippet was executed&#39;)&lt;/script&gt;, error description: error_description</p>\n</body>\n</html>\n"`

The template variables got escaped. This page here also has examples: https://pkg.go.dev/html/template

I also made a small example here: https://go.dev/play/p/yfuM0RodQCh

bgavrilMS (approved):

apps/internal/local/server.go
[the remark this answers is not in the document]
rayluo: @wayneforrest , the escaping effect is what we want, so, if it is somehow already escaping properly, we don't have to fix anything. That being said, I would encourage you to find out where that escaping effect comes from, and then add comment to leave a hint to future maintainer that the escaping is happening. Currently, in this PR, I only saw you removed a pre-existing `html.EscapeString` and I did not dive deep to figure out why you still observed the escaping effect.

Regardless, can you also add a test case to make sure the escaping happens?

apps/internal/local/server.go
```diff
@@ -142,11 +162,19 @@ func (s *Server) handler(w http.ResponseWriter, r *http.Request) {
 
 	headerErr := q.Get("error")
 	if headerErr != "" {
-		desc := html.EscapeString(q.Get("error_description"))
```
rayluo, line 145: Since this line was removed, besides I am not seeing similar escaping on `error`, can we have a unit test case to ensure the escaping happens?

apps/internal/local/server.go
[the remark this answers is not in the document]
wayneforrest: @rayluo, I have added some comments in the code - most importantly is the import statement, as this can easily be altered incorrectly. 

I have also added a test, that should also guard against regression.

apps/internal/local/server_test.go
```diff
@@ -111,6 +112,15 @@ func TestServer(t *testing.T) {
 			statusCode:   200,
 			testTemplate: true,
 		},
+		{
+			desc:              "Error: Query Values missing 'state' key, using default fail error page - XSS test",
+			reqState:          "state",
+			port:              0,
+			q:                 url.Values{"error": []string{"<script>alert('this code snippet was executed')</script>"}, "error_description": []string{"error_description"}},
```
rayluo, line 119: Thanks. Please also add the similar scripts to the "error_description" and have it tested, if it has not been covered by any existing test case.

wayneforrest: Hi @rayluo , I have added some more test cases. (please see my last two commits).
1: Updating more test case for default fail page.
2: Added two more test cases for the SDK _user_ provided option error page (one for each html template field).

rayluo (approved):
Thanks for the recent improvements on the XSS protection test cases. That part looks good to me now.

Per offline discussion with @chlowell , he is still in the process of reviewing whether this new API can be utilized in Azure SDK. Let's wait for him to provide further comment and eventually an approval.

chlowell:
I talked this over with SDK architects today and they pointed out that `text/template` uses reflection features that prevent the linker from identifying dead code ([see this presentation for an explanation](https://youtu.be/EkG177eRcco?si=ddGtx0CLs1dBd7Ed)). Using that package in MSAL would therefore increase the binary size of dependent applications unnecessarily.

Simply continuing to interpolate details into the error page with `fmt.Sprintf` as I've suggested elsewhere doesn't work well either because `fmt` methods output warnings when arguments don't match verbs:
```go
fmt.Sprintf("%s") // %!s(MISSING)
fmt.Sprintf("", "foo") // %!(EXTRA string=foo)
```

So, I'm thinking the best solution is to take a callback for customizing the error page: give applications a struct with error details and let them decide how to make HTML from it. We'd export the [Result type](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/4961ef926304f2aa240b1751c4b87f83145356d2/apps/internal/local/server.go#L46-L51) (perhaps as `public.AcquireTokenInteractiveResult`?), and instead of a `[]byte` for the error page, `WithSystemBrowserOptions` would take a `func(ResultTypeNameTBD) ([]byte, error)`. Thoughts?

chlowell (changes requested):
`html/template` unnecessarily increases application size 😞

wayneforrest:
> `html/template` unnecessarily increases application size 😞

Hi @chlowell, this comes as a big surprise, but also very interesting. Thank you for sharing the video.

I did a very rough test build, using this example as the SDK application:

```
package main

+import (
	"fmt"

	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
)

+func main() {
	_, _ = public.New("client_id", public.WithAuthority("https://login.microsoftonline.com/your_tenant"))
	fmt.Println("vim-go")
}
```

And got these results: `go build ./...`
4601314 main branch - #e331a85899
4867506 using html/template / reflection
-rwxr-xr-x   1 wayneforrest  staff   4.4M 21 Sep 22:32 microsoft-authentication-library-for-go
-rwxr-xr-x   1 wayneforrest  staff   4.6M 21 Sep 22:35 microsoft-authentication-library-for-go

> diff = 260KB

With the callback solution suggestion, would the XSS protection be the responsibility of the SDK application?

An alternative would be to go back to  using `html.EscapeString` and do manual interpolation.
Here is an example application / approach I was thinking of:
https://go.dev/play/p/wnmCQAi0WYC

Please let me know your thoughts. I am happy to implement either solution.

rayluo:
> With the callback solution suggestion, would the XSS protection be the responsibility of the SDK application?

The stakes are high. SDK shall achieve that.

wayneforrest:
@rayluo / @chlowell  Please see my last two commits:
1: I removed the html template, and reverted to using `html.EscapeString` as it was before. 
2: Then also keeping the `concept` of a template, but instead doing the parsing / interpolation manually,
     circumventing the use of the `html/template` package.

apps/internal/local/server.go
```diff
@@ -142,11 +164,23 @@ func (s *Server) handler(w http.ResponseWriter, r *http.Request) {
...
 		// Note: It is a little weird we handle some errors by not going to the failPage. If they all should,
 		// change this to s.error() and make s.error() write the failPage instead of an error code.
-		_, _ = w.Write([]byte(fmt.Sprintf(failPage, headerErr, desc)))
-		s.putResult(Result{Err: fmt.Errorf(desc)})
+
+		errDesc := q.Get("error_description")
+
+		if bytes.Contains(s.errorPage, code) {
+			s.errorPage = bytes.Replace(s.errorPage, code, []byte(html.EscapeString(headerErr)), 1) // provides XSS protection
+		}
```
chlowell, line 174: 1. Contains is just wasting time because Replace is a no-op when the slice doesn't contain the target
1. Replace returns a new slice you assign to `s.errorPage`. This makes the output of the each Replace the input to the next Replace. A Replaced slice has at least one fewer instance of the template argument. So, the net effect is that each argument replacement is permanent and once all arguments have been Replaced, all subsequent error pages are the same
1. you're replacing only the first instance of the argument. Why not replace all of them?

```suggestion
		errorPage := bytes.Replace(s.errorPage, code, []byte(html.EscapeString(headerErr)), -1) // provides XSS protection
```

apps/internal/local/server_test.go
```diff
@@ -129,6 +217,61 @@ func TestServer(t *testing.T) {
...
+		if len(test.successPage) > 0 {
+			if !bytes.Equal(content, test.successPage) {
+				t.Errorf("TestServer(%s): -want/+got:\ntest option error page", test.desc)
+			}
+			continue
+		}
+
+		if test.testTemplate {
+			if test.testErrCodeXSS || test.testErrDescriptionXSS {
+				if !strings.Contains(string(content), "&lt;script&gt;alert(&#39;this code snippet was executed&#39;)&lt;/script&gt;") {
```
chlowell, line 229: It would be much simpler to add the (hardcoded) expected output to `test`. That would let you write e.g. `if !bytes.Equal(test.expected, content)` instead of all these conditions

apps/public/public.go
```diff
@@ -522,13 +522,44 @@ type interactiveAuthOptions struct {
...
 
 // AcquireInteractiveOption is implemented by options for AcquireTokenInteractive
 type AcquireInteractiveOption interface {
 	acquireInteractiveOption()
 }
 
+// WithSystemBrowserOptions sets the optional success and error pages.
+// The error page supports two optional html template variables {{.Code}} and {{.Err}},
+// which will be replaced with the corresponding error code, and descriptions.
+func WithSystemBrowserOptions(successPage []byte, errorPage []byte) interface {
```
chlowell, line 537: Just FYI, you can use type list syntax for function parameters
```suggestion
func WithSystemBrowserOptions(successPage, errorPage []byte) interface {
```

chlowell (changes requested):

apps/internal/local/server.go
wayneforrest: Hi @chlowell thanks for the feedback. I will address these a bit later (maybe tonight/tomorrow), but will get to it.

wayneforrest: Hi @chlowell I have made some more changes, please see my last 3 commits:

1: using Replace all
2: updated tests, using an `expected` value defined in the test definition
3: using type list syntax

chlowell: You haven't fixed the bug: https://play.golang.com/p/0I-sHtF-n83. Please also simplify the test. If each test case carries its expected content, the test body doesn't need a pile of conditional blocks to validate the actual content. It can simply check whether the actual content equals the expected content

wayneforrest: Hi @chlowell, can you please check again:
1: I have updated tests, removing the conditional logic. It's much simpler.
2: Updated the error page handling. I am no longer storing updated `errorspage` in the server struct.

wayneforrest: @chlowell, are you free to take some time to review again please, and test ?

Issue #407, opened 2023-04-12
[Feature Request] add State option to AuthCodeURL


karelbilek:
**Is your feature request related to a problem? Please describe.**
AuthCodeURL doesn't have State option

**Describe the solution you'd like**
Add an option to pass State to pass to AD oauth

**Describe alternatives you've considered**
just adding "&state={state}" to the URL that's generated

**Additional context**
https://auth0.com/docs/secure/attack-protection/state-parameters

Issue #227, opened 2021-06-19
[Feature Request] 


samueldominguez:
**Is your feature request related to a problem? Please describe.**
When using `CreateAuthCodeURL` I can't set a challenge or state, so I have to manually add it to the URL.

**Describe the solution you'd like**
Allow `CreateAuthCodeURL` to set a challenge.

**Describe alternatives you've considered**
Manipulating the URL after it has been generated by `CreateAuthCodeURL` but the functionality to generate the code verifier and so on is already present in the package so why not do this automatically.

**Additional context**
N/A.

simonoff:
As I see such library not supported by Microsoft? Why then recommend to use MSAL only then in docs?

zc07:
It has been a year and a half since this issue was being created. I have the same request for allowing custom `session_state` and `code_challenge`.

* **Session State**: In the current implementation, we can only get a random `session_state` from AAD. Allowing custom state in `GenerateAuthCodeURL()` provides feature parity to other MSAL libraries such as MSAL.js.

* **Code Challenge**: If I understand correctly, MSAL for Golang does support PKCE when using the `AcquireTokenBy...()` methods, which completes the second half of the workflow. But there is no documentation/example/support on the first half. 
    * Using CodeLogin as an example, the challenge must be set when generating the auth code URL, then being verified in `AcquireTokenByAuthCode()`. The first part is missing in this library.

These should be essential for something to be using in production.

PR #491, opened 2024-06-20
Support passing a user-defined state value to `AuthCodeURL` on public clients

dcormier:
Related to #485, #227, and #407.

### DIFF
--- a/apps/public/public.go
+++ b/apps/public/public.go
@@ -142,7 +142,7 @@ func New(clientID string, options ...Option) (Client, error) {
 
 // authCodeURLOptions contains options for AuthCodeURL
 type authCodeURLOptions struct {
-	claims, loginHint, tenantID, domainHint string
+	claims, loginHint, tenantID, domainHint, state string
 }
 
 // AuthCodeURLOption is implemented by options for AuthCodeURL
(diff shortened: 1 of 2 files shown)


bgavrilMS:
Why do you need to pass in your own state on public client? MSAL should perform both legs (authorization call and token call) on public client. 

I can understand this scenario for confidential client, where you need MSAL only to generate the auth code URL, redirect the user there and then MSAL calls AcquireTokenByAuthCode

dcormier:
Maybe it isn't needed. I'm using the confidential client and needed it. I started working on it before seeing #485 was sitting there. I had done both sides since the functionality appeared similar, so I figured I might as well submit it.

rayluo:
> Why do you need to pass in your own state on public client? MSAL should perform both legs (authorization call and token call) on public client.
> 
> I can understand this scenario for confidential client, where you need MSAL only to generate the auth code URL, redirect the user there and then MSAL calls AcquireTokenByAuthCode

Agreed with what @bgavrilMS said.

Also want to add that some MSALs historically have the [generate auth url function even in public client, and it looks like MSAL Go does](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/491/files#diff-c2412326515057bec0db0c5d2e9a23ebc4a9cdfd194d9675ca515e59f2b76ccbR156). So, technically we could have added state support for that legacy function. But perhaps we shouldn't bother, and just deprecate that function and ask app developers to use public client's interactive function instead. If you would also agree, let's close this PR and close issues that are public-client-only as wontfix.

PR #498, opened 2024-08-13
Added a new variable in AuthRepsonse "IsFromCache"

4gust:
# Add `AuthResultMetadata` and `TokenSource` to `AuthResult`

## Description

This pull request introduces two main changes to the authentication result:

1. **`AuthResultMetadata` Struct**: A new struct, `AuthResultMetadata`, has been added to the `AuthResult` struct. This struct will be responsible for holding additional metadata related to authentication results.

2. **`TokenSource` Property**: A new property, `TokenSource`, has been introduced in the `AuthResultMetadata` struct. The `TokenSource` property indicates the origin of the token, with two possible values:
   - `IdentityProvider`: The token was obtained directly from an identity provider.
   - `Cached`: The token was retrieved from a cache.

## Changes

- **Added**: `AuthResultMetadata` class to the `AuthResult` struct.
- **Added**: `TokenSource` within `AuthResultMetadata` with possible values of `IdentityProvider` and `Cached`.
- **Updated**: `AuthResult` class to include the `AuthResultMetadata` instance.

## How to Test

1. Run the authentication process and inspect the `AuthResult` object.
2. Verify that the `AuthResultMetadata` object is correctly populated.
3. Check the `TokenSource` value to ensure it accurately reflects whether the token was retrieved from the `IdentityProvider` or from `Cached`.

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -83,13 +83,13 @@ type AcquireTokenOnBehalfOfParameters struct {
 // AuthResult contains the results of one token acquisition operation in PublicClientApplication
 // or ConfidentialClientApplication. For details see https://aka.ms/msal-net-authenticationresult
 type AuthResult struct {
-	Account            shared.Account
-	IDToken            accesstokens.IDToken
-	AccessToken        string
-	ExpiresOn          time.Time
-	GrantedScopes      []string
-	DeclinedScopes     []string
-	AuthResultMetadata AuthResultMetadata
+	Account        shared.Account
+	IDToken        accesstokens.IDToken
+	AccessToken    string
+	ExpiresOn      time.Time
+	GrantedScopes  []string
+	DeclinedScopes []string
+	Metadata       AuthResultMetadata
 }
 
 // AuthResultMetadata which contains meta data for the AuthResult
@@ -101,8 +101,9 @@ type TokenSource int
 
 // These are all the types of token flows.
 const (
-	IdentityProvider TokenSource = 0
-	Cache            TokenSource = 1
+	SourceUnknown    TokenSource = 0
+	IdentityProvider TokenSource = 1
+	Cache            TokenSource = 2
 )
 
 // AuthResultFromStorage creates an AuthResult from a storage token response (which is generated from the cache).
@@ -130,7 +131,7 @@ func AuthResultFromStorage(storageTokenResponse storage.TokenResponse) (AuthResu
 		ExpiresOn:      storageTokenResponse.AccessToken.ExpiresOn.T,
 		GrantedScopes:  grantedScopes,
 		DeclinedScopes: nil,
-		AuthResultMetadata: AuthResultMetadata{
+		Metadata: AuthResultMetadata{
 			TokenSource: Cache,
 		},
 	}, nil
@@ -147,7 +148,7 @@ func NewAuthResult(tokenResponse accesstokens.TokenResponse, account shared.Acco
 		AccessToken:   tokenResponse.AccessToken,
 		ExpiresOn:     tokenResponse.ExpiresOn.T,
 		GrantedScopes: tokenResponse.GrantedScopes.Slice,
-		AuthResultMetadata: AuthResultMetadata{
+		Metadata: AuthResultMetadata{
 			TokenSource: IdentityProvider,
 		},
 	}, nil
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -83,14 +83,27 @@ type AcquireTokenOnBehalfOfParameters struct {
 // AuthResult contains the results of one token acquisition operation in PublicClientApplication
 // or ConfidentialClientApplication. For details see https://aka.ms/msal-net-authenticationresult
 type AuthResult struct {
-	Account        shared.Account
-	IDToken        accesstokens.IDToken
-	AccessToken    string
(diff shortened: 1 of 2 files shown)


apps/internal/base/base.go
```diff
@@ -109,7 +110,7 @@ func AuthResultFromStorage(storageTokenResponse storage.TokenResponse) (AuthResu
 			return AuthResult{}, fmt.Errorf("problem decoding JWT token: %w", err)
 		}
 	}
-	return AuthResult{account, idToken, accessToken, storageTokenResponse.AccessToken.ExpiresOn.T, grantedScopes, nil}, nil
+	return AuthResult{account, idToken, accessToken, storageTokenResponse.AccessToken.ExpiresOn.T, grantedScopes, nil, true}, nil
```
chlowell, line 113: We should use named field syntax to prevent positioning bugs i.e. 
```go
return AuthResult{
    Account: account,
    ...
}
```

apps/internal/base/base.go
```diff
@@ -123,6 +124,7 @@ func NewAuthResult(tokenResponse accesstokens.TokenResponse, account shared.Acco
 		AccessToken:   tokenResponse.AccessToken,
 		ExpiresOn:     tokenResponse.ExpiresOn.T,
 		GrantedScopes: tokenResponse.GrantedScopes.Slice,
+		IsFromCache:   false,
```
chlowell, line 127: `false` is the zero value of `bool`
```suggestion
```

chlowell (commented):
Why would an application care whether a token came from the cache?

apps/internal/base/base.go
```diff
@@ -89,6 +89,7 @@ type AuthResult struct {
 	ExpiresOn      time.Time
 	GrantedScopes  []string
 	DeclinedScopes []string
+	IsFromCache    bool
```
bgavrilMS, line 92: Please update some unit tests to cover this. I am fairly sure there are tests that cover retrieval from cache / network.

bgavrilMS (changes requested):
Needs unit tests

apps/internal/base/base.go
4gust: Updated this!

apps/internal/base/base.go
4gust: Changed with the suggestion

apps/internal/base/base.go
4gust: I updated the test that checks for cache now also checks for `IsFromCache` to `true`
and likewise when we do a noncached check it checks for `IsFromCache` to `false`

AndyOHart (approved):
Looks good to me

rayluo:
> **New Field in `AuthResult`:**
> * **Field Name:** `IsFromCache`
> * **Type:** `bool`

Other MSALs added a new field `token_source` which value can be `identity_provider`, `cache` and `broker`.

apps/internal/base/base.go
rayluo, line 92: Other MSALs added a new field `token_source` which value can be `identity_provider`, `cache` and `broker`.

4gust: Thank you for the comment, I looked into msal .Net and msal java and updated the pullrequest with this
As, we don't have support for broker atm, I did not add it. 
When in Future we add that we can update the `TokenSource`

AndyOHart (approved):
Changes seem good

apps/internal/base/base.go
```diff
@@ -83,14 +83,28 @@ type AcquireTokenOnBehalfOfParameters struct {
...
-	ExpiresOn      time.Time
-	GrantedScopes  []string
-	DeclinedScopes []string
+	Account            shared.Account
+	IDToken            accesstokens.IDToken
+	AccessToken        string
+	ExpiresOn          time.Time
+	GrantedScopes      []string
+	DeclinedScopes     []string
+	AuthResultMetadata AuthResultMetadata
```
bgavrilMS, line 92: Should we just call it "Metadata"?  naming is hard :)

rayluo (approved):

apps/internal/base/base.go
chlowell: 👍 I think "AuthResultMetadata" is a reasonable name for the type ("Metadata" could also work depending on what else is in the package), however the field name should be "Metadata"

apps/internal/base/base.go
```diff
@@ -83,14 +83,28 @@ type AcquireTokenOnBehalfOfParameters struct {
...
+// AuthResultMetadata which contains meta data for the AuthResult
+type AuthResultMetadata struct {
+	TokenSource TokenSource
+}
+
+type TokenSource int
+
+// These are all the types of token flows.
+const (
+	IdentityProvider TokenSource = 0
```
chlowell, line 104: The constant with value 0 becomes the default because underlying type is `int`, whose zero value is 0:
```go
ar := AuthResult{}
ar.Metadata.TokenSource == IdentityProvider // true
```
This may be irrelevant because the field is always set explicitly, however if you want a different default and/or a "source unknown" constant, that should have value `0`.

4gust: Updated with the suggestion

apps/internal/base/base.go
4gust: Updated

bgavrilMS (approved):

PR #499, opened 2024-08-21
Implement Initial API Classes

AndyOHart:
* Adds initial public API based on the public API doc [here](https://microsofteur.sharepoint.com/:w:/t/DevExDublin/ERkeM2S2EglKibwGUP1ucigB-ySqCZQ3prVKwm6iSS-yEw?e=j3Q3nu)
* Adds new sample type for testing
* Creates a sample class to just run the public methods
* Creates a base test class that will be updated in another PR

### DIFF
--- a/apps/managedidentity/managedidentity.go
+++ b/apps/managedidentity/managedidentity.go
@@ -1,0 +1,137 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+/*
+Package managedidentity provides a client for retrieval of Managed Identity applications.
+The Managed Identity Client is used to acquire a token for managed identity assigned to
+an azure resource such as Azure function, app service, virtual machine, etc. to acquire a token
+without using credentials.
+*/
+package managedidentity
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/authority"
+)
+
+const (
+	// AzureArc represents the source to acquire token for managed identity is Azure Arc.
+	AzureArc = 0
+
+	// DefaultToIMDS indicates that the source is defaulted to IMDS since no environment variables are set.
+	DefaultToIMDS = 1
+)
+
+// Client is a client that provides access to Managed Identity token calls.
+type Client struct {
+	AuthParams      authority.AuthParams // DO NOT EVER MAKE THIS A POINTER! See "Note" in New(). also may remove from here
+	cacheAccessorMu *sync.RWMutex
+	// base                ops.HTTPClient
+	// managedIdentityType Type
+	// Token    			*oauth.Client
+	// pmanager manager 	// todo :  expose the manager from base.
+	// cacheAccessor   		cache.ExportReplace
+}
+
+// clientOptions are optional settings for New(). These options are set using various functions
+// returning Option calls.
+type clientOptions struct {
+	claims     string // bypasses cache, does nothing else
+	httpClient ops.HTTPClient
+	// disableInstanceDiscovery bool // always false
+	// clientId     string
+}
+
+type withClaimsOption struct{ Claims string }
+type withHTTPClientOption struct{ HttpClient ops.HTTPClient }
+
+// Option is an optional argument to New().
+type Option interface{ apply(*clientOptions) }
+type ClientOption interface{ ClientOption() }
+type AcquireTokenOption interface{ AcquireTokenOption() }
+
+// Source represents the managed identity sources supported.
+type Source int
+
+type systemAssignedValue string
+
+type ID interface {
+	value() string
+}
+
+func SystemAssigned() ID {
+	return systemAssignedValue("")
+}
+
+type ClientID string
+type ObjectID string
+type ResourceID string
+
+func (s systemAssignedValue) value() string { return string(s) }
+func (c ClientID) value() string            { return string(c) }
+func (o ObjectID) value() string            { return string(o) }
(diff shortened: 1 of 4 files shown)


apps/managedidentity/managedidentity.go
```diff
@@ -0,0 +1,137 @@
...
+	"sync"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/authority"
+)
+
+const (
+	// AzureArc represents the source to acquire token for managed identity is Azure Arc.
+	AzureArc = 0
```
bgavrilMS, line 24: Do the numbers matter? In languages like C# and Java, 0 is reserved for a value indicating "none" or "unknown" or smth.

apps/managedidentity/managedidentity_test.go
```diff
@@ -0,0 +1,46 @@
...
+func TestManagedIdentity(t *testing.T) {
+	client, err := fakeClient(SystemAssigned())
+
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = client.AcquireToken(context.Background(), "scope", WithClaims("claim"))
+
+	if err == nil {
```
bgavrilMS, line 29: What is the expected error?

docs/managedidentity_public_api.md
```diff
@@ -0,0 +1,206 @@
...
+```go
+// Acquires tokens from the configured managed identity on an azure resource.
+//
+// Resource: scopes application is requesting access to
+// Options: [WithClaims]
+func (client Client) AcquireToken(context context.Context, resource string, options ...AcquireTokenOption) (base.AuthResult, error) {
+    return base.AuthResult{}, nil
+}
+
+// Source represents the managed identity sources supported.
```
bgavrilMS, line 19: Pls move this to a separate section (maybe an appendix), as this API is really only meant for Azure SDK and for testing purposes.

apps/managedidentity/managedidentity.go
AndyOHart: The numbers just ensure that the corresponding value always matches that number,
previously we had something like this:

```
const (
    Planet = iota // 0
    Star               // 1
    Comet           // 2
    Galaxy           // 3
)
```

however Charles notified us that if you do it this way, and add another value to this const list at a later stage, it increments the numbers:

```
const (
    Planet = iota // 0
    Star               // 1
    Earth             // 2
    Comet           // 3
    Galaxy           // 4
)
```

So if we ever relied on the value of 2 for Comet previously, it will now be different and relate to Earth, so putting the numerical value here just ensures that those values stay the same

docs/managedidentity_public_api.md
```diff
@@ -0,0 +1,206 @@
...
+func WithClaims(claims string) AcquireTokenOption {
+    // implementation details
+}
+```
+
+## Error Handling
+
+Error handling in GO is different to what we used to in languages like Java or Swift.
+There is no concept of ‘exceptions’, instead we just return errors and immediately check if an error was returned and handle it there and then.  
+Basically, we will return client-side errors like so:
```
bgavrilMS, line 115:
```suggestion
The SDK will return client-side errors like so:
```

apps/managedidentity/managedidentity_test.go
AndyOHart: Right now this test class can be ignored as it is to be filled in as part of the next PR, but in general this will be an Error from the AcquireToken call, we will probably check specifics down the line as the tests are mapped out and designed. These initial tests were just to make sure the code was compiling ok

docs/managedidentity_public_api.md
```diff
@@ -0,0 +1,206 @@
...
+
+```go
+if err != nil {
+    return errors.New("Some Managed Identity Error here”) 
+}
+```
+
+This will be inside of any client methods that throw errors, using descriptive errors based on the .NET and Java Implementation. These errors will be propagated down the chain and handled when they are received
+
+For service side errors it works a little differently
```
bgavrilMS, line 125: The service errors need to include the MSI Source. This is a requirement from Azure SDK, but also to help the team with supportability.

docs/managedidentity_public_api.md
AndyOHart: Not sure I follow, should the public exposed data not have some documentation provided with it?

docs/managedidentity_public_api.md
AndyOHart: Yeah as you mentioned on the public API document, service errors will be returned and handled differently.
Is what is mentioned in the document not okay and you want it changed or are you just noting this here?

bgavrilMS: The service errors from MSI have to follow the same pattern as the service error from AAD. In addition to that, the Azure source (VM, Arc etc.) should be clear in the errors (e.g. in the message), to help troubleshooting.

AndyOHart: Yep, I see in Java they return a string which is getting the error response, description, correlation id etc, so we will do the same.
This is just copied from the public api doc, is there something more that you would like me to add here?

apps/managedidentity/managedidentity.go
chlowell: There's another subtlety here--the constant having value 0 becomes the default because the underlying type is `int` (more details in https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/498#discussion_r1725666156). Whether this matters depends on how the constant is used. I only point it out for consideration in case you want a different default value or to identify when a field hasn't been assigned.

AndyOHart: I see, in that case, should we have 0 be None like what was discussed before, or have IMDS or AzureArc be 0, which would be the default.
I think since we usually default to imds, we could make DefaultToIMDS = 0 instead. Make sense @bgavrilMS

bgavrilMS (approved):

PR #500, opened 2024-09-02
Acquire token for IMDS SAMI

4gust:
# Add ManagedIdentity Class and acquireToken Function for SystemAssigned Identity

## Overview

This PR introduces a new `ManagedIdentity` class in the Go codebase, specifically designed to handle operations related to managed identities in Azure. The primary feature added in this PR is the `acquireToken` function within the `ManagedIdentity` class, which currently supports only SystemAssigned managed identities.

## Changes

- **ManagedIdentity**: 
  - Created a new class named `ManagedIdentity` to encapsulate managed identity operations.
  - The class is intended to support different types of managed identities (SystemAssigned, UserAssigned) in the future, but for now, it only supports SystemAssigned identities.

- **acquireToken Function**: 
  - Implemented an `acquireToken` function in the `ManagedIdentity` class.
  - This function is responsible for acquiring an access token for the SystemAssigned managed identity.
  - The function makes a call to the Azure Instance Metadata Service (IMDS) endpoint to fetch the token.

## Implementation Details

- **SystemAssigned Identity**:
  - The `acquireToken` function currently assumes that the managed identity is SystemAssigned.
  - It retrieves the access token using the endpoint: `http://127.0.0.1:40342/metadata/identity/oauth2/token`.

- **Error Handling**:
  - The function includes basic error handling to manage cases where the token cannot be acquired.
  - If the HTTP request to the IMDS endpoint fails, the function returns an error with appropriate context.

## Testing

- **Unit Tests**:
  - Added unit tests for the `acquireToken` function to ensure it behaves as expected when fetching tokens for a SystemAssigned identity.
  - Tests include scenarios for successful token acquisition and error cases (e.g., network failure).

## Checklist

- [x] Code compiles successfully.
- [x] Unit tests added and pass locally. (more work to be done here)

## Notes

- This is the initial implementation of the `ManagedIdentity` class with limited functionality. Please review and suggest improvements, especially concerning the design for future extensibility.


apps/managedidentity/managedidentity.go
```diff
@@ -11,63 +11,85 @@ package managedidentity
...
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops"
-	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/authority"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/accesstokens"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/shared"
 )
 
 const (
-	// DefaultToIMDS indicates that the source is defaulted to IMDS since no environment variables are set.
-	DefaultToIMDS = 0
+	MetaHTTPHeadderName               = "Metadata"
```
AndyOHart, line 27: I would maybe move everything to one const(), and just separate into sections using comments perhaps?

apps/managedidentity/managedidentity.go
```diff
@@ -11,63 +11,66 @@ package managedidentity
...
-	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/authority"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/accesstokens"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/shared"
 )
 
+// General request querry parameter names
 const (
-	// DefaultToIMDS indicates that the source is defaulted to IMDS since no environment variables are set.
-	DefaultToIMDS = 0
+	MetaHTTPHeadderName               = "Metadata"
```
chlowell, line 28: Reconsider exporting all these constants. Do applications have any reason to use them?

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
 func (r ResourceID) value() string          { return string(r) }
+func SystemAssigned() ID {
+	return systemAssignedValue("")
+}
+
+// Client is a client that provides access to Managed Identity token calls.
+type Client struct {
+	// cacheAccessorMu *sync.RWMutex
+	httpClient ops.HTTPClient
+	MiType     ID
```
chlowell, line 90: Applications can change this at any time because it's exported. This client would be more consistent with the confidential client and I expect easier to reason about if the ID were settable only at construction.

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	switch client.MiType.(type) {
+	case ClientID:
+		msiParameters.Add(MIQuerryParameterClientId, client.MiType.value())
+	case ResourceID:
+		msiParameters.Add(MIQuerryParameterResourceId, client.MiType.value())
+	case ObjectID:
+		msiParameters.Add(MIQuerryParameterObjectId, client.MiType.value())
+	case systemAssignedValue: // not adding anything
+	default:
+		return base.AuthResult{}, fmt.Errorf("Type not suported")
```
chlowell, line 219: Would be helpful to note the type name
```suggestion
		return base.AuthResult{}, fmt.Errorf("unsupported type %T", client.miType)
```

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	// pmanager manager 	// todo :  expose the manager from base.
+	// cacheAccessor   		cache.ExportReplace
+}
+
+type ClientOptions struct {
+	httpClient ops.HTTPClient
+}
+
+type AcquireTokenOptions struct {
+	Claims string
```
chlowell, line 100:
```suggestion
	claims string
```

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
 // ID: [SystemAssigned()], [ClientID("clientID")], [ResourceID("resourceID")], [ObjectID("objectID")]
 //
 // Options: [WithHTTPClient]
-func New(id ID, options ...Option) (Client, error) {
+func New(id ID, options ...ClientOption) (Client, error) {
 	fmt.Println("idType: ", id.value())
 
-	opts := clientOptions{
-		claims: "claims",
+	opts := ClientOptions{ // work on this side where
```
chlowell, line 129:
```suggestion
	opts := ClientOptions{
```

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	return client, nil
+}
+
+func getTokenForURL(url *url.URL, httpClient ops.HTTPClient) (accesstokens.TokenResponse, error) {
+	req, err := http.NewRequest(http.MethodGet, url.String(), nil)
 	if err != nil {
-		return Client{}, err
+		return accesstokens.TokenResponse{}, err
 	}
+	req.Header.Add("Metadata", "true")
```
chlowell, line 150:
```suggestion
	req.Header.Add(metaHTTPHeaderName, "true")
```

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	// could shell  POST request
+	// azure arc  GET
+	// default :: IMDS  GET
+	//
+	// Sources that send GET requests: App Service, Azure Arc, IMDS, Service Fabric
+	//
+	// Sources that send POST requests: Cloud Shell
+
+	var msiEndpoint *url.URL
+	msiEndpoint, err := url.Parse(IMDSEndpoint)
```
chlowell, line 197: We don't need to parse this every time--it's constant

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+		return base.AuthResult{}, fmt.Errorf("Type not suported")
+
+	}
+
+	msiEndpoint.RawQuery = msiParameters.Encode()
+	token, err := getTokenForURL(msiEndpoint, client.httpClient)
+	if err != nil {
+		return base.AuthResult{}, fmt.Errorf("URL not formed")
+	}
+	println("Access token **  ", token.AccessToken)
```
chlowell, line 228:
```suggestion
```

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	case systemAssignedValue: // not adding anything
+	default:
+		return base.AuthResult{}, fmt.Errorf("Type not suported")
+
+	}
+
+	msiEndpoint.RawQuery = msiParameters.Encode()
+	token, err := getTokenForURL(msiEndpoint, client.httpClient)
+	if err != nil {
+		return base.AuthResult{}, fmt.Errorf("URL not formed")
```
chlowell, line 226:
```suggestion
		return base.AuthResult{}, err
```

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	msiParameters.Add("api-version", "2018-02-01")
+	msiParameters.Add("resource", resource)
+
+	if len(o.Claims) > 0 {
+		msiParameters.Add("claims", o.Claims)
+	}
+
+	switch client.MiType.(type) {
+	case ClientID:
+		msiParameters.Add(MIQuerryParameterClientId, client.MiType.value())
```
chlowell, line 212: I wouldn't die on this hill, but you don't need this method
```suggestion
		msiParameters.Add(MIQuerryParameterClientId, string(client.miType))
```

apps/managedidentity/managedidentity.go
```diff
@@ -76,59 +79,154 @@ func (s systemAssignedValue) value() string { return string(s) }
...
-		// manager:         storage.New(token),
-		// pmanager:        storage.NewPartitionedManager(token),
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return accesstokens.TokenResponse{}, err
 	}
 
-	return client, err
+	// Pull out response body
+	responseBytes, err := io.ReadAll(resp.Body)
```
chlowell, line 158: What if the response status isn't `OK`?

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,90 @@ package managedidentity
...
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	_, err = client.AcquireToken(context.Background(), "scope", WithClaims("claim"))
-
-	if err == nil {
-		t.Errorf("TestManagedIdentityWithClaims: unexpected nil error from TestManagedIdentityWithClaims")
+	if _, err := client.AcquireToken(context.Background(), "fakeresource"); err == nil {
+		t.Errorf("TestManagedIdentity: Should have returned error for incorrect http request.")
```
chlowell, line 90:
```suggestion
		t.Fatal("TestManagedIdentity: Should have returned error for incorrect http request.")
```

chlowell (changes requested):

apps/managedidentity/managedidentity.go
4gust: I think the function does the same thing, and when i tried this it failed for some reason, as now it becomes of type ID interface in this function and to do that we need to do like this
`client.miType.(string)` only then it would work and this for me looks better

apps/managedidentity/managedidentity.go
4gust: We will need to parse the URL constant no ?
I could not find a way to make url constant as url is struct.

apps/managedidentity/managedidentity.go
```diff
@@ -11,63 +11,51 @@ package managedidentity
...
-	// clientId     string
-}
-
-type withClaimsOption struct{ Claims string }
-type withHTTPClientOption struct{ HttpClient ops.HTTPClient }
+// UAMI querry parameter name
+const (
+	miQuerryParameterClientId   = "client_id"
+	miQuerryParameterObjectId   = "object_id"
+	miQuerryParameterResourceId = "mi_res_id"
```
chlowell, line 37:
```suggestion
	miQuerryParameterResourceId = "msi_res_id"
```
https://github.com/Azure/azure-rest-api-specs/blob/dba6ed1f03bda88ac6884c0a883246446cc72495/specification/imds/data-plane/Microsoft.InstanceMetadataService/stable/2018-10-01/imds.json#L234

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +64,136 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	req.Header.Add(metaHTTPHeadderName, "true")
 
-	authParams := authority.NewAuthParams(id.value(), authInfo)
-	client := Client{ // Note: Hey, don't even THINK about making Base into *Base. See "design notes" in public.go and confidential.go
-		AuthParams:      authParams,
-		cacheAccessorMu: &sync.RWMutex{},
-		// manager:         storage.New(token),
-		// pmanager:        storage.NewPartitionedManager(token),
+	resp, err := httpClient.Do(req)
+	if err != nil {
```
bgavrilMS, line 132: nit: Is there an more general mechanism like an error factory that can take some of this logic? Seems like it's going to be needed quite a lot.

Probably ok for now, but I can see this type of code being repeated for Azure Arc etc. And also for talking to ESTS.

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +64,136 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	client := Client{
+		miType:     id,
+		httpClient: opts.httpClient,
 	}
 
-	authInfo, err := authority.NewInfoFromAuthorityURI("authorityURI", true, false)
+	return client, nil
+}
+
+func getTokenForURL(url *url.URL, httpClient ops.HTTPClient) (accesstokens.TokenResponse, error) {
```
bgavrilMS, line 124: nit: getTokenFromHttp or getTokenFromImds ?

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +64,136 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+		return base.AuthResult{}, nil
+	}
+	msiParameters := msiEndpoint.Query()
+	msiParameters.Add(apiVersionQuerryParameterName, "2018-02-01")
+	msiParameters.Add(resourceQuerryParameterName, resource)
 
-// Detects and returns the managed identity source available on the environment.
-func GetSource() Source {
-	return DefaultToIMDS
+	if len(o.claims) > 0 {
```
bgavrilMS, line 176: I do not think this is needed. Claims should just be ignored for now. Only bypass the token cache (but caching isn't implemented)

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,120 @@ package managedidentity
...
+}
+
+func (c *FakeClient) CloseIdleConnections() {}
+func (c *FakeClient) Do(req *http.Request) (*http.Response, error) {
+	println(c.responseType)
+	w := makeResponse(c.responseType)
+	return &w, nil
+}
+
+func makeResponse(responseType int) http.Response {
```
bgavrilMS, line 46: Consider creating separate methods for each response type. This param makes the code hard to read.

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,120 @@ package managedidentity
...
+	if responseType == 1 {
+		return http.Response{
+			StatusCode: http.StatusOK,
```
bgavrilMS, line 49: Please also have tests for: 

404 (this is when IMDS service may not be available)
501 and 503 (service errors)
400 + actual error message (e.g. to simulate a user error, such as invalid UAMI client ID specified)

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,120 @@ package managedidentity
...
+			Header:     make(http.Header),
+		}
+	}
 }
+func TestManagedIdentityIMDS_SAMISuccess(t *testing.T) {
+	fakeHTTPClient := FakeClient{responseType: 1}
 
-func TestManagedIdentity(t *testing.T) {
-	client, err := fakeClient(SystemAssigned())
+	client, err := fakeMIClient(SystemAssigned(), WithHTTPClient(&fakeHTTPClient))
```
bgavrilMS, line 71: Consider ways to make these tests more readable. For example this private function doesn't really seem to be needed. A bit of dup code to improve readability is preferred.

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +64,136 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+
+	var msiEndpoint *url.URL
+	msiEndpoint, err := url.Parse(imdsEndpoint)
+	if err != nil {
+		fmt.Println("Error creating URL: ", err)
+		return base.AuthResult{}, nil
+	}
+	msiParameters := msiEndpoint.Query()
+	msiParameters.Add(apiVersionQuerryParameterName, "2018-02-01")
+	msiParameters.Add(resourceQuerryParameterName, resource)
```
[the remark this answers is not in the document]
bgavrilMS: (internally you can just parse the resource string and put the correct format on the wire)

bgavrilMS (changes requested):
Looks good, some more tests are needed I think, particularly for getting error responses from IMDS (like 400 error with details about the error)

apps/managedidentity/managedidentity.go
chlowell: That's right, my suggestion wouldn't work because `client.miType` is an interface type. You would need the concrete type:
```go
switch t := client.miType.(type) {
case ClientID:
	msiParameters.Add(MIQuerryParameterClientId, string(t))
// ...
```

apps/managedidentity/managedidentity.go
chlowell: Line 135 is a response to my comment but perhaps that comment was too vague. As written, when the response isn't OK, we return an error (string) indicating the status code. That's it--just the status code. We should at least return a [CallErr](https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-library-for-go/apps/errors#CallErr) so the caller can inspect the response. It would also be good to include any body content in the error message so callers don't always have to inspect the response to figure out what went wrong.

apps/managedidentity/managedidentity.go
chlowell: For better or worse, this verbose and repetitive error handling is idiomatic 😬

This block is fine because a non-nil error from `Do` means the client didn't get a response. This function can't do anything more without a response, so it must stop here and return an error.

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +64,136 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+		miType:     id,
+		httpClient: opts.httpClient,
 	}
 
-	authInfo, err := authority.NewInfoFromAuthorityURI("authorityURI", true, false)
+	return client, nil
+}
+
+func getTokenForURL(url *url.URL, httpClient ops.HTTPClient) (accesstokens.TokenResponse, error) {
+	req, err := http.NewRequest(http.MethodGet, url.String(), nil)
```
chlowell, line 125: This should be [NewRequestWithContext](https://pkg.go.dev/net/http#NewRequestWithContext). All requests must honor the caller's `Context`

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +64,136 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	}
+	return r, nil
+
 }
 
 // Acquires tokens from the configured managed identity on an azure resource.
 //
 // Resource: scopes application is requesting access to
 // Options: [WithClaims]
 func (client Client) AcquireToken(context context.Context, resource string, options ...AcquireTokenOption) (base.AuthResult, error) {
```
chlowell, line 159: Generally best not to shadow names
```suggestion
func (client Client) AcquireToken(ctx context.Context, resource string, options ...AcquireTokenOption) (base.AuthResult, error) {
```

apps/managedidentity/managedidentity.go
chlowell: Sorry for the distraction, you make a good point: although you don't strictly speaking need a `url.URL`, that type's API is the most convenient way to format a query string.

apps/managedidentity/managedidentity.go
4gust: Thank you for this, i will update it with CallErr

apps/managedidentity/managedidentity.go
4gust: Adapted this method.

apps/managedidentity/managedidentity.go
4gust: This is local method which only does network call only, same will be used for others as well

apps/managedidentity/managedidentity.go
4gust: Noted!

apps/managedidentity/managedidentity.go
chlowell: There's also a broader consideration here: managed identity APIs differ in more than their URL, so the URL isn't enough input to make a valid token request. Consider separating constructing requests from sending them so you can handle each API's particulars easily ([for example](https://github.com/Azure/azure-sdk-for-go/blob/cc560d64d482b2ea001b4f6f1f19db63be91d389/sdk/azidentity/managed_identity_client.go#L316)).

apps/managedidentity/managedidentity.go
4gust: Removed the method, now made it using `http.Request`
Which in turn will have better customization 
also added context pass through
But I am not sure how we can use or where so for now i just have `_ context.Context`

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,120 @@ package managedidentity
...
+	return &w, nil
+}
+
+func makeResponse(responseType int) http.Response {
+	if responseType == 1 {
+		return http.Response{
+			StatusCode: http.StatusOK,
+			Body: io.NopCloser(strings.NewReader(`{
+      "access_token": "fakeToken",
+      "refresh_token": "",
```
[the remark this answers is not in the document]
chlowell: It doesn't: https://github.com/Azure/azure-rest-api-specs/blob/3519c80fe510a268f6e59a29ccac8a53fdec15b6/specification/imds/data-plane/Microsoft.InstanceMetadataService/stable/2018-02-01/imds.json#L533-L578

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,244 @@ package managedidentity
...
+		resource string
+		claims   string
+		wantErr  bool
+	}{
+		{
+			name:     "System Assigned",
+			id:       SystemAssigned(),
+			resource: "https://management.azure.com",
+			claims:   "",
+			wantErr:  false,
```
chlowell, line 108: Just a minor style point: these assignments are unnecessary because the zero values of `string` and `bool` are `""` and `false`

apps/managedidentity/managedidentity.go
chlowell: [NewRequestWithContext](https://pkg.go.dev/net/http#NewRequestWithContext) takes a `Context` as its first argument. So, pass that function the caller's `Context`

4gust: I removed this method as it only takes URL, and I created new method `getTokenForRequest`

apps/managedidentity/managedidentity.go
```diff
@@ -11,63 +11,80 @@ package managedidentity
...
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops"
-	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/authority"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/accesstokens"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/shared"
 )
 
 const (
-	// DefaultToIMDS indicates that the source is defaulted to IMDS since no environment variables are set.
```
bgavrilMS, line 23: Comment still makes sense.

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +93,155 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+}
+
+type ClientOption func(o *ClientOptions)
 
-func (w withClaimsOption) AcquireTokenOption()           {}
-func (w withHTTPClientOption) AcquireTokenOption()       {}
-func (w withHTTPClientOption) apply(opts *clientOptions) { opts.httpClient = w.HttpClient }
+type AcquireTokenOption func(o *AcquireTokenOptions)
 
 // WithClaims sets additional claims to request for the token, such as those required by conditional access policies.
```
bgavrilMS, line 117:
```suggestion
// WithClaims sets additional claims to request for the token, such as those required by token revocation or conditional access policies.
```

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +93,155 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+
+	return client, nil
+}
+
+func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims string) (*http.Request, error) {
+	var msiEndpoint *url.URL
+	msiEndpoint, err := url.Parse(imdsEndpoint)
 	if err != nil {
-		return Client{}, err
+		return nil, errors.New("error creating URL as parsing the URL filed")
```
bgavrilMS, line 157: Pls review all the erorr messages and make them more actionalbe. Put yourself in the shoes of an app developer who has trouble getting a token. 

Error messages serve a dual purpose: 

1. Tell the app developer how to mitigate the problem (if possible)
2. Give us (the MSAL team) enough details to troubleshoot

bgavrilMS, line 157: Arguably this one should never occur.

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +93,155 @@ func (s systemAssignedValue) value() string { return string(s) }
...
-	authInfo, err := authority.NewInfoFromAuthorityURI("authorityURI", true, false)
+	client := Client{
+		miType:     id,
+		httpClient: opts.httpClient,
+	}
+
+	return client, nil
+}
+
+func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims string) (*http.Request, error) {
```
bgavrilMS, line 153: `ID` seems too generic as a name.

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +93,155 @@ func (s systemAssignedValue) value() string { return string(s) }
...
 // Client to be used to acquire tokens for managed identity.
 // ID: [SystemAssigned()], [ClientID("clientID")], [ResourceID("resourceID")], [ObjectID("objectID")]
 //
 // Options: [WithHTTPClient]
-func New(id ID, options ...Option) (Client, error) {
-	fmt.Println("idType: ", id.value())
-
-	opts := clientOptions{
-		claims: "claims",
+func New(id ID, options ...ClientOption) (Client, error) {
```
bgavrilMS, line 136: How about `managedIdentityId ManagedIdentityId` like in .NET?

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +93,155 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	if len(claims) > 0 {
+		msiParameters.Add("claims", claims)
+	}
+
+	switch t := id.(type) {
+	case ClientID:
+		if len(string(t)) > 0 {
+			msiParameters.Add(miQuerryParameterClientId, string(t))
+		} else {
+			return nil, fmt.Errorf("clientId parameter is empty for %T", t)
```
bgavrilMS, line 176: These checks should happen in the constructor (the New function)? Otherwise we will repeat them for each new source we add (Arc, App Service etc.)

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +93,155 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	msiEndpoint.RawQuery = msiParameters.Encode()
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, msiEndpoint.String(), nil)
+	if err != nil {
+		return nil, errors.New("error creating request")
+	}
+	return req, nil
+}
+
+func getTokenForRequest(req *http.Request, httpClient ops.HTTPClient) (accesstokens.TokenResponse, error) {
+	req.Header.Add(metaHTTPHeadderName, "true")
```
bgavrilMS, line 204: Logically this one belongs to the request creator function above.

apps/managedidentity/managedidentity.go
bgavrilMS: Thanks @chlowell, that makes sense. +1 to including the error message. 

For errors such as 404, 503 etc. we can only include the HttpStatus code (e.g. 503 - Service Not Available) and maybe a hint to the developer on what to do, e.g. "Managed Identity request failed with HTTP error code 404. Source: IMDS. Possible cause: Not running on an Azure host or the IMDS service is temporarily unavailable"

For 400 error (Bad Request), we need to make sure we extract the exact error message from the service. For example "Incorrect client ID". 

It is necessary to have tests for these 2 type of service errors.

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,399 @@ package managedidentity
...
+	return []resourceTestData{
+		{source: DefaultToIMDS, endpoint: imdsEndpoint, resource: resource},
+		{source: DefaultToIMDS, endpoint: imdsEndpoint, resource: resourceDefaultSuffix},
+	}
+}
+
+func Test_SystemAssigned_Returns_Token_Failure(t *testing.T) {
+	testCases := []ErrorResponse{
+		{StatusCode: 404, Message: "IMDS service not available", CorrelationID: "121212"},
+		{StatusCode: 501, Message: "Service error 1", CorrelationID: "121212"},
```
bgavrilMS, line 216: There is no message 404 or 5xx errors, only with 400. Please add a test for this.

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,399 @@ package managedidentity
...
+	jsonResponse, _ := json.Marshal(response)
+	return string(jsonResponse)
+}
 
-	if err == nil {
-		t.Errorf("TestManagedIdentity: unexpected nil error from TestManagedIdentity")
+func getMsiErrorResponseNoRetry() string {
+	response := ErrorResponse{
+		StatusCode:    123,
+		Message:       "Not one of the retryable error responses",
```
bgavrilMS, line 129: You didn't implement retry policies so I don't see the point of this.

bgavrilMS (changes requested):
Small comments. Blocker is throwing a consistent exception for service errors, as per the https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/main/apps/errors/error_design.md

apps/managedidentity/managedidentity.go
chlowell: 👍 not all APIs expect this header

apps/managedidentity/managedidentity.go
chlowell: Users have no practical reason to reference this type but if they did, they would have to qualify it by package name. So this suggestion would have them write `managedidentity.ManagedIdentityId` instead of `managedidentity.ID`. A type name that repeats its package name is said to stutter and that's considered bad style in Go.

apps/managedidentity/managedidentity_test.go
4gust: Thanks, updated this.
will take note of this for furture.

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +93,155 @@ func (s systemAssignedValue) value() string { return string(s) }
...
 func (o ObjectID) value() string            { return string(o) }
 func (r ResourceID) value() string          { return string(r) }
+func SystemAssigned() ID {
+	return systemAssignedValue("")
+}
+
+type Client struct {
+	httpClient ops.HTTPClient
+	miType     ID
+}
```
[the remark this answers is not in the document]
4gust: I plan on adding GetSource() in future.
Noted.

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +93,155 @@ func (s systemAssignedValue) value() string { return string(s) }
...
-	return client, err
+	msiEndpoint.RawQuery = msiParameters.Encode()
+	req, err := http.NewRequest(http.MethodGet, msiEndpoint.String(), nil)
+	if err != nil {
+		return nil, errors.New("error creating request")
+	}
+	return req, nil
+}
+
+func getTokenForRequest(ctx context.Context, req *http.Request, httpClient ops.HTTPClient) (accesstokens.TokenResponse, error) {
```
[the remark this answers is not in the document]
4gust: Updated it

apps/managedidentity/managedidentity_test.go
4gust: Removed the method

apps/managedidentity/managedidentity_test.go
4gust: I updated the error testing like in Azure sdk. 
This also tests for service errors.

apps/managedidentity/managedidentity.go
4gust: Updated this added this for `createIMDSAuthRequest`

apps/managedidentity/managedidentity.go
```diff
@@ -76,62 +93,155 @@ func (s systemAssignedValue) value() string { return string(s) }
...
+	case systemAssignedValue: // not adding anything
+	default:
+		return nil, fmt.Errorf("unsupported type %T", id)
 	}
 
-	return client, err
+	msiEndpoint.RawQuery = msiParameters.Encode()
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, msiEndpoint.String(), nil)
+	if err != nil {
+		return nil, errors.New("error creating request")
```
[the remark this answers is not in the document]
4gust: I updated errors to have more information.

apps/managedidentity/managedidentity.go
4gust: Moved this to `New()` for `Client`
Added test to check that

apps/managedidentity/managedidentity.go
4gust: Passed as much information we get in the error in return errors.

apps/managedidentity/managedidentity.go
```diff
@@ -11,63 +11,80 @@ package managedidentity
...
+const (
+	imdsEndpoint   = "http://169.254.169.254/metadata/identity/oauth2/token"
+	imdsAPIVersion = "2018-02-01"
+)
 
-type withClaimsOption struct{ Claims string }
-type withHTTPClientOption struct{ HttpClient ops.HTTPClient }
+const (
+	// DefaultToIMDS indicates that the source is defaulted to IMDS since no environment variables are set.
+	defaultToIMDS = 0
```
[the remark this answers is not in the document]
4gust: Removed it.

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,401 @@ package managedidentity
...
+			req, err := createIMDSAuthRequest(context.Background(), tt.id, tt.resource, tt.claims)
+			if tt.wantErr {
+				if err == nil {
+					t.Fatal(err)
+				}
+				return
+			}
+			if req == nil {
+				t.Fatal("createIMDSAuthRequest() returned nil request")
+				return
```
chlowell, line 367: `t.Fatal` stops test execution
```suggestion
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,401 @@ package managedidentity
...
+			}
+			if req.Method != http.MethodGet {
+				t.Fatal("createIMDSAuthRequest() method is not GET")
+			}
+			if !strings.HasPrefix(req.URL.String(), imdsEndpoint) {
+				t.Fatal("createIMDSAuthRequest() URL is not matched.")
```
chlowell, line 373: A test failure like this is easier to debug when the message includes want/got values
```suggestion
			if got := req.URL.String(); !strings.HasPrefix(got, imdsEndpoint) {
				t.Fatalf("wanted %q, got %q", imdsEndpoint, got)
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,401 @@ package managedidentity
...
+func getSuccessfulResponse(resource string) string {
+	expiresOn := time.Now().Add(1 * time.Hour).Unix()
+	response := SuccessfulResponse{
+		AccessToken: "fakeToken",
+		ExpiresOn:   expiresOn,
+		Resource:    resource,
+		TokenType:   "Bearer",
+		ClientID:    "client_id",
+	}
+	jsonResponse, _ := json.Marshal(response)
```
chlowell, line 93: A test calling this function should fail when it gets a non-nil error. Test bugs may be unlikely here but not checking the error makes the eventual failure harder to understand or allows a test to pass when it should fail

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,401 @@ package managedidentity
...
+}
+
+type ErrorResponse struct {
+	StatusCode    int    `json:"statusCode"`
+	Message       string `json:"message"`
+	CorrelationID string `json:"correlationId,omitempty"`
+}
+
+type fakeClient struct{}
+type errorClient struct {
```
chlowell, line 49: 🤔 you could have just one fake client type. In fact, I imagine you could use [this one](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/main/apps/internal/mock/mock.go#L49-L50) if you add a `WithStatusCode` option for it

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,401 @@ package managedidentity
...
+	errResponse ErrorResponse
+}
+
+func fakeMIClient(mangedIdentityId ID, options ...ClientOption) (Client, error) {
```
chlowell, line 53: Why not call `New` instead?

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,401 @@ package managedidentity
...
+			queryString.Add(key, value)
+		}
 	}
+
+	return endpoint + "?" + queryString.Encode()
 }
 
-func TestManagedIdentityWithClaims(t *testing.T) {
-	client, err := fakeClient(ClientID("123"))
+func expectedRequest(source Source, resource string, id ID) (*http.Request, error) {
```
chlowell, line 122: unused

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,401 @@ package managedidentity
...
+			client, err := New(tt.id)
+			if tt.wantErr {
+				if err == nil {
+					t.Fatal("client New() should return a error but did not.")
+				}
+				return
+			}
+			if err != nil {
+				t.Fatal("client New() error while creating client")
+			} else {
```
chlowell, line 306: You don't need the `else` because `t.Fatal` stops execution

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,401 @@ package managedidentity
...
+			client, err := fakeMIClient(SystemAssigned(), WithHTTPClient(&fakeErrorClient))
+			if err != nil {
+				t.Fatal(err)
+			}
+			resp, err := client.AcquireToken(context.Background(), resource)
+			if err == nil {
+				t.Fatalf("testManagedIdentity: Should have encountered the error")
+			}
+			switch e := err.(type) {
+			case errors.CallErr:
```
chlowell, line 207: Looks like these tests should fail when the error is a different type. However, as written they just skip error validation in that case. Try
```go
var e errors.CallErr
if errors.As(err, &e) {
	// TODO
} else {
	t.Fatalf("wanted %T, got %T", e, err)
}
```

apps/managedidentity/managedidentity_test.go
4gust: Good idea, will update.

apps/managedidentity/managedidentity_test.go
4gust: Removing

apps/internal/mock/mock.go
```diff
@@ -59,6 +59,14 @@ func (c *Client) AppendResponse(opts ...responseOption) {
 	c.resp = append(c.resp, r)
 }
 
+func (c *Client) AppendCustomResponse(status int, opts ...responseOption) {
```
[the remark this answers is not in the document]
4gust: Good Idea, updated it.

apps/internal/mock/mock.go
```diff
@@ -46,6 +46,13 @@ func WithCallback(callback func(*http.Request)) responseOption {
 	})
 }
 
+// WithCallback sets a callback to invoke before returning the response.
+func WithHttpStatusCode(statusCode int) responseOption {
```
chlowell, line 50: https://go.dev/wiki/CodeReviewComments#initialisms
```suggestion
func WithHTTPStatusCode(statusCode int) responseOption {
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,401 @@ package managedidentity
...
+			if resp.AccessToken != "" {
+				t.Fatalf("testManagedIdentity: accesstoken should be nil")
+			}
+
+		})
+	}
+}
+
+func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
+	testCases := createResourceData()
```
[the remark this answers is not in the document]
4gust: Updated the test.
I though it might be useful when we test UAMI but this makes sense

apps/tests/devapps/managedidentity_sample.go
```diff
@@ -3,36 +3,18 @@ package main
...
-		fmt.Println(error)
+	miSystemAssigned, err := mi.New(mi.SystemAssigned())
+	if err != nil {
+		fmt.Println(err)
 	}
-
-	miResourceIdAssigned, error := mi.New(mi.ResourceID("resource id 123"))
-	if error != nil {
-		fmt.Println(error)
+	temp, err := miSystemAssigned.AcquireToken(context.Background(), "https://management.azure.com/")
```
bgavrilMS, line 15: result, err

apps/tests/devapps/managedidentity_sample.go
```diff
@@ -3,36 +3,18 @@ package main
...
-	miObjectIdAssigned, error := mi.New(mi.ObjectID("object id 123"))
-	if error != nil {
-		fmt.Println(error)
-	}
-
-	miSystemAssigned.AcquireToken(context.Background(), "resource", mi.WithClaims("claim"))
-	miClientIdAssigned.AcquireToken(context.Background(), "resource")
-	miResourceIdAssigned.AcquireToken(context.Background(), "resource", mi.WithClaims("claim"))
-	miObjectIdAssigned.AcquireToken(context.Background(), "resource")
+	fmt.Println("token : ", temp.AccessToken)
```
bgavrilMS, line 19: Pls do not "print" access tokens. It's bad practice. Better to print the token expiration time instead and maybe the token source (cache / identity provider).

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,319 @@ package managedidentity
...
-		t.Errorf("TestManagedIdentity: unexpected nil error from TestManagedIdentity")
+			if err != nil {
+				t.Fatal(err)
+			}
+			result, err := client.AcquireToken(context.Background(), testCase.resource)
+			if !strings.HasPrefix(url, testCase.endpoint) {
+				t.Fatalf("TestManagedIdentity: URL request is not on %s fgot %s", testCase.endpoint, url)
+			}
+			if err != nil {
+				t.Fatalf("TestManagedIdentity: unexpected nil error from TestManagedIdentity %s", err.Error())
```
chlowell, line 166:
```suggestion
				t.Fatal(err)
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,319 @@ package managedidentity
...
+		t.Run(tt.name, func(t *testing.T) {
+			client, err := New(tt.id)
+			if tt.wantErr {
+				if err == nil {
+					t.Fatal("client New() should return a error but did not.")
+				}
+				return
+			}
+			if err != nil {
+				t.Fatal("client New() error while creating client")
```
chlowell, line 224: what was the error?
```suggestion
				t.Fatal(err)
```

apps/tests/devapps/managedidentity_sample.go
```diff
@@ -3,36 +3,18 @@ package main
...
+	if err != nil {
+		fmt.Println(err)
 	}
-
-	miResourceIdAssigned, error := mi.New(mi.ResourceID("resource id 123"))
-	if error != nil {
-		fmt.Println(error)
+	result, err := miSystemAssigned.AcquireToken(context.Background(), "https://management.azure.com/")
+	if err != nil {
+		println(err.Error())
```
chlowell, line 17:
```suggestion
		fmt.Println(err)
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,319 @@ package managedidentity
...
+	responseBody := ErrorRespone{
+		Err:  err,
+		Desc: desc,
+	}
+	if len(err) == 0 && len(desc) == 0 {
+		jsonResponse, error := json.Marshal(responseBody)
+		return jsonResponse, error
 	}
+	jsonResponse, error := json.Marshal(responseBody)
+	return jsonResponse, error
```
chlowell, line 68:
```
if foo {
    bar()
} else {
    bar()
}
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,319 @@ package managedidentity
...
-		return Client{}, err
+type response struct {
+	body     []byte
+	callback func(*http.Request)
+	code     int
+}
+
+func getSuccessfulResponse(resource string) ([]byte, error) {
+	expiresOn := time.Now().Add(1 * time.Hour).Unix()
+	response := SuccessfulResponse{
```
chlowell, line 47: You want the mock tests to be realistic, so you may as well start now by parameterizing this for the source and returning a response having exactly the same body fields as a real one

apps/managedidentity/managedidentity_test.go
```diff
@@ -4,43 +4,319 @@ package managedidentity
...
+					t.Fatal("client New() did not assign a correct value to type.")
+				}
+			}
+		})
+	}
 
-	if err != nil {
-		t.Fatal(err)
+}
+func TestCreateIMDSAuthRequest(t *testing.T) {
```
chlowell, line 234: Suggest merging this into `Test_SystemAssigned_Returns_Token_Success`. Driving tests through the public API is generally better--what happens if we refactor `createIMDSAuthRequest` out of existence or `AcquireToken` stops calling it?--and that test already has a callback validating the request URL.

AndyOHart (commented):
Looks good to me

bgavrilMS (approved):

apps/managedidentity/managedidentity_test.go
4gust: Updated and now only tests are for public api and we check for every step, even for creating of request and return a fatal error

apps/managedidentity/managedidentity_test.go
```diff
@@ -221,102 +209,12 @@ func TestCreatingIMDSClient(t *testing.T) {
...
 			} else {
 				if client.miType.value() != tt.id.value() {
 					t.Fatal("client New() did not assign a correct value to type.")
 				}
 			}
 		})
 	}
-
-}
-func TestCreateIMDSAuthRequest(t *testing.T) {
```
chlowell: I suggested merging this into another test, not simply deleting it; the coverage is important

chlowell (changes requested):

apps/managedidentity/managedidentity_test.go
4gust: I added that test to `Test_SystemAssigned_Returns_Token_Success`

apps/managedidentity/managedidentity_test.go
```diff
@@ -117,56 +105,56 @@ func Test_SystemAssigned_Returns_AcquireToken_Failure(t *testing.T) {
...
 			mockClient := mock.Client{}
-			responseBody, err := getSuccessfulResponse(resource)
+			responseBody, err := getSuccessfulResponse(resource, testCase.miType)
 			if err != nil {
-				t.Fatalf("TestManagedIdentity: Error while forming json response : %s", err.Error())
+				t.Fatalf("error while forming json response : %s", err.Error())
 			}
-			mockClient.AppendResponse(mock.WithHTTPStatusCode(http.StatusOK), mock.WithBody(responseBody), mock.WithCallback(func(r *http.Request) { url = r.URL.String() }))
-			client, err := New(SystemAssigned(), WithHTTPClient(&mockClient))
+			mockClient.AppendResponse(mock.WithHTTPStatusCode(http.StatusOK), mock.WithBody(responseBody))
```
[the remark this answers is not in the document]
4gust: Updated to check alot more values in the request.

apps/managedidentity/managedidentity_test.go
```diff
@@ -129,35 +130,155 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
+			name:     "Object ID",
+			id:       UserAssignedObjectID("test-object-id"),
+			resource: "https://graph.microsoft.com",
+		},
+		{
+			name:     "With Claims",
+			id:       SystemAssigned(),
+			resource: "https://management.azure.com",
+			claims:   "test-claims",
+		},
```
chlowell, line 231: This is the only case in this set that adds coverage. When I suggested merging these tests, I meant what you've now done above: validate HTTP requests as sent by `AcquireToken` instead of as returned by `createIMDSAuthRequest`. The connection between public API and HTTP requests matters, not how the client constructs requests.

Given IMDS doesn't support claims and all the other cases starting at line 193 are redundant, I recommend deleting them.

apps/tests/devapps/main.go
```diff
@@ -36,9 +36,8 @@ func main() {
 	} else if exampleType == "6" {
 		// This sample does not use a serialized cache - it relies on in-memory cache by reusing the app object
 		// This works well for app tokens, because there is only 1 token per resource, per tenant.
-		acquireTokenClientCertificate()
-
-		// this time the token comes from the cache!
+		// acquireTokenClientCertificate()
+		// // this time the token comes from the cache!
```
chlowell, line 40: 👀 do you intend to make this change?

chlowell (approved):

apps/tests/devapps/main.go
4gust: I Was testing the api on the VM so, to get these changes there pushed this, now reverted.

PR #502, opened 2024-09-05
Implement Skeleton Test Cases

AndyOHart:
* Updates managedidentity.go to match @4gust  PR, will update accordingly as his is merged
* Added skeleton tests based on Java and .NET
* Added some helper functions for tests
* Added test table for GetSource and Test_SystemAssigned_Returns_Token_Success, which acts similar to the @ParamaterizedTest in Java
* Updated some modules


chlowell:
Why vendor `yaml.v3`? Also, any tests for #500 belong in that PR.

AndyOHart:
> Why vendor `yaml.v3`? Also, any tests for #500 belong in that PR.

I was trying to use the testing lib and it wasn't working so I tried the vendor command and I think it brought all this stuff in. 
I left in to see what you think, maybe everything  there should be reverted?

Also for the test case, yes it is, but this PR was meant to be done before that PR to put some stuff in place, but I had to take some last minute unplanned leave so it was delayed. So once that PR is merged I will be merging it into this branch too, this branch is not to be merged until that branch is finished

go.mod
```diff
@@ -10,4 +10,10 @@ require (
 	github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8
 )
 
-require golang.org/x/sys v0.5.0 // indirect
+require (
```
bgavrilMS, line 13: Are all these ok to use as per the Open Source guidelines? https://repos.opensource.microsoft.com/

go.mod
```diff
@@ -10,4 +10,10 @@ require (
...
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	github.com/stretchr/testify v1.9.0 // indirect
+	golang.org/x/sys v0.5.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
```
bgavrilMS, line 18: I am very hesitant to accept new dependencies for any MSAL. There has to be a very strong reason to add a new dependency. 

I am not sure GoLang has a concept of "test dependencies" simialar to Java and Node or a clean separation of product vs test projects like .NET. So any test dependency today is seen as a product dependency. Unless the sitation has evolved @chlowell ?

As such, I consider all of the above as product depedencies, which provide an unacceptable burden to MSAL GO. 

- dependencies can have security issues. At worst, it creates a security vulnerability in MSAL Go, and possible in services using them. At "best", it's a just a matter of upgrading that dependency to a version that's not affected. Either way, it takes a LOT of time. 
- All dependencies need to be registered with https://repos.opensource.microsoft.com/ and by the looks of some of them, it's not going to happen. Even registering 4 new dependencies is going to take a long time.
image

go.sum
```diff
@@ -8,6 +10,13 @@ github.com/montanaflynn/stats v0.7.0 h1:r3y12KyNxj/Sb/iOE46ws+3mS1+MZca1wlHQFPsY
...
 github.com/pkg/browser v0.0.0-20210911075715-681adbf594b8/go.mod h1:HKlIX3XHQyzLZPlr7++PzdhaXEj94dEiJgZDTsxEqUI=
+github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
+github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/stretchr/testify v1.9.0 h1:HtqpIVDClZ4nwg75+f6Lvsy/wHu+3BoSGCbBAcpTsTg=
+github.com/stretchr/testify v1.9.0/go.mod h1:r2ic/lqez/lEtzL7wO/rwa5dbSLXVDPFyf8C91i36aY=
 golang.org/x/sys v0.0.0-20210616045830-e2b7044e8c71/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
 golang.org/x/sys v0.5.0 h1:MUK/U/4lj1t1oPg0HfuXDN/Z1wv31ZJ/YcPiGccS4DU=
 golang.org/x/sys v0.5.0/go.mod h1:oPkhp1MJrh7nUepCBck5+mAzfO9JrbApNNgaTdGDITg=
+gopkg.in/check.v1 v0.0.0-20161208181325-20d25e280405/go.mod h1:Co6ibVJAznAaIkqp8huTwlJQCZ016jof/cbN4VW5Yz0=
+gopkg.in/yaml.v3 v3.0.1 h1:fxVm/GzAzEWqLHuvctI91KS9hhNmmWOoWu0XTYJS7CA=
```
bgavrilMS, line 21: Referencing an open source dependency is one thing, and copying it locally is even more stringent. We need to check with Legal if this is permitted etc. I am not willing to go down this road for a yaml parser? Why do we need a yaml parser?

bgavrilMS (changes requested):
No new dependencies for MSAL Go pls.

bgavrilMS:
My recommendation with this PR is: 

- move any test to the existing IMDS release
- any "not implemented" test should go in a test spec document 
- use ^^ in future development

go.mod
chlowell: > I am not sure GoLang has a concept of "test dependencies" simialar to Java and Node or a clean separation of product vs test projects like .NET. So any test dependency today is seen as a product dependency. Unless the sitation has evolved @chlowell ?

Go does separate the dependencies required to build a module from those required to run its tests. When an application installs a module, it also installs that module's build dependencies, but not its test dependencies. So, packages imported only in MSAL Go's `*_test.go` files aren't distributed to users. Unfortunately, as you can see, this distinction isn't visible in `go.mod`.

chlowell:
> I was trying to use the testing lib and it wasn't working so I tried the vendor command and I think it brought all this stuff in. I left in to see what you think, maybe everything there should be reverted?
> 

Vendoring means adding an internal copy of some dependency to a module so that module can use the copy instead of the upstream original. That should never be necessary to run the tests, and in general it's something we'd do only in extraordinary circumstances.

AndyOHart:
> > I was trying to use the testing lib and it wasn't working so I tried the vendor command and I think it brought all this stuff in. I left in to see what you think, maybe everything there should be reverted?
> 
> Vendoring means adding an internal copy of some dependency to a module so that module can use the copy instead of the upstream original. That should never be necessary to run the tests, and in general it's something we'd do only in extraordinary circumstances.

Thanks, honestly not sure what brought everything in, I ran some commands the IDE said when i couldn't run tests and after I ran it, the tests worked, however it seems that it was actual some IDE issue in VS Code not cleaning up the vendor folder correctly, now everything works

PR #508, opened 2024-09-23
Fix local integration tests

AndyOHart:
* Makes some changes to integration_test.go to make the failing integration tests work again
* Note: TestConfidentialClientWithSecret and TestOnBehalfOf still do not work, they will require a bit of extra work that will be tracked in a separate issue.


bgavrilMS:
I think this fix should go directly into `main` branch. No need to base on MSI branch.

bgavrilMS (approved):

apps/tests/integration/integration_test.go
```diff
@@ -23,16 +25,20 @@ import (
...
+	labSecretUrl           = "https://msidlab.com/api/LabSecret"
 
-const (
 	organizationsAuthority = microsoftAuthorityHost + "organizations/"
-	microsoftAuthority     = microsoftAuthorityHost + "microsoft.onmicrosoft.com"
+	microsoftAuthority     = microsoftAuthorityHost + "72f988bf-86f1-41af-91ab-2d7cd011db47"
 	//msIDlabTenantAuthority = microsoftAuthorityHost + "msidlab4.onmicrosoft.com" - Will be needed in the future
+
+	// Default values
+	certificateAlias = "LabAuth.MSIDLab.com"
```
chlowell, line 39: unused

apps/tests/integration/integration_test.go
```diff
@@ -92,21 +98,39 @@ type secret struct {
...
+	app, err := confidential.New(microsoftAuthority, defaultClientId, cred, confidential.WithX5C())
 	if err != nil {
 		return nil, err
 	}
 
 	return &labClient{app: app}, nil
 }
+
+func getCertDataFromFile() ([]*x509.Certificate, crypto.PrivateKey, error) {
+	data, err := os.ReadFile(pemFile)
```
chlowell, line 120: How would this work in CI (`pemFile` is constant)?

AndyOHart: > How would this work in CI (`pemFile` is constant)?

This PR is just to get it working locally, the next step is looking into the CI side of things. 
Potentially downloading the cert on ci machine, using openssl to convert and save to a location and then processing it in a similar way

PR #514, opened 2024-10-04
Added in memory cache support for IMDS - SAMI and UAMI

4gust:
# Pull Request: Implementing Cache Support for Managed identity `AcquireToken`

## Overview
This PR introduces caching support for Managed identity `AcquireToken`, it is in-memory cache only.
Moed `storage` from Internal to base folder so we can reuse that for ManagedIdentity


apps/internal/base/storage/items.go
```diff
@@ -83,7 +84,7 @@ type AccessToken struct {
 }
 
 // NewAccessToken is the constructor for AccessToken.
-func NewAccessToken(homeID, env, realm, clientID string, cachedAt, expiresOn, extendedExpiresOn time.Time, scopes, token, tokenType, authnSchemeKeyID string) AccessToken {
+func NewAccessToken(homeID, env, realm, clientID string, cachedAt, expiresOn, extendedExpiresOn time.Time, scopes, token, tokenType, authnSchemeKeyID string, resource string) AccessToken {
```
bgavrilMS, line 87: Do we really need resource here? Can you reuse scopes instead? From caching point of view, it's just a string, we don't need to differentiate. The logical layer (the MSI classes) will know how to deal with it.

apps/internal/base/storage/partitioned_storage.go
```diff
@@ -91,6 +91,7 @@ func (m *PartitionedManager) Write(authParameters authority.AuthParams, tokenRes
 	environment := authParameters.AuthorityInfo.Host
 	realm := authParameters.AuthorityInfo.Tenant
 	clientID := authParameters.ClientID
+	resource := tokenResponse.Resource
```
bgavrilMS, line 94: MSAL uses the **request** scopes for caching. This is because the client could ask for scopes "S1, S2, S3" and they could get "S1, S2, S3, S4, S5" or just "S1, S3".

apps/managedidentity/managedidentity.go
```diff
@@ -64,13 +71,20 @@ func (c UserAssignedClientID) value() string   { return string(c) }
...
 }
 
 type Client struct {
 	httpClient ops.HTTPClient
 	miType     ID
 	// source     Source reenable when required in future sources
+	cacheManager manager
+}
+
+type manager interface {
```
bgavrilMS, line 84: Do we still need this? We discussed about not offering serialziation API for managed identity.

chlowell:
Another way to do this that's a little roundabout but simpler is to use [confidential.NewCredFromTokenProvider](https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential#NewCredFromTokenProvider) as does azidentity. Basically, you wrap your authentication code in a function [like this](https://github.com/Azure/azure-sdk-for-go/blob/109f5f85bf9e7424953b90f48ae0ee01569518fe/sdk/azidentity/managed_identity_client.go#L202-L211) and construct a `confidential.Client` [like this](https://github.com/Azure/azure-sdk-for-go/blob/109f5f85bf9e7424953b90f48ae0ee01569518fe/sdk/azidentity/managed_identity_credential.go#L126-L137) to call it and handle token caching. Then you can use that confidential client [like this](https://github.com/Azure/azure-sdk-for-go/blob/109f5f85bf9e7424953b90f48ae0ee01569518fe/sdk/azidentity/confidential_client.go#L104-L107) to read the cache and request new tokens. This approach is somewhat indirect but has the benefit of not requiring a change outside of `apps/managedidentity` and you know it works because azidentity's been doing it for a couple years.

apps/managedidentity/managedidentity.go
4gust: This is private interface, not exposed for outside. We need this so that we can have access to cache storage in client.

4gust:
After discussion with @bgavrilMS we decided that we will keep the working of this same as other MSAL's

apps/managedidentity/managedidentity.go
4gust: Removed.

apps/internal/base/storage/partitioned_storage.go
4gust: Updated that removed resource, now it is attached from request

apps/internal/base/storage/items.go
4gust: Removed.

apps/internal/base/storage/storage.go
```diff
@@ -173,10 +173,11 @@ func (m *Manager) Write(authParameters authority.AuthParams, tokenResponse acces
 	environment := authParameters.AuthorityInfo.Host
 	realm := authParameters.AuthorityInfo.Tenant
 	clientID := authParameters.ClientID
+
 	target := strings.Join(tokenResponse.GrantedScopes.Slice, scopeSeparator)
+	//tokenResponse.Resource
```
bgavrilMS, line 178: nit: dead code

apps/managedidentity/managedidentity.go
```diff
@@ -46,6 +49,9 @@ const (
 	// IMDS
 	imdsEndpoint   = "http://169.254.169.254/metadata/identity/oauth2/token"
 	imdsAPIVersion = "2018-02-01"
+
+	//system assigned managed identity
```
chlowell, line 53: I'd delete this as redundant but if you want a comment, please put a space after the `//` to avoid linter complaints

apps/managedidentity/managedidentity.go
```diff
@@ -223,9 +235,44 @@ func (client Client) AcquireToken(ctx context.Context, resource string, options
...
 	tokenResponse, err := client.getTokenForRequest(req)
 	if err != nil {
 		return base.AuthResult{}, err
 	}
-	return base.NewAuthResult(tokenResponse, shared.Account{})
+	return client.AuthResultFromToken(ctx, fakeAuthParams, tokenResponse, true)
+}
+
+func (c Client) AuthResultFromToken(ctx context.Context, authParams authority.AuthParams, token accesstokens.TokenResponse, cacheWrite bool) (base.AuthResult, error) {
+	if !cacheWrite {
```
chlowell, line 265: When would `cacheWrite` be false?

chlowell (changes requested):
Blocker is HTTP client customization and tenant/instance discovery. They're intertwined but the big one is discovery. Please add a test that fails when `Client` sends a discovery request. [confidential_test.go](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/bf74752ff2c5c2168d37ac090b2973cb5886ffc5/apps/confidential/confidential_test.go#L1106) has tests that mock a sequence of HTTP responses and fail when the client sends an unexpected request

apps/managedidentity/managedidentity.go
```diff
@@ -64,7 +70,13 @@ func (c UserAssignedClientID) value() string   { return string(c) }
 func (o UserAssignedObjectID) value() string   { return string(o) }
 func (r UserAssignedResourceID) value() string { return string(r) }
 func SystemAssigned() ID {
-	return systemAssignedValue("")
+	return systemAssignedValue(systemAssignedManagedIdentity)
+}
+
+var cacheManager *storage.Manager = storage.New(oauth.New(http.DefaultClient))
```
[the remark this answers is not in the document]
bgavrilMS: Yes, that is the goal - one cache per process for managed identity tokens.

apps/managedidentity/managedidentity_test.go
```diff
@@ -12,15 +12,16 @@ import (
 	"time"
```
bgavrilMS, line 12: Could you add some tests where you can verify the whole "1 cache per process" goal? i.e. create 2 clients, get a token from client1 (source should be IdP), get a token from client2 (source should be cache)

bgavrilMS (changes requested):
Missing static cache tests

apps/managedidentity/managedidentity.go
4gust: This is if we decide if we don't want to have cache for one method.
This is the active measure to have that. 
would you recommend we remove that ?

apps/managedidentity/managedidentity_test.go
4gust: I actually create a second client `secondFakeClient` which after every test and gets the token and checks if it is from cache

apps/managedidentity/managedidentity.go
bgavrilMS: The codepath with `cacheWrite=false` is never used by the product code, so it's essentially dead code. Let's not have dead code in the product.

Aside: MSAL always writes to cache.

apps/managedidentity/managedidentity_test.go
```diff
@@ -180,13 +189,35 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
 					t.Fatalf("resource objectid is incorrect, wanted %s got %s", i.value(), query.Get(miQueryParameterObjectId))
 				}
 			}
-			if err != nil {
-				t.Fatal(err)
+			if result.Metadata.TokenSource != base.IdentityProvider {
+				t.Fatalf("wanted Indenity tokensource, got %d", result.Metadata.TokenSource)
```
bgavrilMS, line 193:
```suggestion
				t.Fatalf("Expected IndenityProvider tokensource, got %d", result.Metadata.TokenSource)
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -180,13 +189,35 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
+			secondFakeClient, err := New(testCase.miType, WithHTTPClient(&mockClient))
+			if err != nil {
+				t.Fatal(err)
+			}
+			secondResult, err := secondFakeClient.AcquireToken(context.Background(), testCase.resource)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if secondResult.Metadata.TokenSource != base.Cache {
+				t.Fatalf("cache result wanted Indenity token source, got %d", secondResult.Metadata.TokenSource)
```
bgavrilMS, line 218:
```suggestion
				t.Fatalf("Expected token source to be cache, got %d", secondResult.Metadata.TokenSource)
```

bgavrilMS (approved):
LGTM, but please resolve @chlowell's comments.

apps/managedidentity/managedidentity_test.go
```diff
@@ -180,13 +189,35 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
-			if err != nil {
-				t.Fatal(err)
+			if result.Metadata.TokenSource != base.IdentityProvider {
+				t.Fatalf("wanted Indenity tokensource, got %d", result.Metadata.TokenSource)
 			}
 			if result.AccessToken != token {
 				t.Fatalf("wanted %q, got %q", token, result.AccessToken)
 			}
-
+			storage.FakeValidate = func(storage.AccessToken) error {
```
bgavrilMS, line 198: Can we have some tests that don't do so much mocking, but instead rely on injecting HTTP responses ?

I think we want to test: 

- create MI client with { SAMI, UAMI by client_Id, UAMI by obj_id, UAMI by res_id } 
- get token, assert it's from Identity Provider
- get token again, assert it's from cache
- create MI2 client, get token, assert it's from cache.

4gust: That is exactly what we are doing, We are injecting and http response 

- we are checking weather the http response was called.
- we are checking if the URL that it requested was correct.
- we also check if the first token is from identity provider
- we call `AcquireToken` second time and check if its from cache.
- We then create a new Managed identity of same type and check again if its called from cache.

The ` FakeValidate` we see here is that function that is used when we try to retrieve it from the cache, in reading process the storage checks if the token is valid or not. As, we have a fake token we don't that fails. To overcome that we provide a function that overrides that validation processes while testing. In the end we also check that was the validate function was called or not. This `FakeValidated` is what was used to check in other API's we have.

apps/managedidentity/managedidentity_test.go
```diff
@@ -180,13 +189,35 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
+				tokenValidation = true
+				return nil
+			}
+			cacheResult, err := client.AcquireToken(context.Background(), testCase.resource)
+			if cacheResult.Metadata.TokenSource != 2 {
+				t.Fatalf("wanted cache token source, got %d", result.Metadata.TokenSource)
+			}
+			if !tokenValidation {
+				t.Fatal("token validation didnt happen")
```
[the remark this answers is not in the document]
4gust: When the token is in cache there is a check that checks for the validity of the token, 
`Validate()` in storage.go as we don't have a real token here, we test that calling fake validity which only happens when fetching from cache.

AndyOHart (approved):
LGTM

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
chlowell: 🤔 the trouble is that the cache does instance discovery to figure out environment aliases and needs an HTTP client to do that. Obviously we don't want managed identity clients to request instance metadata in any case, so we'll just disable discovery (@4gust there's an option on `authority.Info`) and the cache won't use its client. But that doesn't help with this constructor.

Separating instance discovery from the cache might be ideal but that's major surgery. I suppose we could just leave this line as-is and comment that the cache never uses the client because instance discovery is always disabled. Taking a closer look, there's a less hacky solution: the outer argument here is `*oauth.Client`, so `storage.New(nil)` is valid. It's not perfect because we would need nil checks wherever the cache uses its client, but that's better than constructing an unused client.

In any case, we need a test that fails when Client sends an unexpected request.

apps/tests/devapps/managedidentity/serialized_cache.json
chlowell, line 1: Can we delete this file? Looks like it's never used, which makes sense because we test cache serialization elsewhere and managed identity isn't relevant to that

4gust: done

apps/managedidentity/managedidentity.go
```diff
@@ -206,6 +216,7 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
 	}
 	var r accesstokens.TokenResponse
 	err = json.Unmarshal(responseBytes, &r)
+	r.GrantedScopes.Slice = append(r.GrantedScopes.Slice, req.URL.Query().Get(resourceQuerryParameterName))
```
chlowell, line 219: I suppose this is a reasonable assumption when the response doesn't provide a value, but shouldn't we prefer the response's `resource`?

chlowell (changes requested):

apps/managedidentity/managedidentity.go
```diff
@@ -64,7 +68,15 @@ func (c UserAssignedClientID) value() string   { return string(c) }
...
 func SystemAssigned() ID {
-	return systemAssignedValue("")
+	return systemAssignedValue(systemAssignedManagedIdentity)
+}
+
+// cache never uses the client because instance discovery is always disabled.
+// so oauth can be nil here.
+var cacheManager *storage.Manager = storage.New(nil)
+
+func resetCache() {
```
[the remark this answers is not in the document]
4gust: There is a similar method in `storage` so I used similar approach as it.
I will write it inside test only so we don't need it here

apps/managedidentity/managedidentity.go
4gust: We discussed with @bgavrilMS about this, 
We don't read `resource` from response, and we assume that those will always be same.

apps/managedidentity/managedidentity.go
```diff
@@ -223,9 +234,41 @@ func (client Client) AcquireToken(ctx context.Context, resource string, options
...
 		return base.AuthResult{}, err
 	}
+
+	fakeAuthInfo, err := authority.NewInfoFromAuthorityURI("https://login.microsoftonline.com/managed_identity", false, true)
+	if err != nil {
+		return base.AuthResult{}, err
+	}
+
+	fakeAuthParams := authority.NewAuthParams(client.miType.value(), fakeAuthInfo)
+	storageTokenResponse, err := cacheManager.Read(ctx, fakeAuthParams)
```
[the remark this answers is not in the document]
4gust: We need the `fakeAuthParams` to be passed in the `authResultFromToken` so its outside of the if
If we have `if o.claims == ""` then we don't read but we still do save it.

apps/managedidentity/managedidentity_test.go
```diff
@@ -180,13 +189,35 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
 			}
 			if result.AccessToken != token {
 				t.Fatalf("wanted %q, got %q", token, result.AccessToken)
 			}
-
+			storage.FakeValidate = func(storage.AccessToken) error {
+				tokenValidation = true
+				return nil
+			}
+			cacheResult, err := client.AcquireToken(context.Background(), testCase.resource)
```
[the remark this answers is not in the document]
4gust: updated!

apps/managedidentity/managedidentity.go
```diff
@@ -64,7 +68,15 @@ func (c UserAssignedClientID) value() string   { return string(c) }
 func (o UserAssignedObjectID) value() string   { return string(o) }
 func (r UserAssignedResourceID) value() string { return string(r) }
 func SystemAssigned() ID {
-	return systemAssignedValue("")
+	return systemAssignedValue(systemAssignedManagedIdentity)
+}
+
+// cache never uses the client because instance discovery is always disabled.
+// so oauth can be nil here.
+var cacheManager *storage.Manager = storage.New(nil)
```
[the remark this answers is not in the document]
4gust: Updating using `sync.Once{}` and using a method to get the cacheManager

apps/tests/devapps/managedidentity/managedidentity_sample.go
```diff
@@ -8,27 +8,50 @@ import (
...
-	result, err := miSystemAssigned.AcquireToken(context.Background(), "https://management.azure.com/")
+	result, err := miSystemAssigned.AcquireToken(context.Background(), "https://management.azure.com")
 	if err != nil {
 		fmt.Println(err)
 	}
 	fmt.Println("token expire at : ", result.ExpiresOn)
+	fmt.Println("token source : ", result.Metadata.TokenSource)
+	miSystemAssigned1, err := mi.New(mi.SystemAssigned())
+	if err != nil {
+		fmt.Println(err)
```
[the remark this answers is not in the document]
4gust: done

apps/tests/devapps/managedidentity/managedidentity_sample.go
```diff
@@ -8,27 +8,50 @@ import (
 )
 
 func runIMDSSystemAssigned() {
+
 	miSystemAssigned, err := mi.New(mi.SystemAssigned())
 	if err != nil {
 		fmt.Println(err)
 	}
-	result, err := miSystemAssigned.AcquireToken(context.Background(), "https://management.azure.com/")
+	result, err := miSystemAssigned.AcquireToken(context.Background(), "https://management.azure.com")
```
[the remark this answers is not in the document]
4gust: updated

apps/tests/devapps/managedidentity/managedidentity_sample.go
```diff
@@ -8,27 +8,50 @@ import (
...
 	if err != nil {
 		fmt.Println(err)
 	}
 	fmt.Println("token expire at : ", result.ExpiresOn)
+	fmt.Println("token source : ", result.Metadata.TokenSource)
+	miSystemAssigned1, err := mi.New(mi.SystemAssigned())
+	if err != nil {
+		fmt.Println(err)
+	}
+	result1, err1 := miSystemAssigned1.AcquireToken(context.Background(), "https://management.azure.com")
```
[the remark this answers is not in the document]
4gust: updated

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
4gust: Removed the method `reset` and test does a separate reset.

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
chlowell: That doesn't address the apparent risk of panic

chlowell (changes requested):
`storage.New(nil)` is acceptable but you need to ensure `storage.Manager` never dereferences that pointer

apps/managedidentity/managedidentity_test.go
4gust: Removed the `fakeValidation` removed form test.

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
4gust: Updated, and added a nil check before using `cacheManager`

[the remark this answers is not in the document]
chlowell: I'm not worried about `cacheManager` being nil, although it doesn't hurt to check. Readers and static analysis can recognize it's impossible for `storage.New` to return nil.

What I'm harping on is that `storage.Manager` uses the pointer argument to `storage.New` and assumes it isn't nil. I don't expect readers or static analysis to recognize that as used in this package, `storage.Manager` never dereferences that pointer. So, `storage.Manager` must check whether the pointer is nil before dereferencing it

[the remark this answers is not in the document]
4gust: So basically, when we are creating a `storage.New(nil)`
the nil part is reference to a `Manager.requests` variable inside manager. 
Which is only used when calling `metadata` which we check before calling that is that `Manager.requests` is nil. 

Basically when `storage` is calling/using `Manager.request` we need to check weather its nil or not, as it's the only argument that is assigned from `New()` and there is not check for it being nil.

apps/managedidentity/managedidentity_test.go
```diff
@@ -146,17 +147,26 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
 			mockClient.AppendResponse(mock.WithHTTPStatusCode(http.StatusOK), mock.WithBody(responseBody), mock.WithCallback(func(r *http.Request) {
 				localUrl = r.URL
 			}))
+			// resetting cache
+			cacheManager = storage.New(nil)
```
chlowell, line 151: This affects the whole process--all other tests--so it's best to restore the original value after this test exits

```suggestion
			before := cacheManager
			defer func() { cacheManager = before }()
			cacheManager = storage.New(nil)
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -146,17 +147,26 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
 			client, err := New(testCase.miType, WithHTTPClient(&mockClient))
-
 			if err != nil {
 				t.Fatal(err)
 			}
 			result, err := client.AcquireToken(context.Background(), testCase.resource)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if localUrl == nil {
```
chlowell, line 160: same block as below, why not combine the two conditions with `||`

apps/internal/base/storage/storage.go
```diff
@@ -173,10 +173,10 @@ func (m *Manager) Write(authParameters authority.AuthParams, tokenResponse acces
 	environment := authParameters.AuthorityInfo.Host
 	realm := authParameters.AuthorityInfo.Tenant
 	clientID := authParameters.ClientID
+
```
chlowell, line 176: Can you please revert all the whitespace changes unrelated to the feature? They're just noise in the diff and will make history more difficult to navigate

apps/tests/devapps/managedidentity/managedidentity_sample.go
```diff
@@ -3,31 +3,46 @@ package main
...
+		log.Fatal(err)
 	}
-	result, err := miSystemAssigned.AcquireToken(context.Background(), "https://management.azure.com/")
+	result, err := miSystemAssigned.AcquireToken(context.TODO(), "https://management.azure.com")
 	if err != nil {
-		fmt.Println(err)
+		log.Fatal(err)
 	}
 	fmt.Println("token expire at : ", result.ExpiresOn)
+	fmt.Println("token source : ", result.Metadata.TokenSource)
```
chlowell, line 22: Should an application developer care about the `TokenSource`? Is `token source : 1` useful output?

chlowell (approved):

PR #519, opened 2024-10-11
Acquire Token for Azure Arc

AndyOHart:
# Acquire Token for Azure Arc 

## Description

This PR adds functionality for acquiring tokens for Azure Arc.
This includes the following:

- Acquiring token via SAMI
- Error handling for UAMI
- Environment detection
- Probing

## Changes

- Some general refactoring of names etc
- Adding required variables for Azure Arc
- Source retrieval
- Retrieval of endpoints via env vars
- Validation logic for env vars and checking for himds
- Handling of Azure Arc request, along with the mandatory 401 return and retry
- Some changes to tests

## Testing

- Added unit tests to verify logic for `GetSource`, `getAzureArcEnvironmentVariables`, `validateAzureArcEnvironment`, `fileExists` and `handleAzureArcResponse`
- Updates tests for `AcquireToken` to check for AzureArc


apps/managedidentity/managedidentity.go
```diff
@@ -129,24 +169,91 @@ func New(id ID, options ...ClientOption) (Client, error) {
...
+			return base.AuthResult{}, err
+		}
+
+		// need to perform preliminary request to retrieve the secret key challenge provided by the HIMDS service
+		// this is done when we get a 401 response, which will be handled by the response handler
+		tokenResponse, err = client.getTokenForRequest(ctx, req, client.handleAzureArcResponse)
+		if err != nil {
+			return base.AuthResult{}, err
+		}
+	default:
```
bgavrilMS, line 230: I think you should treat `DefaultToImds` like this, but the rest of the cases (AppService etc.) should throw some sort of "not implemented exception".

apps/managedidentity/managedidentity.go
```diff
@@ -173,17 +280,104 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+func createAzureArcAuthRequest(ctx context.Context, id ID, resource string, claims string) (*http.Request, error) {
+	identityEndpoint, _ := getAzureArcEnvironmentVariables()
+	var msiEndpoint *url.URL
+
+	if identityEndpoint == "" {
+		identityEndpoint = azureArcEndpoint
+	}
+
+	if _, ok := id.(systemAssignedValue); !ok {
+		return nil, errors.New("Azure Arc doesn't support specifying a user-assigned managed identity at runtime")
```
bgavrilMS, line 292:
```suggestion
		return nil, errors.New("Azure Arc doesn't support user assigned managed identities")
```

apps/managedidentity/managedidentity.go
```diff
@@ -173,17 +280,104 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+		return nil, fmt.Errorf("couldn't parse %q: %s", identityEndpoint, parseErr)
+	}
+
+	msiParameters := msiEndpoint.Query()
+	msiParameters.Set(apiVersionQueryParameterName, azureArcAPIVersion)
+	resource = strings.TrimSuffix(resource, "/.default")
+	msiParameters.Set(resourceQueryParameterName, resource)
+
+	if len(claims) > 0 {
+		msiParameters.Set("claims", claims)
```
bgavrilMS, line 307: Interesting. I wasn't aware of this.

apps/managedidentity/managedidentity.go
```diff
@@ -33,21 +36,53 @@ const (
...
+	MsiEndpointEnvVar                   = "MSI_ENDPOINT"
+	IdentityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
 )
 
+var supportedAzureArcPlatforms = map[string]string{
+	"windows": fmt.Sprintf("%s%s", os.Getenv("ProgramData"), windowsTokenPath),
+	"linux":   linuxSupportedPath,
+}
+
+var azureArcFileDetection = map[string]string{
```
bgavrilMS, line 79: nit: naming - azureArcOsToFileMap

apps/managedidentity/managedidentity.go
```diff
@@ -173,17 +280,104 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+}
+
+// GetEnvironmentVariables returns the identity and IMDS endpoints
+func getAzureArcEnvironmentVariables() (string, string) {
+	identityEndpoint := os.Getenv(IdentityEndpointEnvVar)
+	imdsEndpoint := os.Getenv(ArcIMDSEnvVar)
+
+	if identityEndpoint == "" || imdsEndpoint == "" {
+		platform := runtime.GOOS
+		fileDetectionPath, retrievedPath := azureArcFileDetection[platform]
```
bgavrilMS, line 326: Maybe add a comment on what happens on Mac.

apps/managedidentity/managedidentity.go
```diff
@@ -173,17 +280,104 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+			if _, err := os.Stat(fileDetectionPath); err == nil {
+				identityEndpoint = azureArcEndpoint
+				imdsEndpoint = himdsExecutableHelperString
+			}
+		}
+	}
+	return identityEndpoint, imdsEndpoint
+}
+
+func validateAzureArcEnvironment(identityEndpoint, imdsEndpoint string, platform string) bool {
```
bgavrilMS, line 338: Some of this logic is pretty complex and we're going to struggle with supportability since there are no logs. Let's please add a story to have some logging, which we can tackle for GA. 

CC @chlowell for an opinion

bgavrilMS (approved):

apps/managedidentity/managedidentity.go
```diff
@@ -33,21 +36,53 @@ const (
...
 	// IMDS
 	imdsEndpoint   = "http://169.254.169.254/metadata/identity/oauth2/token"
 	imdsAPIVersion = "2018-02-01"
+
+	// Azure Arc
+	azureArcEndpoint                  = "http://127.0.0.1:40342/metadata/identity/oauth2/token"
+	azureArcAPIVersion                = "2020-06-01"
+	azureArcFileExtension             = ".key"
+	azureArcMaxFileSizeBytes    int64 = 4096
+	linuxSupportedPath                = "/var/opt/azcmagent/tokens/"
```
neha-bhargava, line 59: nit: linuxTokenPath and linuxHimdsPath would make this more readable IMO.

apps/managedidentity/managedidentity.go
```diff
@@ -173,17 +280,104 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+
+	if identityEndpoint == "" {
+		identityEndpoint = azureArcEndpoint
+	}
+
+	if _, ok := id.(systemAssignedValue); !ok {
+		return nil, errors.New("Azure Arc doesn't support specifying a user-assigned managed identity at runtime")
+	}
+
+	identityEndpoint = strings.Replace(identityEndpoint, "localhost", "127.0.0.1", -1)
```
neha-bhargava, line 295: Why do you need to replace this string?

apps/managedidentity/managedidentity.go
neha-bhargava: I don't think we are passing claims to the MSI endpoints. When claims are present, currently we are just bypassing the cache.

apps/managedidentity/managedidentity.go
```diff
@@ -173,17 +280,104 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, msiEndpoint.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("error creating http request %s", err)
+	}
+	req.Header.Set(metaHTTPHeaderName, "true")
+	return req, nil
+}
+
+// GetEnvironmentVariables returns the identity and IMDS endpoints
+func getAzureArcEnvironmentVariables() (string, string) {
```
neha-bhargava, line 320: nit: name of the function is misleading. This tries to detect the azure arc env. Something like detectAzureArcEnv or validateAzureArcEnv maybe?

apps/managedidentity/managedidentity.go
```diff
@@ -173,17 +280,104 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+	imdsEndpoint := os.Getenv(ArcIMDSEnvVar)
+
+	if identityEndpoint == "" || imdsEndpoint == "" {
+		platform := runtime.GOOS
+		fileDetectionPath, retrievedPath := azureArcFileDetection[platform]
+
+		if retrievedPath {
+			if _, err := os.Stat(fileDetectionPath); err == nil {
+				identityEndpoint = azureArcEndpoint
+				imdsEndpoint = himdsExecutableHelperString
```
neha-bhargava, line 331: Why is this needed? Are you using this anywhere else after the env is determined?

apps/managedidentity/managedidentity.go
neha-bhargava, line 338: I am a bit confused with this method and the method above. `getAzureArcEnvironmentVariables` at line 320. I think either one is needed.

apps/managedidentity/managedidentity.go
```diff
@@ -204,28 +398,76 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
...
-		return base.AuthResult{}, err
-	}
-	tokenResponse, err := client.getTokenForRequest(req)
-	if err != nil {
-		return base.AuthResult{}, err
+		if len(wwwAuthenticateHeader) == 0 {
+			return accesstokens.TokenResponse{}, errors.New("response has no www-authenticate header")
+		}
+
+		var secretFilePath string
```
neha-bhargava, line 414: nit: it will be more readable to have the file path validation in another method.

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,95 +80,129 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+	testCases := []sourceTestData{
+		{name: "testAzureArcSystemAssigned", source: AzureArc, endpoint: imdsEndpoint, expectedSource: AzureArc, miType: SystemAssigned()},
+		{name: "testAzureArcUserClientAssigned", source: AzureArc, endpoint: imdsEndpoint, expectedSource: AzureArc, miType: UserAssignedClientID("clientId")},
+		{name: "testAzureArcUserResourceAssigned", source: AzureArc, endpoint: imdsEndpoint, expectedSource: AzureArc, miType: UserAssignedResourceID("resourceId")},
+		{name: "testAzureArcUserObjectAssigned", source: AzureArc, endpoint: imdsEndpoint, expectedSource: AzureArc, miType: UserAssignedObjectID("objectId")},
+		{name: "testDefaultToImds", source: DefaultToIMDS, endpoint: imdsEndpoint, expectedSource: DefaultToIMDS, miType: SystemAssigned()},
+		{name: "testDefaultToImdsEmptyEndpoint", source: DefaultToIMDS, endpoint: "", expectedSource: DefaultToIMDS, miType: SystemAssigned()},
+		{name: "testDefaultToImdsLinux", source: DefaultToIMDS, endpoint: imdsEndpoint, expectedSource: DefaultToIMDS, miType: SystemAssigned()},
+		{name: "testDefaultToImdsEmptyEndpointLinux", source: DefaultToIMDS, endpoint: "", expectedSource: DefaultToIMDS, miType: SystemAssigned()},
```
neha-bhargava, line 154: Missing test cases for default to imds with user assigned managed identity

apps/managedidentity/managedidentity_test.go
```diff
@@ -248,3 +376,369 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+
+	// Test case: file exists
+	if !fileExists(tmpFile.Name()) {
+		t.Errorf("expected file to exist, but it doesn't")
+	}
+
+	// Test case: file does not exist
+	nonExistentFilePath := tmpFile.Name() + "_nonexistent"
+	if fileExists(nonExistentFilePath) {
+		t.Errorf("expected file not to exist, but it does")
```
neha-bhargava, line 599: Can you explain this test case?

apps/managedidentity/managedidentity_test.go
```diff
@@ -248,3 +376,369 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+		},
+		{
+			name:           "Invalid secret file size",
+			statusCode:     http.StatusUnauthorized,
+			headers:        map[string]string{wwwAuthenticateHeaderName: "Basic realm=" + filepath.Join(testCaseFilePath, "secret.key")},
+			expectedError:  "invalid secret file size",
+			platform:       runtime.GOOS,
+			createMockFile: true,
+		},
+		// todo readd when we can figure out how to corrupt a file
```
neha-bhargava, line 679: I don't think it is possible to mock the scenarios where the file is found at the file path expected by azure arc env.

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
neha-bhargava: Yes this one is a msal client exception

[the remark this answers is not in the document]
chlowell: They return the same type, `error`. The difference is interpolation. `errors.New` takes a string while `fmt.Errorf` takes a string containing formatting verbs and arguments to interpolate into that string. So, `fmt.Errorf` exists to shorten lines like `errors.New(fmt.Sprintf("error number %d", num))`

apps/managedidentity/managedidentity.go
AndyOHart: Anything to change here or ok to resolve the conversation?

apps/managedidentity/managedidentity.go
AndyOHart: Created here: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3066894

chlowell: Logging needs a thoughtful end-to-end design because users, particularly Azure SDK, need to control MSAL's log output. I probably wouldn't block GA on it because MSAL doesn't log anything today and managed identity clients will return a string from the point of failure when anything goes wrong.

apps/managedidentity/managedidentity.go
chlowell: Right, something like
```
case DefaultToIMDS:
    // ... the current default case ...
default:
    return fmt.Errorf("unsupported source %q", client.source)
```
would be tidier

apps/managedidentity/managedidentity.go
AndyOHart: You mean something like 'Only Windows & Linux is supported by the install script, so running on Mac is not possible"

apps/managedidentity/managedidentity.go
chlowell: We shouldn't set this parameter if Arc doesn't expect it (to my knowledge, it doesn't)

apps/managedidentity/managedidentity.go
AndyOHart: I saw MSAL JS was doing this so I added it here: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/f0c5f0250741cfd673538ce51c19a345fdf9be75/lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts#L192

apps/managedidentity/managedidentity.go
```diff
@@ -204,28 +398,76 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
...
+
+		// split the header to get the secret file path
+		parts := strings.Split(wwwAuthenticateHeader, "Basic realm=")
+		if len(parts) > 1 {
+			secretFilePath = parts[1]
+		} else {
+			return accesstokens.TokenResponse{}, errors.New("basic realm= not found in the string")
```
chlowell, line 421: Errors about unexpected values are much more useful when they include the unexpected value

apps/managedidentity/managedidentity.go
AndyOHart: Hmm, well the function is doing that but it is also returning the identity and imds vars, maybe validateAndRetrieveAzureArcEnvVars?

apps/managedidentity/managedidentity.go
AndyOHart: This is another thing I saw in MSAL JS: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/f0c5f0250741cfd673538ce51c19a345fdf9be75/lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts#L49

apps/managedidentity/managedidentity.go
```diff
@@ -204,28 +398,76 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
...
+
+		// Attempt to read the contents of the secret file
+		secret, err := os.ReadFile(secretFilePath)
+		if err != nil {
+			return accesstokens.TokenResponse{}, errors.New("unable to read the secret file")
+		}
+
+		authHeaderValue := fmt.Sprintf("Basic %s", string(secret))
+
+		req, err := createAzureArcAuthRequest(ctx, SystemAssigned(), "https://management.azure.com", claims)
```
chlowell, line 462: Need the caller's resource here

apps/managedidentity/managedidentity.go
```diff
@@ -129,24 +169,91 @@ func New(id ID, options ...ClientOption) (Client, error) {
...
+		if err != nil {
+			return base.AuthResult{}, err
+		}
+	default:
+		req, err = createIMDSAuthRequest(ctx, client.miType, resource, o.claims)
+		if err != nil {
+			return base.AuthResult{}, err
+		}
+
+		tokenResponse, err = client.getTokenForRequest(ctx, req, nil)
```
chlowell, line 236: Consider using this switch just to get the token request, not to send it. Arc would be a little different in that you'd get the key in this block, but organizing the switch around creating the `http.Request` you're going to send would save a lot of repetition

apps/managedidentity/managedidentity.go
```diff
@@ -173,17 +280,104 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
 	responseBytes, err := io.ReadAll(resp.Body)
 	defer resp.Body.Close()
+
 	if err != nil {
 		return accesstokens.TokenResponse{}, err
 	}
+
 	switch resp.StatusCode {
+	case http.StatusUnauthorized:
+		if resHandler != nil {
```
chlowell, line 378: I suggest sending the key discovery request somewhere else and leaving `getTokenForRequest` entirely responsible for token requests. The complexity here doesn't seem worthwhile to me:
- "response handler" implies it applies to all responses but really it's an "authentication challenge handler"
- this looks like a general purpose mechanism but it applies only to Arc
- the callback for Arc recurses into this function; what happens when the server responds with another 401?

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,95 +80,129 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
 
-type resourceTestData struct {
-	source   Source
-	endpoint string
-	resource string
-	miType   ID
+func createMockFile(path string, size int64) {
+	dir := filepath.Dir(path)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		panic(err)
```
chlowell, line 94: `t.Fatal("...")`

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,95 +80,129 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+	}
+	f.Close()
 }
 
-type errorTestData struct {
-	code          int
-	err           string
-	desc          string
-	correlationid string
+func getMockFilePath() (string, error) {
```
chlowell, line 110: Looks like the real file path, on Linux anyway. You could refactor the implementation to make using arbitrary paths in tests simple, [for example](https://github.com/Azure/azure-sdk-for-go/blob/60b1b7ce748104aec8519bc4a8450b04baafc0ce/sdk/azidentity/managed_identity_credential_test.go#L160-L166)

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,95 +80,129 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
 
 	for _, testCase := range testCases {
-		t.Run(http.StatusText(testCase.code), func(t *testing.T) {
-			fakeErrorClient := mock.Client{}
-			responseBody, err := makeResponseWithErrorData(testCase.err, testCase.desc)
-			if err != nil {
-				t.Fatalf("error while forming json response : %s", err.Error())
+		t.Run(string(testCase.source), func(t *testing.T) {
+			unsetEnvVars()
+			setEnvVars(t, testCase.source)
```
chlowell, line 160: [t.Setenv()](https://pkg.go.dev/testing#T.Setenv)

chlowell (changes requested):
Blockers are
- apparent recursion in `getTokenForRequest`
- hardcoded resource
- invalid query parameter "claims"

apps/managedidentity/managedidentity.go
AndyOHart: Your right, refactored! Thanks

apps/managedidentity/managedidentity_test.go
AndyOHart: Testing the true and false return values of the fileExists method, testing if a file exists when it shouldnt or vice versa

apps/managedidentity/managedidentity_test.go
AndyOHart: Was thinking the same but thought i would leave this here incase anybody had an idea

apps/managedidentity/managedidentity.go
AndyOHart: Based on another comment I merged the 2 functions and renamed it to getAndValidateAzureArcEnvVars

apps/managedidentity/managedidentity.go
AndyOHart: Thanks, addressed it

apps/managedidentity/managedidentity_test.go
AndyOHart: Using the setEnvVars(t, testCase.source) instead as it sets based on the source, instead of having to set for each source twice

apps/managedidentity/managedidentity.go
neha-bhargava: The reason this is there in MSAL JS is because the flow is setup in a way where the env variables are confirmed again later in the flow to verify that it is arc env. I don't think you do that here. I think you can skip this.

apps/managedidentity/managedidentity.go
neha-bhargava: Hmm, if you get the endpoint from the env variable I think you should not alter that

apps/managedidentity/managedidentity.go
AndyOHart: Ok perfect will update thanks

4gust:
LGTM

4gust (approved):

apps/managedidentity/managedidentity.go
AndyOHart: Ok sounds good will update :)

apps/managedidentity/managedidentity_test.go
AndyOHart: Removed

apps/managedidentity/managedidentity_test.go
AndyOHart: Updated with what I think your saying

apps/managedidentity/managedidentity.go
```diff
@@ -204,28 +292,183 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
...
+	}
+
+	if identityEndpoint == "" || imdsEndpoint == "" {
+		platform := runtime.GOOS
+
+		// Only platforms inside the azureArcOsToFileMap are supported
+		fileDetectionPath, platformSupported := azureArcOsToFileMap[platform]
+
+		if platformSupported && fileExists(fileDetectionPath) {
+			identityEndpoint = azureArcEndpoint
```
neha-bhargava, line 389: returning identity endpoint from this method is all that you need. You can skip returning the imds endpoint. Since at this point you have confirmed that the env is Azure arc and imds endpoint is only used for that purpose.

apps/managedidentity/managedidentity.go
```diff
@@ -204,28 +292,183 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
...
+	}
+	req.Header.Set(metaHTTPHeaderName, "true")
+	return req, nil
+}
+
+func createAzureArcAuthRequest(ctx context.Context, id ID, resource string) (*http.Request, error) {
+	identityEndpoint, _, _ := getAndValidateAzureArcEnvVars()
+	var msiEndpoint *url.URL
+
+	if identityEndpoint == "" {
```
neha-bhargava, line 345: This will never be empty, since the method getAndValidateAzureArcEnvVars will return it

apps/managedidentity/managedidentity.go
```diff
@@ -105,6 +140,11 @@ func WithHTTPClient(httpClient ops.HTTPClient) ClientOption {
 //
 // Options: [WithHTTPClient]
 func New(id ID, options ...ClientOption) (Client, error) {
+	source, err := GetSource(id)
```
neha-bhargava, line 143: What is the scope of this source? Is it for the app level or request level?

apps/managedidentity/managedidentity.go
```diff
@@ -204,28 +292,183 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
...
 	if err != nil {
-		return base.AuthResult{}, err
+		return nil, fmt.Errorf("error creating http request %s", err)
+	}
+	req.Header.Set(metaHTTPHeaderName, "true")
+	return req, nil
+}
+
+func createAzureArcAuthRequest(ctx context.Context, id ID, resource string) (*http.Request, error) {
+	identityEndpoint, _, _ := getAndValidateAzureArcEnvVars()
```
neha-bhargava, line 342: I don't think you need to call this method again. You can simply use the azure arc endpoint since it is confirmed that the source is arc.

neha-bhargava: The flow in Go is different from how it is .Net and JS there can be some internal implementation differences I guess

neha-bhargava (commented):
Main comment is to revisit the arc source detection and acquire token flow. This area seems to need some refactoring.

apps/managedidentity/managedidentity.go
AndyOHart: This is to actually get the endpoint as in the GetSource it is more to just get what source we want, we don't set the endpoint there, while here we are getting the endpoint to use

apps/managedidentity/managedidentity.go
AndyOHart: It is the equivalent of the .NET implementation:

```
        public ManagedIdentityClient(RequestContext requestContext)
        {
            using (requestContext.Logger.LogMethodDuration())
            {
                _identitySource = SelectManagedIdentitySource(requestContext);
            }
        }
```

For getting it when we create the ManagedIdentityClient

apps/managedidentity/managedidentity.go
AndyOHart: This is to actually get the endpoint as in the GetSource it is more to just get what source we want, we don't set the endpoint there, while here we are getting the endpoint to use

apps/managedidentity/managedidentity.go
AndyOHart: Nice spot thanks! Updated!

apps/managedidentity/managedidentity.go
```diff
@@ -129,60 +168,108 @@ func New(id ID, options ...ClientOption) (Client, error) {
...
-		msiParameters.Set(miQueryParameterResourceId, string(t))
-	case UserAssignedObjectID:
-		msiParameters.Set(miQueryParameterObjectId, string(t))
-	case systemAssignedValue: // not adding anything
+		// need to perform preliminary request to retrieve the secret key challenge provided by the HIMDS service
+		// this is done when we get a 401 response, which will be handled by the response handler
+		tokenResponse, err = client.getTokenForRequest(ctx, req)
+		if err != nil {
+			switch callErr := err.(type) {
+			case errors.CallErr:
```
chlowell, line 227: It's strange to implement the expected success case as error handling. I believe a more straightforward approach [like this](https://github.com/Azure/azure-sdk-for-go/blob/7a626d2e848f739e4f6f94b275f1f1b2eeb56300/sdk/azidentity/managed_identity_client.go#L425) would be easier to understand and maintain.

If you want to keep this approach, please at least use [errors.As](https://pkg.go.dev/errors#As) to make it more robust

apps/managedidentity/managedidentity_test.go
```diff
@@ -16,21 +19,49 @@ import (
...
 type SuccessfulResponse struct {
 	AccessToken string `json:"access_token"`
 	ExpiresOn   int64  `json:"expires_on"`
 	Resource    string `json:"resource"`
 	TokenType   string `json:"token_type"`
 }
 
-type ErrorRespone struct {
+// Mock fileExists function for testing
+var mockFileExists = func() bool {
```
chlowell, line 60: unused

apps/managedidentity/managedidentity_test.go
```diff
@@ -16,21 +19,49 @@ import (
...
-	// test Resources
-	resource              = "https://demo.azure.com"
-	resourceDefaultSuffix = "https://demo.azure.com/.default"
-
-	token = "fakeToken"
+	// Test Resources
+	resource              = "https://management.azure.com"
+	resourceDefaultSuffix = "https://management.azure.com/.default"
+	token                 = "fakeToken"
+	azureArcTestEndpoint  = "http://127.0.0.1:40342/metadata/identity/oauth2/token"
```
chlowell, line 26: unused

apps/managedidentity/managedidentity.go
```diff
@@ -204,28 +291,177 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
...
-// Acquires tokens from the configured managed identity on an azure resource.
-//
-// Resource: scopes application is requesting access to
-// Options: [WithClaims]
-func (client Client) AcquireToken(ctx context.Context, resource string, options ...AcquireTokenOption) (base.AuthResult, error) {
-	o := AcquireTokenOptions{}
+func fileExists(path string) bool {
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			return false
```
chlowell, line 302: If we land here, the path's existence is undetermined. The error means that `Stat` failed for some reason other than the path's nonexistence. Since the path's existence isn't important at the time of calling this function, I suggest instead checking the key file path while handling the authentication challenge. We have to do that anyway, and at that time can't proceed without the file and will return an error if it isn't readable for any reason

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,95 +79,130 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+	case CloudShell:
+		t.Setenv(MsiEndpointEnvVar, "msiEndpointEnvVar value")
+	case ServiceFabric:
+		t.Setenv(IdentityEndpointEnvVar, "identityEndpointEnvVar value")
+		t.Setenv(IdentityHeaderEnvVar, "identityHeaderEnvVar value")
+		t.Setenv(IdentityServerThumbprintEnvVar, "identityServerThumbprintEnvVar value")
+	}
+}
+
+func unsetEnvVars() {
```
chlowell, line 133: `t.Setenv("VARIABLE", "")` is better because it has no side effect on other tests

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,95 +79,130 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
-			responseBody, err := makeResponseWithErrorData(testCase.err, testCase.desc)
-			if err != nil {
-				t.Fatalf("error while forming json response : %s", err.Error())
+		t.Run(string(testCase.source), func(t *testing.T) {
+			unsetEnvVars()
+			setEnvVars(t, testCase.source)
+
+			if runtime.GOOS == "linux" {
+				originalPath := azureArcOsToFileMap[runtime.GOOS]
+				azureArcOsToFileMap[runtime.GOOS] = "fake/fake"
```
chlowell, line 164: Generally best not to modify global state in a test and when you must, to restore the prior value after the test

chlowell (changes requested):

apps/managedidentity/managedidentity_test.go
AndyOHart: Used it across tests now and added azureArcImdsEndpoint

apps/managedidentity/managedidentity.go
```diff
@@ -204,28 +291,177 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
...
-		return base.AuthResult{}, err
+		return nil, fmt.Errorf("error creating http request %s", err)
 	}
-	return base.NewAuthResult(tokenResponse, shared.Account{})
+	req.Header.Set(metaHTTPHeaderName, "true")
+	return req, nil
+}
+
+// GetEnvironmentVariables returns the identity and IMDS endpoints
+func getAndValidateAzureArcEnvVars() (string, bool) {
```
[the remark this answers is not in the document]
AndyOHart: I think Neha had a comment that looked like we had 2 methods that were doing the same thing, the validation part, similar to .NET that is used in the GetSource func called ValidateAzureArcEnvironment function
So i extracted that logic into the getAzureArcEndpoint() call, I am going to revert back and have validateAzureArcEnvironment for just the GetSource and then for the AzureRequest we can get the correct endpoint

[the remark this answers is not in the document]
neha-bhargava: I can try to answer this
1. At this point we are trying to validate that the code is running on an azure arc env. If the env is azure arc the expectation is to find a combination of those 2 env variables. But in some env like linux the env variables are not detected by the process so we check the presence of file and declare that the env is azure arc. 
2. Validate when both are empty because if either is set and not both that means it is some other Azure env where the code is running and not Azure arc. Unfortunately, looking at env variables combination presence is the way to detect if managed identity is available and what Azure env it is.
3. Because while GetSource method only cares about detecting the env, createAzureArcAuthRequest knows that the env is azure arc and only looks for the returned endpoint. 

IMO this can be simplified by not calling this method in createAzureArcAuthRequest and just use the arc endpoint.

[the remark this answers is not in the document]
AndyOHart: Removed second function, kept the validateAzureArcEnvironment function
Request just uses the default arc endpoint now since we know its azure arc

apps/managedidentity/managedidentity.go
```diff
@@ -204,28 +291,177 @@ func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRe
...
+	req.Header.Set(metaHTTPHeaderName, "true")
+	return req, nil
+}
+
+func createAzureArcAuthRequest(ctx context.Context, id ID, resource string) (*http.Request, error) {
+	identityEndpoint, _ := getAndValidateAzureArcEnvVars()
+	var msiEndpoint *url.URL
+
+	if _, ok := id.(systemAssignedValue); !ok {
+		return nil, errors.New("Azure Arc doesn't support user assigned managed identities")
```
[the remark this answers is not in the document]
AndyOHart: Could you clarify what you mean? its using errors.New so not sure exactly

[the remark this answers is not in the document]
bgavrilMS: Shouldn't this be `fmt.Errorf("Azure Arc ...")`?

apps/managedidentity/managedidentity.go
AndyOHart: Opted to use errors.As

apps/managedidentity/managedidentity.go
AndyOHart: Going based on the .NET implementation, we need to check the file path for validateAzureArcEnvironment in GetSource to know if we have HIMDS or not, the key file path checking is also being done in the authentication challenge before we get file size

apps/tests/devapps/managedidentity/docs/azure_arc_linux.md
```diff
@@ -0,0 +1,204 @@
+# Setting Up Azure Arc on Ubuntu(linux)
```
neha-bhargava, line 1: This is great :)

neha-bhargava (approved):

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,122 +75,178 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+	f, err := os.Create(path)
+	if err != nil {
+		t.Fatalf("failed to create file: %v", err)
+	}
+	if size > 0 {
+		if err := f.Truncate(size); err != nil {
+			t.Fatalf("failed to truncate file: %v", err)
+		}
+	}
+	f.Close()
```
chlowell, line 100: As written, this file handle leaks when `f.Truncate` returns an error. You need to `defer` this after `os.Create`

apps/managedidentity/managedidentity_test.go
```diff
@@ -16,21 +19,45 @@ import (
...
-	resource              = "https://demo.azure.com"
-	resourceDefaultSuffix = "https://demo.azure.com/.default"
-
-	token = "fakeToken"
+	// Test Resources
+	resource                 = "https://management.azure.com"
+	resourceDefaultSuffix    = "https://management.azure.com/.default"
+	token                    = "fakeToken"
+	azureArcIdentityEndpoint = "http://127.0.0.1:40342/metadata/identity/oauth2/token"
+	azureArcImdsEndpoint     = "http://127.0.0.1:40342"
```
chlowell, line 27: unused

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,122 +75,178 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
-		{code: http.StatusBadRequest,
-			err:           "invalid_request",
-			desc:          "Identity not found",
-			correlationid: "121212",
-		},
+func setEnvVars(t *testing.T, source Source) {
+	switch source {
+	case AzureArc:
+		t.Setenv(identityEndpointEnvVar, azureArcIdentityEndpoint)
+		t.Setenv(imdsEndVar, "http://127.0.0.1:40342 value")
```
chlowell, line 113: shouldn't this and the other endpoint values be a valid URL?

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,122 +75,178 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+	getAzureArcPlatformPath = func(platform string) string {
+		return path
+	}
+
+	return func() { getAzureArcPlatformPath = originalFunc }
+}
+
+func setCustomAzureArcFilePath(path string) func() {
+	originalFunc := getAzureArcFilePath
+	getAzureArcFilePath = func(platform string) string {
```
chlowell, line 145: JFYI, you don't need to name an unused argument
```suggestion
	getAzureArcFilePath = func(string) string {
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,122 +75,178 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
-		t.Run(http.StatusText(testCase.code), func(t *testing.T) {
-			fakeErrorClient := mock.Client{}
-			responseBody, err := makeResponseWithErrorData(testCase.err, testCase.desc)
-			if err != nil {
-				t.Fatalf("error while forming json response : %s", err.Error())
+		t.Run(string(testCase.source), func(t *testing.T) {
+			unsetEnvVars(t)
+			setEnvVars(t, testCase.source)
+
+			if runtime.GOOS == "linux" {
```
chlowell, line 173: What about `windows`?

apps/managedidentity/managedidentity_test.go
```diff
@@ -248,3 +387,205 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+				StatusCode: tc.statusCode,
+				Header:     make(http.Header),
+			}
+
+			for k, v := range tc.headers {
+				response.Header.Set(k, v)
+			}
+
+			if tc.createMockFile {
+				expectedFilePath := filepath.Join(testCaseFilePath)
```
chlowell, line 563: `Join` with fewer than 2 arguments is a no-op
```suggestion
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -248,3 +387,205 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+					createMockFile(t, mockFilePath, 0)
+				}
+
+				defer os.Remove(mockFilePath)
+			}
+
+			contextToUse := context.Background()
+			client := &Client{}
+
+			if tc.name == "token request fail" {
```
chlowell, line 580: This is fragile. Parameterize the context or test this behavior separately (what do you want to test?)

apps/managedidentity/managedidentity_test.go
```diff
@@ -48,122 +75,178 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
-			client, err := New(testCase.miType, WithHTTPClient(&mockClient))
 
+			client, err := New(testCase.miType, WithHTTPClient(&mockClient))
 			if err != nil {
 				t.Fatal(err)
 			}
+
 			result, err := client.AcquireToken(context.Background(), testCase.resource)
+			if err != nil {
+				if testCase.source == AzureArc && err.Error() == "azure Arc doesn't support user assigned managed identities" {
```
chlowell, line 232: This block will go away when `New` returns the expected error, but take a closer look at it anyway. What happens when this condition is false?

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
chlowell: Yes. If you have a constant error message, call `errors.New`. `fmt.Errorf` is for interpolating values into a string containing formatting verbs like `%s`.

But the `New` I was referring to is the constructor in this package, `managedidentity.New`. That function must return this error when an application running on Arc specifies a user-assigned identity. It can't wait until the application calls `AcquireToken`

chlowell (changes requested):

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
AndyOHart: OK thanks changed

apps/managedidentity/managedidentity_test.go
AndyOHart: Removing this

apps/managedidentity/managedidentity_test.go
AndyOHart: Was for testing that we get an error back from the handleAzureArcResponse at the end of the flow. Changed to be parameterized

chlowell: Why does that require a nil Context? Also, the fragile part is branching on the test name

chlowell (changes requested):
Blocker is the failing test

apps/managedidentity/managedidentity_test.go
```diff
@@ -50,112 +61,326 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+		endpoint       string
+		expectedSource Source
+		miType         ID
+	}{
+		{name: "testAzureArcSystemAssigned", source: AzureArc, endpoint: "imdsDefaultEndpoint", expectedSource: AzureArc, miType: SystemAssigned()},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(string(testCase.source), func(t *testing.T) {
+			unsetEnvVars(t)
```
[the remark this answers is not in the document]
AndyOHart: This test is passing for me locally and on linux / windows vms. What is going wrong for you?
The test case doesn't need env its just to check that we return true for getting himds when on endpoints are sent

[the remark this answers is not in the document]
chlowell: But it doesn't set IDENTITY_ENDPOINT or IMDS_ENDPOINT or fake the HIMDS path. I'm surprised it passes on a machine not managed by Arc. It fails like this on my machine:
```
=== FAIL: apps/managedidentity TestAzureArcReturnsWhenHimdsFound/AzureArc (0.00s)
    managedidentity_test.go:205: expected AzureArc, got DefaultToIMDS
```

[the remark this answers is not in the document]
AndyOHart: Sorry I meant to say its passing on linux and windows VMs, not locally since I'm on Mac.
I guess it not setting identityEndpoint and imdsEndpoint is ok since the first if is checking if they are not empty to return true, otherwise continue on. I'm guessing the VMs have himds in whatever location its checking and thats why, let me look more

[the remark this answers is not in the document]
AndyOHart: Can you try again, should work now

apps/managedidentity/managedidentity_test.go
AndyOHart: does not apply anymore

apps/managedidentity/managedidentity_test.go
```diff
@@ -274,3 +588,234 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+
+			result := isAzureArcEnvironment(tc.identityEndpoint, tc.imdsEndpoint, tc.platform)
+			if result != tc.expectedResult {
+				t.Fatalf(errorExpectedButGot, tc.expectedResult, result)
+			}
+		})
+	}
+}
+
+func TestHandleAzureArcResponse(t *testing.T) {
```
chlowell, line 704: This actually tests a different function

apps/managedidentity/managedidentity_test.go
```diff
@@ -274,3 +588,234 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+			platform:         runtime.GOOS,
+			expectedResult:   false,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			setCustomAzureArcFilePath(t, fakeAzureArcFilePath)
+
+			result := isAzureArcEnvironment(tc.identityEndpoint, tc.imdsEndpoint, tc.platform)
```
chlowell, line 696: would be better to test this through public API i.e. `GetSource`--more durable, covers the same code--as in `TestAzureArcReturnsWhenHimdsFound`. Actually, looks like this test should merge with that one

chlowell (changes requested):

apps/managedidentity/managedidentity_test.go
```diff
@@ -50,112 +61,345 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+		t.Run(string(testCase.source), func(t *testing.T) {
+			unsetEnvVars(t)
+
+			// Get system dependent mock file path
+			var mockFilePath string
+			if runtime.GOOS == "windows" {
+				mockFilePath = filepath.Join(os.TempDir(), "himds.exe")
+			} else {
+				mockFilePath = filepath.Join("/tmp", "himds")
+			}
```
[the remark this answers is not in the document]
bgavrilMS: This is the new guidance we got from AzureArc. The env variable based detection is not great on Linux. So they now recommend to use file based detection, i.e. detect the arc agent via file check. The agent's internal files are differnt on Win vs Linux.

Note that this is not an improvement that Azure SDK have made afaik, but one that was flagged to MSAL team.

@4gust can point you at the spec.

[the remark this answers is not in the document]
4gust: Its needed as Windows has `.exe` at the end

apps/managedidentity/managedidentity_test.go
```diff
@@ -50,112 +61,347 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+		{source: DefaultToIMDS, endpoint: imdsDefaultEndpoint, resource: resourceDefaultSuffix, miType: SystemAssigned(), apiVersion: imdsAPIVersion},
+		{source: DefaultToIMDS, endpoint: imdsDefaultEndpoint, resource: resource, miType: UserAssignedClientID("clientId"), apiVersion: imdsAPIVersion},
+		{source: DefaultToIMDS, endpoint: imdsDefaultEndpoint, resource: resourceDefaultSuffix, miType: UserAssignedResourceID("resourceId"), apiVersion: imdsAPIVersion},
+		{source: DefaultToIMDS, endpoint: imdsDefaultEndpoint, resource: resourceDefaultSuffix, miType: UserAssignedObjectID("objectId"), apiVersion: imdsAPIVersion},
+	}
+	for _, testCase := range testCases {
+		t.Run(string(testCase.source)+"-"+testCase.miType.value(), func(t *testing.T) {
+			unsetEnvVars(t)
+			setEnvVars(t, testCase.source)
+			setCustomAzureArcFilePath(t, fakeAzureArcFilePath)
```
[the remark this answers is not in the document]
4gust: It does not run on VM,
need some modification on this.

apps/managedidentity/managedidentity_test.go
```diff
@@ -50,112 +61,345 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+				t.Errorf(errorExpectedButGot, testCase.expectedSource, actualSource)
+			}
+		})
+	}
+}
+
+func TestAzureArcReturnsWhenHimdsFound(t *testing.T) {
+	if runtime.GOOS == "darwin" {
+		t.Skip("Skipping test on macOS as HIMDS is not supported")
+	}
```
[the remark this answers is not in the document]
4gust: Actually this test cannot run on my machine, I have a Mac for work. 
When I run all tests this will fails when it does not support Mac
thats why this condition.

apps/managedidentity/managedidentity_test.go
```diff
@@ -274,3 +588,234 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+			expectedError:  "invalid secret file size, expected 4096, file size was 5000",
+			platform:       runtime.GOOS,
+			createMockFile: true,
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.platform+" "+tc.name, func(t *testing.T) {
+			if tc.platform != "linux" && tc.platform != "windows" && tc.platform != "testPlatform" {
+				t.Skip("Skipping test because current platform is not linux or windows")
```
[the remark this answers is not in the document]
4gust: When running tests locally on a Mac, the `runtime.GOOS` will return Darwin. In this case, we do not create the mock file on macOS systems. This distinction is important to ensure our testing environment behaves correctly across different platforms.

apps/managedidentity/managedidentity_test.go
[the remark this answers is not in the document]
chlowell: Why? This test replaces the platform-dependent real implementation with a fake one returning the path to a file the test itself creates.

apps/managedidentity/managedidentity_test.go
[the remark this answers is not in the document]
chlowell: How does it fail?

apps/managedidentity/managedidentity_test.go
```diff
@@ -50,129 +62,447 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+	}
+
+	t.Cleanup(func() { getAzureArcFilePath = originalFunc })
+}
+
+func TestGetSource(t *testing.T) {
+	testCases := []struct {
+		name           string
+		source         Source
+		expectedSource Source
```
chlowell, line 149: These fields always have equal values. Do you really need both?

apps/managedidentity/managedidentity_test.go
```diff
@@ -50,112 +61,347 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
-		{source: DefaultToIMDS, endpoint: imdsEndpoint, resource: resource, miType: UserAssignedClientID("clientId")},
-		{source: DefaultToIMDS, endpoint: imdsEndpoint, resource: resourceDefaultSuffix, miType: UserAssignedResourceID("resourceId")},
-		{source: DefaultToIMDS, endpoint: imdsEndpoint, resource: resourceDefaultSuffix, miType: UserAssignedObjectID("objectId")},
+func TestAzureArcAcquireTokenReturnsTokenSuccess(t *testing.T) {
+	testCaseFilePath, err := getMockFilePath(t)
+	if err != nil {
+		t.Fatalf("failed to get mock file path: %v", err)
+	}
+
+	testCases := []struct {
```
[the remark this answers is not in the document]
chlowell: Still some issues here:
- all cases have the same value for `miType` and `apiVersion`
- `expectedError` serves the same function as `shouldFail` and is more useful
- this test is all about errors: all but one case expects one. Should that one case be a separate test?

apps/managedidentity/managedidentity_test.go
```diff
@@ -50,129 +62,461 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
+	if actualSource != AzureArc {
+		t.Errorf(errorExpectedButGot, AzureArc, actualSource)
+	}
+}
+
+func TestIMDSAcquireTokenReturnsTokenSuccess(t *testing.T) {
+	testCases := []struct {
+		resource   string
+		miType     ID
+		apiVersion string
```
chlowell, line 253: unused

apps/managedidentity/managedidentity_test.go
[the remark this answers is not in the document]
4gust: I removed the is platform condition from here. Comment is older and is now pointing on the IMDS Test case, will resolve this.

apps/managedidentity/managedidentity_test.go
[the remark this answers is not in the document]
4gust: fixed.

apps/managedidentity/managedidentity_test.go
4gust: Deleted this test

apps/managedidentity/managedidentity_test.go
4gust: Removed.

apps/managedidentity/managedidentity_test.go
4gust: Removed

apps/managedidentity/managedidentity_test.go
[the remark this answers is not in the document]
chlowell: Please simplify this. The test is fragile and too complex:
* two "Platform not supported" cases, one of which actually tests something else
* `miType` is `SystemAssigned()` in all but one case
    * that one case doesn't cover everything it should but is redundant anyway because we already have `TestAzureArcUserAssignedFailure` covering all `miType` related errors
* `platform` is `runtime.GOOS` in all but one case
* `shouldFail` is false in only one case
* `expectedError` serves the same function as `shouldFail` and is more useful
* the test body hardcodes test case parameter values

Part of the problem is that the test tries to do too much. It would be better to write separate, more focused tests. How about this:
* `TestAzureArc`: success; client acquires token. Cases exercise source detection by environment variable and by HIMDS path
* `TestAzureArcErrors`: AcquireToken error cases for Arc. Cases cover secret file errors: invalid path, missing file, invalid size

apps/managedidentity/managedidentity_test.go
```diff
@@ -217,6 +281,293 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
+	setEnvVars(t, AzureArc)
+	mockClient := mock.Client{}
+
+	setCustomAzureArcFilePath(t, fakeAzureArcFilePath)
+	for _, testCase := range []ID{
+		UserAssignedClientID("client"),
+		UserAssignedObjectID("ObjectId"),
+		UserAssignedResourceID("resourceid")} {
+		_, err := New(testCase, WithHTTPClient(&mockClient))
+		if err != nil {
```
chlowell, line 383: just a thought for simplifying this; remember `t.Fatal()` is tantamount to `return`
```go
if err == nil {
	t.Fatal(`expected error: azure arc not supported error"`)
}
if err.Error() != "azure Arc doesn't support user assigned managed identities" {
	t.Fatalf(`expected error: azure arc not supported error, got error: "%v"`, err)
}
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -50,129 +61,182 @@ func getSuccessfulResponse(resource string) ([]byte, error) {
...
-		t.Run(http.StatusText(testCase.code), func(t *testing.T) {
-			fakeErrorClient := mock.Client{}
-			responseBody, err := makeResponseWithErrorData(testCase.err, testCase.desc)
-			if err != nil {
-				t.Fatalf("error while forming json response : %s", err.Error())
-			}
-			fakeErrorClient.AppendResponse(mock.WithHTTPStatusCode(testCase.code),
-				mock.WithBody(responseBody))
-			client, err := New(SystemAssigned(), WithHTTPClient(&fakeErrorClient))
+func unsetEnvVars(t *testing.T) {
```
chlowell, line 119: If all tests that set environment variables do so via `t.Setenv` (looks like they do), you don't need this function

apps/managedidentity/managedidentity_test.go
```diff
@@ -217,6 +281,293 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
+			}
+
+			mockClient.AppendResponse(mock.WithHTTPStatusCode(http.StatusOK), mock.WithHTTPHeader(headers),
+				mock.WithBody(responseBody))
+
+			// resetting cache
+			before := cacheManager
+			defer func() { cacheManager = before }()
+			cacheManager = storage.New(nil)
+
```
chlowell, line 489:
```suggestion
```
1. you don't need a token response because you're testing that the client returns an error before sending a token request
2. you don't need to reset the cache because this test doesn't write to the cache, and tests that do write to the cache reset it

apps/managedidentity/managedidentity_test.go
```diff
@@ -217,6 +281,293 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
+func TestSystemAssignedReturnsAcquireTokenFailure(t *testing.T) {
+	testCases := []struct {
+		code          int
+		err           string
+		desc          string
+		correlationID string
+	}{
+		{code: http.StatusNotFound,
+			err:           "",
+			desc:          "",
```
chlowell, line 516:
```suggestion
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -217,6 +281,293 @@ func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {
...
+		})
+	}
+}
+
+func TestSystemAssignedReturnsAcquireTokenFailure(t *testing.T) {
+	testCases := []struct {
+		code          int
+		err           string
+		desc          string
+		correlationID string
```
chlowell, line 512: unused

apps/managedidentity/managedidentity_test.go
```diff
@@ -274,3 +628,22 @@ func TestCreatingIMDSClient(t *testing.T) {
 		})
 	}
 }
+
+func TestAzureArcUserAssignedFailure(t *testing.T) {
```
chlowell, line 632: duplicates `TestAzureArcOnlySystemAssignedSupported`

chlowell (changes requested):
Blocker is respecting `IDENTITY_ENDPOINT`

apps/managedidentity/managedidentity_test.go
4gust: Point 2 make sense.
Point 1 the error that we test are for acquire token errors, as the first request will always go give `http.StatusUnauthorized` which requires the `client` to be available

PR #521, opened 2024-10-21
Implemented Retry Policy

4gust:
# Pull Request: Implement Retry Policy for Client

## Overview

This pull request introduces a retry policy for network calls in the client. The retry policy aims to enhance the resilience of our application by allowing it to automatically retry failed requests up to a maximum of three times.

## Changes Made

- **Retry Policy Implementation**: 
  - A retry mechanism has been added to handle transient errors in network calls.
  - The default maximum retry count is set to **3**.

- **Configuration Option to Disable Retry Policy**:
  - Introduced an optional method `WithRetryPolicyDisabled()` for the client.
  - When this method is used, the maximum retry count is reduced to **1** (effectively disabling the retry mechanism).

### Default Behavior (Retries Enabled)

By default, the client will automatically retry failed requests up to **3** times with delay of **1** second. Here’s how to create the client with the retry policy enabled:

```go
client, err := mi.New(SystemAssigned())
if err != nil {
    // Handle error
}
```
### Default Behavior (Retries Disabled)

By default, the client will not retry the failed request. Here’s how to create the client with the retry policy disabled:
```go
client, err := mi.New(SystemAssigned(), WithRetryPolicyDisabled())
if err != nil {
    // Handle error
}
```
### Error List when retry is done.
```go
	http.StatusNotFound,                      // 404
	http.StatusGone,                          // 410
	http.StatusTooManyRequests,               // 429 // retry after.
	http.StatusInternalServerError,           // 500
	http.StatusNotImplemented,                // 501
	http.StatusBadGateway,                    // 502
	http.StatusServiceUnavailable,            // 503
	http.StatusGatewayTimeout,                // 504
	http.StatusHTTPVersionNotSupported,       // 505
	http.StatusVariantAlsoNegotiates,         // 506
	http.StatusInsufficientStorage,           // 507
	http.StatusLoopDetected,                  // 508
	http.StatusNotExtended,                   // 510
	http.StatusNetworkAuthenticationRequired, // 511
```

### DIFF
--- a/apps/managedidentity/managedidentity.go
+++ b/apps/managedidentity/managedidentity.go
@@ -16,7 +16,9 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"slices"
 	"strings"
+	"time"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/errors"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base"
@@ -50,7 +52,27 @@ const (
 	imdsAPIVersion = "2018-02-01"
 
 	systemAssignedManagedIdentity = "system_assigned_managed_identity"
+	defaultRetryCount             = 3
 )
+
+// IMDS docs recommend retrying 404, 410, 429 and 5xx
+// https://learn.microsoft.com/entra/identity/managed-identities-azure-resources/how-to-use-vm-token#error-handling
+var retryStatusCodes = []int{
+	http.StatusNotFound,                      // 404
+	http.StatusGone,                          // 410
+	http.StatusTooManyRequests,               // 429 // retry after.
+	http.StatusInternalServerError,           // 500
+	http.StatusNotImplemented,                // 501
+	http.StatusBadGateway,                    // 502
+	http.StatusServiceUnavailable,            // 503
+	http.StatusGatewayTimeout,                // 504
+	http.StatusHTTPVersionNotSupported,       // 505
+	http.StatusVariantAlsoNegotiates,         // 506
+	http.StatusInsufficientStorage,           // 507
+	http.StatusLoopDetected,                  // 508
+	http.StatusNotExtended,                   // 510
+	http.StatusNetworkAuthenticationRequired, // 511
+}
 
 type Source string
 
@@ -75,13 +97,14 @@ func SystemAssigned() ID {
 var cacheManager *storage.Manager = storage.New(nil)
 
 type Client struct {
-	httpClient ops.HTTPClient
-	miType     ID
-	// source     Source reenable when required in future sources
+	httpClient          ops.HTTPClient
+	miType              ID
+	retryPolicyDisabled bool
 }
 
 type ClientOptions struct {
-	httpClient ops.HTTPClient
+	httpClient         ops.HTTPClient
+	retryPolicyDiabled bool
 }
 
 type AcquireTokenOptions struct {
@@ -107,13 +130,20 @@ func WithHTTPClient(httpClient ops.HTTPClient) ClientOption {
 	}
 }
 
+func WithRetryPolicyDisabled() ClientOption {
+	return func(o *ClientOptions) {
+		o.retryPolicyDiabled = true
+	}
+}
+
 // Client to be used to acquire tokens for managed identity.
 // ID: [SystemAssigned], [UserAssignedClientID], [UserAssignedResourceID], [UserAssignedObjectID]
 //
 // Options: [WithHTTPClient]
 func New(id ID, options ...ClientOption) (Client, error) {
 	opts := ClientOptions{
-		httpClient: shared.DefaultClient,
+		httpClient:         shared.DefaultClient,
+		retryPolicyDiabled: false,
 	}
 	for _, option := range options {
 		option(&opts)
@@ -136,8 +166,9 @@ func New(id ID, options ...ClientOption) (Client, error) {
 		return Client{}, fmt.Errorf("unsupported type %T", id)
 	}
 	client := Client{
-		miType:     id,
-		httpClient: opts.httpClient,
+		miType:              id,
+		httpClient:          opts.httpClient,
+		retryPolicyDisabled: opts.retryPolicyDiabled,
 	}
 	return client, nil
 }
@@ -178,8 +209,24 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
 	return req, nil
 }
 
+func retry(attempts int, delay time.Duration, c ops.HTTPClient, req *http.Request) (*http.Response, error) {
+	resp, err := c.Do(req)
+	if err == nil && !slices.Contains(retryStatusCodes, resp.StatusCode) {
+		return resp, nil // Success
+	}
+	if attempts-1 < 1 {
+		return resp, nil
+	}
+	time.Sleep(delay)
+	return retry(attempts-1, delay, c, req)
+}
+
 func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenResponse, error) {
-	resp, err := client.httpClient.Do(req)
+	retryCount := 1 // defaul Count
+	if !client.retryPolicyDisabled {
+		retryCount = defaultRetryCount
+	}
+	resp, err := retry(retryCount, time.Second, client.httpClient, req) //client.httpClient.Do(req)
 	if err != nil {
 		return accesstokens.TokenResponse{}, err
 	}
@@ -254,10 +301,10 @@ func (client Client) AcquireToken(ctx context.Context, resource string, options 
 	if err != nil {
 		return base.AuthResult{}, err
 	}
-	return authResultFromToken(fakeAuthParams, tokenResponse)
+	return client.authResultFromToken(fakeAuthParams, tokenResponse)
 }
 
-func authResultFromToken(authParams authority.AuthParams, token accesstokens.TokenResponse) (base.AuthResult, error) {
+func (c Client) authResultFromToken(authParams authority.AuthParams, token accesstokens.TokenResponse) (base.AuthResult, error) {
 	if cacheManager == nil {
 		return base.AuthResult{}, fmt.Errorf("cache instance is nil")
 	}
--- a/apps/managedidentity/managedidentity_test.go
+++ b/apps/managedidentity/managedidentity_test.go
@@ -102,7 +102,7 @@ func Test_SystemAssigned_Returns_AcquireToken_Failure(t *testing.T) {
 			}
 			fakeErrorClient.AppendResponse(mock.WithHTTPStatusCode(testCase.code),
 				mock.WithBody(responseBody))
-			client, err := New(SystemAssigned(), WithHTTPClient(&fakeErrorClient))
+			client, err := New(SystemAssigned(), WithHTTPClient(&fakeErrorClient), WithRetryPolicyDisabled())
 			if err != nil {
 				t.Fatal(err)
 			}
@@ -126,6 +126,74 @@ func Test_SystemAssigned_Returns_AcquireToken_Failure(t *testing.T) {
 			}
 		})
 	}
+}
+
+func Test_RetryPolicy_For_AcquireToken_Request(t *testing.T) {
+	t.Run("Testing retry policy with With One failure", func(t *testing.T) {
+		fakeClient := mock.Client{}
+		responseBody, err := makeResponseWithErrorData("sample error", "sample error desc")
+		if err != nil {
+			t.Fatalf("error while forming json response : %s", err.Error())
+		}
+		errorRetry := 2
+		errorRetryCounter := 0
+		fakeClient.AppendResponse(mock.WithHTTPStatusCode(http.StatusInternalServerError),
+			mock.WithBody(responseBody), mock.WithCallback(func(r *http.Request) {
+				errorRetryCounter += 1
+			}))
+		fakeClient.AppendResponse(mock.WithHTTPStatusCode(http.StatusInternalServerError),
+			mock.WithBody(responseBody), mock.WithCallback(func(r *http.Request) {
+				errorRetryCounter += 1
+			}))
+		fakeClient.AppendResponse(mock.WithHTTPStatusCode(http.StatusUnauthorized),
+			mock.WithBody(responseBody))
+		client, err := New(SystemAssigned(), WithHTTPClient(&fakeClient))
+		if err != nil {
+			t.Fatal(err)
+		}
+		resp, err := client.AcquireToken(context.Background(), resource)
+		if err == nil {
+			t.Fatalf("should have encountered the error")
+		}
+		if resp.AccessToken != "" {
+			t.Fatalf("wanted %q, got %q", "", resp.AccessToken)
+		}
+		if errorRetryCounter != errorRetry {
+			t.Fatalf("expected Number of retry of %d, got %d", errorRetry, errorRetryCounter)
+		}
+	})
+}
+
+func Test_RetryPolicy_For_AcquireToken_Request_MaxTries(t *testing.T) {
+	t.Run("Testing retry policy with With One failure", func(t *testing.T) {
+		fakeErrorClient := mock.Client{}
+		responseBody, err := makeResponseWithErrorData("sample error", "sample error desc")
+		if err != nil {
+			t.Fatalf("error while forming json response : %s", err.Error())
+		}
+		errorRetry := 4
+		errorRetryCounter := 0
+		for i := 0; i < errorRetry; i++ {
+			fakeErrorClient.AppendResponse(mock.WithHTTPStatusCode(http.StatusInternalServerError),
+				mock.WithBody(responseBody), mock.WithCallback(func(r *http.Request) {
+					errorRetryCounter += 1
+				}))
+		}
+		client, err := New(SystemAssigned(), WithHTTPClient(&fakeErrorClient))
+		if err != nil {
+			t.Fatal(err)
+		}
+		resp, err := client.AcquireToken(context.Background(), resource)
+		if err == nil {
+			t.Fatalf("should have encountered the error")
+		}
+		if resp.AccessToken != "" {
+			t.Fatalf("wanted %q, got %q", "", resp.AccessToken)
+		}
+		if errorRetryCounter != defaultRetryCount {
+			t.Fatalf("expected Number of retry of %d, got %d", errorRetry, errorRetryCounter)
+		}
+	})
 }
 
 func Test_SystemAssigned_Returns_Token_Success(t *testing.T) {


apps/managedidentity/managedidentity.go
```diff
@@ -178,8 +209,24 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
 	return req, nil
 }
 
+func retry(attempts int, delay time.Duration, c ops.HTTPClient, req *http.Request) (*http.Response, error) {
+	resp, err := c.Do(req)
+	if err == nil && !slices.Contains(retryStatusCodes, resp.StatusCode) {
+		return resp, nil // Success
+	}
+	if attempts-1 < 1 {
```
bgavrilMS, line 217: Seems like a strange condition

bgavrilMS (approved):

apps/managedidentity/managedidentity.go
```diff
@@ -178,8 +209,24 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
 	return req, nil
 }
 
+func retry(attempts int, delay time.Duration, c ops.HTTPClient, req *http.Request) (*http.Response, error) {
+	resp, err := c.Do(req)
+	if err == nil && !slices.Contains(retryStatusCodes, resp.StatusCode) {
```
chlowell, line 214: The `slices` package requires Go 1.21 but Azure SDK and MSAL still support 1.18. I don't think it's worth upgrading our min version for this because the alternative to `Contains` is a basic `for` loop

apps/managedidentity/managedidentity_test.go
```diff
@@ -128,6 +128,74 @@ func Test_SystemAssigned_Returns_AcquireToken_Failure(t *testing.T) {
 	}
 }
 
+func Test_RetryPolicy_For_AcquireToken_Request(t *testing.T) {
+	t.Run("Testing retry policy with With One failure", func(t *testing.T) {
```
chlowell, line 132: Why run this as a subtest? Isn't it simply the main test? There are no other subtests

apps/managedidentity/managedidentity.go
```diff
@@ -254,10 +301,10 @@ func (client Client) AcquireToken(ctx context.Context, resource string, options
 	if err != nil {
 		return base.AuthResult{}, err
 	}
-	return authResultFromToken(fakeAuthParams, tokenResponse)
+	return client.authResultFromToken(fakeAuthParams, tokenResponse)
 }
 
-func authResultFromToken(authParams authority.AuthParams, token accesstokens.TokenResponse) (base.AuthResult, error) {
+func (c Client) authResultFromToken(authParams authority.AuthParams, token accesstokens.TokenResponse) (base.AuthResult, error) {
```
chlowell, line 307: Unused receiver

apps/managedidentity/managedidentity.go
```diff
@@ -107,13 +130,20 @@ func WithHTTPClient(httpClient ops.HTTPClient) ClientOption {
...
+
 // Client to be used to acquire tokens for managed identity.
 // ID: [SystemAssigned], [UserAssignedClientID], [UserAssignedResourceID], [UserAssignedObjectID]
 //
 // Options: [WithHTTPClient]
 func New(id ID, options ...ClientOption) (Client, error) {
 	opts := ClientOptions{
-		httpClient: shared.DefaultClient,
+		httpClient:         shared.DefaultClient,
+		retryPolicyDiabled: false,
```
chlowell, line 146: `false` is the zero value

apps/managedidentity/managedidentity.go
```diff
@@ -75,13 +97,14 @@ func SystemAssigned() ID {
...
-	// source     Source reenable when required in future sources
+	httpClient          ops.HTTPClient
+	miType              ID
+	retryPolicyDisabled bool
 }
 
 type ClientOptions struct {
-	httpClient ops.HTTPClient
+	httpClient         ops.HTTPClient
+	retryPolicyDiabled bool
```
chlowell, line 107:
```suggestion
	retryPolicyDisabled bool
```

apps/managedidentity/managedidentity.go
```diff
@@ -178,8 +209,24 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
 
+func retry(attempts int, delay time.Duration, c ops.HTTPClient, req *http.Request) (*http.Response, error) {
+	resp, err := c.Do(req)
+	if err == nil && !slices.Contains(retryStatusCodes, resp.StatusCode) {
+		return resp, nil // Success
+	}
+	if attempts-1 < 1 {
+		return resp, nil
+	}
+	time.Sleep(delay)
```
chlowell, line 220: MSAL must observe Context cancellation while waiting to retry. See for example the [Azure SDK retry policy](https://github.com/Azure/azure-sdk-for-go/blob/5635b52606e8f156575c11317b87a2d51013b710/sdk/azcore/runtime/policy_retry.go#L212-L219)

apps/managedidentity/managedidentity.go
```diff
@@ -178,8 +209,24 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+	return retry(attempts-1, delay, c, req)
```
chlowell, line 221: You probably need to rewind the request body before retrying. Make sure you have a test verifying retries have the same content as the initial attempt

chlowell (changes requested):
Is there a spec for this feature? It seems very basic e.g. no backoff, hardcoded retry count and delay. Blockers are
- implicit upgrade to Go 1.21
- cancellation
- rewinding request bodies

4gust:
It is very basic retry, and has a fix 1 second between retry and max count of 3 retries 
This is what we set for Private preview.

apps/managedidentity/managedidentity.go
```diff
@@ -75,13 +96,14 @@ func SystemAssigned() ID {
 var cacheManager *storage.Manager = storage.New(nil)
 
 type Client struct {
-	httpClient ops.HTTPClient
-	miType     ID
-	// source     Source reenable when required in future sources
+	httpClient          ops.HTTPClient
+	miType              ID
+	retryPolicyDisabled bool
```
bgavrilMS, line 101: How about we revert the flag, smth like "internalRetryPolicyEnabled" which defaults to `true`

4gust: You mean that its disabled by-default and you enable like 
```go
miSystemAssigned, err := mi.New(mi.SystemAssigned(), WithEnabledInternalRetryPolicy())
```

bgavrilMS (approved):

AndyOHart (approved):

apps/managedidentity/managedidentity.go
```diff
@@ -178,8 +207,47 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+// retry performs an HTTP request with retries based on the provided options.
+func retry(maxRetries int, c ops.HTTPClient, req *http.Request) (*http.Response, error) {
+	var resp *http.Response
+	var err error
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		tryCtx, tryCancel := context.WithTimeout(req.Context(), time.Second*15)
+		defer tryCancel()
+		cloneReq := req.Clone(tryCtx)
+		resp, err = c.Do(cloneReq)
+		if err == nil && !contains(retryStatusCodes, resp.StatusCode) {
```
chlowell, line 229: This should consider the Source. `retryStatusCodes` follows IMDS docs, and that guidance doesn't make sense for other APIs. For example, in general you don't want to retry 404. Azure SDK retries [these codes](https://github.com/Azure/azure-sdk-for-go/blob/5635b52606e8f156575c11317b87a2d51013b710/sdk/azcore/runtime/policy_retry.go#L51-L56) by default and special-cases IMDS

bgavrilMS: CC @neha-bhargava as this is probably a small bug in all other MSALs (which should not affect Azure SDK)

4gust: After discussions with @bgavrilMS, we have decided to maintain consistency with other MSAL implementations. We will create a bug report to address this issue and ensure that the necessary updates are applied across all relevant areas.

apps/managedidentity/managedidentity.go
```diff
@@ -50,8 +51,20 @@ const (
 	imdsAPIVersion = "2018-02-01"
 
 	systemAssignedManagedIdentity = "system_assigned_managed_identity"
+	defaultRetryCount             = 3
 )
 
+// IMDS docs recommend retrying 408, 429, 500, 502, 503, 504
```
chlowell, line 57: now the comment is incorrect

bgavrilMS:
> Is there a spec for this feature? It seems very basic e.g. no backoff, hardcoded retry count and delay. Blockers are
> 
> * implicit upgrade to Go 1.21
> * cancellation
> * rewinding request bodies

This is the retry policy that was discussed with MSI team and implemented in all MSALs. Happy to redefine this, especially as we introduce MSIv2. But it is important for MSALs to be consistent here. Note that MSALs do not offer a `DefaultAzureCredential` chaining, so the usage scenario is different.

Azure SDK can and should disable the retry policy so as to introduce your own.

bgavrilMS (approved):

chlowell:
I'm sorry, I still don't know which status codes this policy is supposed to retry on. Right now, its retriable codes are reasonable in general cases but incorrect for IMDS. Do you intend it to apply IMDS retry guidance to all services?

apps/managedidentity/managedidentity.go
```diff
@@ -178,8 +199,47 @@ func createIMDSAuthRequest(ctx context.Context, id ID, resource string, claims s
...
+		cloneReq := req.Clone(tryCtx)
+		resp, err = c.Do(cloneReq)
+		if err == nil && !contains(retryStatusCodes, resp.StatusCode) {
+			return resp, nil
+		}
+		if attempt == maxRetries-1 {
+			return resp, err
+		}
+		if resp != nil && resp.Body != nil {
+			io.Copy(io.Discard, resp.Body)
```
chlowell, line 228: Why?

4gust: Discussed Offline about this memory leak

apps/managedidentity/managedidentity.go
```diff
@@ -50,8 +51,19 @@ const (
 	imdsAPIVersion = "2018-02-01"
 
 	systemAssignedManagedIdentity = "system_assigned_managed_identity"
+	defaultRetryCount             = 3
 )
 
+// retry on these codes
+var retryStatusCodes = []int{
```
bgavrilMS, line 58: Slightly mismatched with .NET implementation: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/785fc1843e6244470db00237f2d0e020c1180e95/src/client/Microsoft.Identity.Client/Http/HttpRetryCondition.cs#L25

4gust: Updated the codes to match other MSAL's 

Azure sdk for go are from where the rest of the codes are 
https://github.com/Azure/azure-sdk-for-go/blob/3ebd0d439f9d8aa06a0764a60892a8001b997a34/sdk/azidentity/managed_identity_client.go#L114

apps/managedidentity/managedidentity.go
```diff
@@ -323,9 +360,56 @@ func authResultFromToken(authParams authority.AuthParams, token accesstokens.Tok
...
+	var err error
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		tryCtx, tryCancel := context.WithTimeout(req.Context(), time.Second*15)
+		defer tryCancel()
+		cloneReq := req.Clone(tryCtx)
+		resp, err = c.Do(cloneReq)
+		retrylist := retryStatusCodes
+		if s == DefaultToIMDS {
+			retrylist = append(retrylist, retryCodesForIMDS...)
+		}
```
chlowell, line 385: Note that on an IMDS platform, every retry allocates and discards enough memory to hold `retryStatusCodes` and `retryCodesForIMDS`. It isn't much memory but this makes unnecessary work for the GC. There's no need to recompute this slice because the codes don't change between iterations of this loop or even between calls to this function. Why not simply have `retryCodesForIMDS` hold all the codes for IMDS and `retryStatusCodes` hold all the codes for other sources?

apps/managedidentity/managedidentity.go
```diff
@@ -72,8 +73,34 @@ const (
 	imdsEndVar                          = "IMDS_ENDPOINT"
 	msiEndpointEnvVar                   = "MSI_ENDPOINT"
 	identityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
+
+	defaultRetryCount = 3
 )
 
+// retry codes for IMDS
+var retryCodesForIMDS = []int{
```
chlowell, line 81: This isn't correct. For example, IMDS clients should retry 404. See [IMDS docs](https://learn.microsoft.com/entra/identity/managed-identities-azure-resources/how-to-use-vm-token#error-handling) and [Azure SDK](https://github.com/Azure/azure-sdk-for-go/blob/6bb3a6513654fec9809c4fba235cf9a2c1b7d935/sdk/azidentity/managed_identity_client.go#L108)

apps/managedidentity/managedidentity_test.go
```diff
@@ -400,6 +593,7 @@ func TestAzureArcPlatformSupported(t *testing.T) {
 	}
 	result, err := client.AcquireToken(context.Background(), resource)
 	if err == nil || !strings.Contains(err.Error(), "platform not supported") {
+		println(result.AccessToken)
```
chlowell, line 596: 👀

apps/managedidentity/managedidentity.go
4gust: All the msal's have this `404 or 408 or 429 or 500 or 503 or 504`  
I want to get it right, but the IMDS docs don't have concrete codes only says that `404, 410, 429 and 4xx ` 
If i can get that i will be able to create 2 lists here

bgavrilMS: Let's go with Azure SDK's list @4gust , as that has been tried in production for some time.

bgavrilMS: Azure SDK has a default retry policy for all its services here: https://github.com/Azure/azure-sdk-for-go/blob/6bb3a6513654fec9809c4fba235cf9a2c1b7d935/sdk/azcore/runtime/policy_retry.go#L50 (so retry on 408, 429, 500, 502, 503, 504). 

And for IMDS @chlowell pointed us to it. 

Let's use the same approach - linear retry. I know the IMDS docs suggest exponential backoff, but this is what we discussed in other MSALs and I would like not to diverge here. Let's track any divergence from the other MSALs, otherwise it becomes unscalable. 


CC @neha-bhargava for a perspective on why we chose linear retry in MSALs

neha-bhargava: Linear retry was chosen in MSAL since we do not give flexibility to the user to choose how many times to retry but retry for 3 times (fixed). Azure SDK gives this capability to the user to configure the number of retries. For a small number of retries in MSAL such as 3, it did not make sense to implement exponential backoff.

Azure SDK use their own http client and the retry policy followed in that case is the one that they have configured in their http client so the user experience for them did not change.

neha-bhargava: Also when I checked the Azure SDK .Net code at that time, the exponential backoff computation started after 3 retries with fixed delay. So we decided to keep it linear in MSAL

apps/managedidentity/managedidentity.go
```diff
@@ -323,9 +365,56 @@ func authResultFromToken(authParams authority.AuthParams, token accesstokens.Tok
...
+		cloneReq := req.Clone(tryCtx)
+		resp, err = c.Do(cloneReq)
+		retrylist := retryStatusCodes
+		if s == DefaultToIMDS {
+			retrylist = retryCodesForIMDS
+		}
+		if err == nil && !contains(retrylist, resp.StatusCode) {
+			return resp, nil
+		}
+		if attempt == maxRetries-1 {
```
chlowell, line 394: you can delete this block if you drain the response body at the start of the loop

apps/managedidentity/managedidentity.go
```diff
@@ -323,9 +365,56 @@ func authResultFromToken(authParams authority.AuthParams, token accesstokens.Tok
...
+	}
+	return false
+}
+
+// retry performs an HTTP request with retries based on the provided options.
+func retry(maxRetries int, c ops.HTTPClient, req *http.Request, s Source) (*http.Response, error) {
+	var resp *http.Response
+	var err error
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		tryCtx, tryCancel := context.WithTimeout(req.Context(), time.Second*15)
```
chlowell, line 383: We mustn't do this for applications that disable the retry policy. Those applications must have total control over timeouts. Really `Client` should call `Do` directly as before when the retry policy is disabled

apps/managedidentity/managedidentity.go
```diff
@@ -72,8 +73,39 @@ const (
...
 	msiEndpointEnvVar                   = "MSI_ENDPOINT"
 	identityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
+
+	defaultRetryCount = 3
 )
 
+// retry codes for IMDS
+var retryCodesForIMDS = []int{
+	http.StatusNotFound,                      // 404
+	http.StatusRequestTimeout,                // 408
```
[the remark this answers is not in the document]
chlowell: Docs say not to retry on 408

apps/managedidentity/managedidentity_test.go
```diff
@@ -143,6 +146,196 @@ func TestSource(t *testing.T) {
...
+			if testCase.expectedFail {
+				if err == nil {
+					t.Fatalf("should have encountered the error")
+				}
+				if resp.AccessToken != "" {
+					t.Fatalf("accesstoken should be empty")
+				}
+			} else {
+				if err != nil {
+					t.Fatalf("should have encountered the error")
```
chlowell, line 320: 👀

apps/managedidentity/managedidentity_test.go
```diff
@@ -143,6 +146,196 @@ func TestSource(t *testing.T) {
...
+				t.Fatalf("Expected body %q, got %q", tt.expectedBody, bodyBytes)
+			}
+			if req.Body != nil {
+				reqBodyBytes, err := io.ReadAll(req.Body)
+				if err != nil {
+					t.Fatalf("Failed to read request body: %v", err)
+				}
+				req.Body.Close()
+
+				if string(reqBodyBytes) != tt.requestBody {
```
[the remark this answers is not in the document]
4gust: Noted!

apps/managedidentity/managedidentity_test.go
```diff
@@ -143,6 +146,188 @@ func TestSource(t *testing.T) {
...
+	tests := []struct {
+		name          string
+		mockResponses []struct {
+			body       string
+			statusCode int
+		}
+		expectedStatus int
+		expectedBody   string
+		maxRetries     int
+		requestBody    string
```
chlowell, line 159: This has the same value in all cases and now you don't need it anyway

apps/managedidentity/managedidentity_test.go
```diff
@@ -143,6 +146,188 @@ func TestSource(t *testing.T) {
...
+				if resp.AccessToken != token {
+					t.Fatalf("wanted %q, got %q", token, resp.AccessToken)
+				}
+			}
+			if testCase.disableRetry {
+				if errorRetryCounter != 1 {
+					t.Fatalf("expected Number of retry of 1, got %d", errorRetryCounter)
+				}
+			} else {
+				if errorRetryCounter != testCase.numberOfFails && testCase.numberOfFails < 3 {
```
chlowell, line 323:
```suggestion
			} else if errorRetryCounter != testCase.numberOfFails && testCase.numberOfFails < 3 {
```

apps/managedidentity/managedidentity.go
```diff
@@ -72,8 +73,38 @@ const (
 	imdsEndVar                          = "IMDS_ENDPOINT"
 	msiEndpointEnvVar                   = "MSI_ENDPOINT"
 	identityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
+
+	defaultRetryCount = 3
 )
 
+// retry codes for IMDS
```
chlowell, line 80:
```suggestion
```

apps/managedidentity/managedidentity.go
```diff
@@ -72,8 +73,38 @@ const (
...
+	http.StatusGatewayTimeout,                // 504
+	http.StatusHTTPVersionNotSupported,       // 505
+	http.StatusVariantAlsoNegotiates,         // 506
+	http.StatusInsufficientStorage,           // 507
+	http.StatusLoopDetected,                  // 508
+	http.StatusNotExtended,                   // 510
+	http.StatusNetworkAuthenticationRequired, // 511
+}
+
+// retry on these codes
```
chlowell, line 98:
```suggestion
```

apps/managedidentity/managedidentity.go
```diff
@@ -323,9 +364,55 @@ func authResultFromToken(authParams authority.AuthParams, token accesstokens.Tok
 	return ar, err
 }
 
-func (client Client) getTokenForRequest(req *http.Request) (accesstokens.TokenResponse, error) {
-	var r accesstokens.TokenResponse
-	resp, err := client.httpClient.Do(req)
+// Contains checks if the element is present in the list.
```
chlowell, line 367:
```suggestion
// contains checks if the element is present in the list.
```

apps/managedidentity/managedidentity.go
```diff
@@ -323,9 +364,55 @@ func authResultFromToken(authParams authority.AuthParams, token accesstokens.Tok
...
+
+// retry performs an HTTP request with retries based on the provided options.
+func (c Client) retry(maxRetries int, req *http.Request) (*http.Response, error) {
+	var resp *http.Response
+	var err error
+	for attempt := 0; attempt < maxRetries; attempt++ {
+		tryCtx, tryCancel := context.WithTimeout(req.Context(), time.Second*15)
+		defer tryCancel()
+		if resp != nil && resp.Body != nil {
+			io.Copy(io.Discard, resp.Body)
```
chlowell, line 385: linters warn about implicitly ignoring errors
```suggestion
			_, _ = io.Copy(io.Discard, resp.Body)
```

chlowell (approved):

PR #526, opened 2024-11-06
Fix invalid authority uri

bgavrilMS:
the dSTS commits introduced a regression which was caught by the integration tests (yay!). Sadly the integration tests do not run as part of the PR :(.


4gust (approved):
Looks good to me!

PR #535, opened 2024-11-19
Praposal for logging support for msal-go

4gust:
Praposal for logging support for msal-go


bgavrilMS (approved):
We discussed that we don't need the callback option for 1.18 for now.

apps/design/logPraposal.md
```diff
@@ -0,0 +1,333 @@
...
+
+import "fmt"
+
+// CallbackFunc defines the signature for callback functions
+// we can only have one string to support azure sdk
+type CallbackFunc func(level, message string)
+
+// Logger struct for Go versions <= 1.20.
+type Logger struct {
+	LogCallback CallbackFunc
```
bgavrilMS, line 92: We discussed that for golang 1.18 it's ok to not have any logging.

apps/logger/logger.go
```diff
@@ -8,17 +6,32 @@ import (
...
-	logging     *slog.Logger
-	logCallback CallbackFunc
+	logging *slog.Logger
 }
 
-func New121(input interface{}) (*Logger, error) {
-	if logger, ok := input.(*slog.Logger); ok {
-		return &Logger{logging: logger}, nil
+// New creates a new logger instance
+func New(slogLogger *slog.Logger) (*Logger, error) {
```
bgavrilMS, line 28: I am not sure about this. The moment we put slog in the public API surface, Azure SDK will not be able to use MSAL SDK.

bgavrilMS: See with https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/main/apps/design/release.md for testing with azure sdk

apps/confidential/confidential_121_plus.go
```diff
@@ -0,0 +1,21 @@
...
+package confidential
+
+import (
+	"fmt"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/logger"
+)
+
+//  WithLogger allows for a custom logger to be set.
+func WithLogger(l interface{}) Option {
```
chlowell, line 12: `interface{}` effectively disables compiler type checking here, making this function difficult to use and turning a class of build errors into runtime errors--not great. This option only works if `l` is a `*slog.Logger`; should that be the parameter's type?

apps/internal/logger/logger_121_plus.go
```diff
@@ -0,0 +1,60 @@
...
+		slogLevel = slog.LevelWarn
+	case Debug:
+		slogLevel = slog.LevelDebug
+	default:
+		slogLevel = slog.LevelInfo
+	}
+
+	// Log the entry with the message and fields
+	a.logging.Log(
+		context.Background(),
```
chlowell, line 50: this must be the Context of the encompassing operation i.e., the caller must provide it

apps/internal/logger/logger_121_plus.go
```diff
@@ -0,0 +1,60 @@
...
+func NewLogger(loggerInterface interface{}) (LoggerInterface, error) {
+	if loggerInterface == nil {
+		return &logger{logging: nil}, nil
+	}
+
+	if loggerInterface, ok := loggerInterface.(*slog.Logger); ok {
+		return &logger{logging: loggerInterface}, nil
+	}
+
+	return nil, fmt.Errorf("invalid input for Go 1.21+; expected *slog.Logger")
```
chlowell, line 26: then shouldn't the parameter be of type `*slog.Logger`?

apps/design/logging.md
```diff
@@ -0,0 +1,36 @@
...
+time=2024-12-19T01:25:58.730Z level=INFO msg="This is an info message via slog." username=john_doe age=30
+time=2024-12-19T01:25:58.730Z level=ERROR msg="This is an error message via slog." module=user-service retry=3
+time=2024-12-19T01:25:58.730Z level=WARN msg="Disk space is low." free_space_mb=100
+time=2024-12-19T01:25:58.730Z level=INFO msg="Default log message." module=main
+```
+
+## Key Pointers
+
+1. **For Go <= 1.20**:
+   - User should pass *nil* to `logger.New(nil)`
```
chlowell, line 19: Users can't call `logger.New` because it's in an `internal` package. Is this doc or the implementation incorrect?

apps/confidential/confidential_121_plus.go
```diff
@@ -0,0 +1,21 @@
...
+	return func(o *clientOptions) {
+		logInstance, err := logger.New(l)
+		if err != nil {
+			fmt.Println("Error creating logger with slog:", err)
+			return
```
chlowell, line 17: This won't do. The caller has every reason to believe the call succeeded and won't find out something went wrong until they notice the lack of logging and/or this line in stdout. Returning the error is the right thing to do but impossible here. So, this needs a redesign. Off the top of my head, I suppose the alternatives are:
1. make it impossible for this function to receive an error i.e., make it impossible for `logger.New` to return an error
2. separate logger construction from this function i.e., have users call `logger.New` themselves, check its error, pass the logger to `WithLogger`

chlowell (changes requested):

apps/tests/devapps/logger/logger_sample.go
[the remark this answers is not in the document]
AndyOHart: Was made for testing purposes but can remove

apps/design/logging.md
AndyOHart: This doc is the old proposal that was updated for the newer implementation, but the new implementation has since changed so will have to update this document once we are happy with the implementation

apps/confidential/confidential_121_plus.go
bgavrilMS: Option 2 is not great for developer experience though. The app developer would be exposed to an unnecessary abstraction.

Option 1 looks great. If not possible, would Option 3 be feasible - return the error in the ConfidentialClientApplication / ManagedIdentityApplication builder ?

bgavrilMS: I don't see any updates in the PR

chlowell: The question I'm really circling around here is, do you imagine ever supporting logging via something other than slog? Answering this points the way forward. If you want to support only slog, you can simply have `WithLogger(logger slog.Logger)`, you don't need an abstraction layer like `LoggerInterface`, and you get my option 1 for free because this `WithLogger` can't produce an error. If you want to maintain the possibility of easy non-slog logging, then you need to expose `LoggerInterface`, have `WithLogger` take one of those, and decide what to do when someone calls `WithLogger(nil)`

AndyOHart: I think the main issue or maybe just a misunderstanding on how best to handle it, is for go versions < 1.21
The LoggerInterface i think makes it simple to handle having something that supports 1.21 and also not causing an issue with < 1.21. 
If we remove the loggerInterface would it not cause the code to become a bit more messy and having to use build tags directly inside the confidential.go class, and having clientOptions defined in both places?

apps/confidential/confidential_121_plus.go
```diff
@@ -0,0 +1,16 @@
...
+package confidential
+
+import (
+	"log/slog"
+
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/logger"
+)
+
+// WithLogger allows for a custom logger to be set.
+func WithLogger(l *slog.Logger) Option {
```
bgavrilMS, line 12: Yes, this is a good interface. Not sure about the comment though ... Pls add a link to slog and some details about how a log consumer would set this up.

apps/design/logging.md
```diff
@@ -0,0 +1,36 @@
...
+
+2. **Full `slog` Support for Go 1.21+**:
+   - The `logger.go` file leverages the `slog` package, supporting structured logs, multiple log levels (`info`, `error`, `warn`), and fields.
+
+3. **Structured Logging**:
+   - You can pass key-value pairs using `slog.String`, `slog.Int`, etc., for structured logging, which is handled by `slog` in Go 1.21 and later.
+
+### **Sample**
+
+A sample of the logger can be found in the following location:
```
bgavrilMS, line 30: I would prefer a wiki page or just put the code here. Dev apps are not samples. Samples have their own repos etc.

bgavrilMS (commented):
LGTM if this works. Not signing off as there is some work to be done.

apps/internal/logger/logger.go
```diff
@@ -0,0 +1,44 @@
+//go:build go1.21
```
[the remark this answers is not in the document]
AndyOHart: No this should just be included in the build if Go version is 1.21 or higher

apps/design/logging.md
AndyOHart: I will remove this once PR is approved, as I am using it for some testing currently, but will remove before merge

apps/confidential/confidential.go
```diff
@@ -345,6 +349,7 @@ func New(authority, clientID string, cred Credential, options ...Option) (Client
 	}
 	base.AuthParams.IsConfidentialClient = true
 
+	opts.logger.Log(context.Background(), logger.Info, "Created confidential client", logger.Field("clientID", clientID))
```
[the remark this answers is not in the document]
AndyOHart: Will remove this once the PR is approved as it is here now for testing

apps/confidential/confidential.go
```diff
@@ -318,12 +320,14 @@ func New(authority, clientID string, cred Credential, options ...Option) (Client
...
 	}
 	autoEnabledRegion := os.Getenv("MSAL_FORCE_REGION")
+
 	opts := clientOptions{
 		authority: authority,
 		// if the caller specified a token provider, it will handle all details of authentication, using Client only as a token cache
 		disableInstanceDiscovery: cred.tokenProvider != nil,
 		httpClient:               shared.DefaultClient,
 		azureRegion:              autoEnabledRegion,
+		logger:                   nil,
```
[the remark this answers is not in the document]
AndyOHart: Updated

apps/internal/logger/logger.go
```diff
@@ -0,0 +1,44 @@
...
+
+type Logger struct {
+	*slog.Logger
+}
+
+// New creates a new logger instance for Go 1.21+ with full `slog` logging support.
+// If nil is provided a default logger instance is created.
+func New(slogLogger *slog.Logger) *Logger {
+	if slogLogger == nil {
+		defaultLogger := slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{}))
```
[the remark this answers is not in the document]
AndyOHart: Updated

apps/internal/logger/logger.go
```diff
@@ -0,0 +1,44 @@
...
+type Level = slog.Level
+
+const (
+	Debug = slog.LevelDebug
+	Info  = slog.LevelInfo
+	Warn  = slog.LevelWarn
+	Error = slog.LevelError
+)
+
+type Logger struct {
```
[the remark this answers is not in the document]
AndyOHart: Updated

apps/internal/logger/logger_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// Log messages
+	loggerInstance.Log(context.Background(), slog.LevelInfo, "This is an info message via slog.", slog.Any("username", "john_doe"), slog.Int("age", 30))
+	loggerInstance.Log(context.Background(), slog.LevelError, "This is an error message via slog.", slog.String("module", "user-service"), slog.Int("retry", 3))
+	loggerInstance.Log(context.Background(), slog.LevelWarn, "This is a warn message via slog.", slog.Int("free_space_mb", 100))
+	loggerInstance.Log(context.Background(), slog.LevelDebug, "This is a debug message via slog.", slog.String("module", "main"))
+
+	// Check the output
+	output := buf.String()
+	expectedMessages := []string{
```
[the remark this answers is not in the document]
AndyOHart: Updated

apps/confidential/confidential_121_plus.go
AndyOHart: Updated

apps/confidential/confidential_121_plus.go
```diff
@@ -0,0 +1,19 @@
+//go:build go1.21
+
+package confidential
+
+import (
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/slog"
```
chlowell, line 6: When pkgsite renders docs for `WithLogger`, it creates a link to the definition of the parameter's type i.e., `slog.Logger`. If that type is defined in your `internal/slog` package, the doc link will point there, meaning anyone who clicks that link ends up at docs for an internal package they can't reference. Of course that package references types from `logs/slog` and therefore its docs link in turn to that package, but you can avoid linking to your internal package by importing `logs/slog` here:
```suggestion
	"log/slog"
```
(this doesn't break the build because the compiler is smart enough to realize that your `internal/slog.Logger` is the same type as `logs/slog.Logger`)

apps/confidential/confidential_121_plus.go
chlowell, line 1: How about `with_logger.go`?

apps/internal/slog/logger.go
```diff
@@ -0,0 +1,33 @@
...
+func New(slogLogger *slog.Logger) *Logger {
+	if slogLogger == nil {
+		defaultLogger := slog.Default()
+		return defaultLogger
+	}
+	return slogLogger
+}
+
+// Field creates a slog field for any value
+func Field(key string, value any) any {
```
chlowell, line 31: Do you need this?

apps/internal/slog/logger_121_below.go
```diff
@@ -0,0 +1,29 @@
...
+package slog
+
+import "context"
+
+type Level int
+
+type Logger struct{}
+
+const (
+	Debug Level = iota
```
chlowell, line 12:
```suggestion
	LevelDebug Level = iota
```
etc. These symbols must be the same for all supported versions of Go

apps/internal/slog/logger_test.go
```diff
@@ -0,0 +1,72 @@
...
+				t.Errorf("expected attribute %q not found in output for message %q", attr, expected.msg)
+			}
+		}
+	}
+}
+
+func TestNewLogger_NilLoggerInterface(t *testing.T) {
+	// Test case where slogLogger is nil
+	logInstance := New(nil)
+	if logInstance == nil {
```
chlowell, line 56: But is `logInstance` the default logger?

chlowell (commented):
Are you adding tests for `confidential.Client` as well i.e., testing that `confidential.New(WithLogger(...))` constructs a Client that writes logs?

Is the same API coming for `public.Client` in another PR?

apps/confidential/confidential_121_plus.go
chlowell: 🤔 the important thing about this option is that it enables logging, not that it enables specifying a custom Logger. The parameter type itself expresses that. So I think this comment only needs to cover that this option enables logging and when `l` is nil, Client will use `slog.Default()`, which will lead to a panic when `slog.Default()` returns nil. You don't need to show how to call this function or explicitly link to the slog docs. pkgsite will create a link to the slog docs on the parameter type

AndyOHart: Updated

AndyOHart (approved):
PR was created by Nilesh but it also seems good to me from working on it

apps/confidential/confidential_121_plus.go
AndyOHart: Interesting to know thanks!

chlowell:
How will we handle logging PII? Another argument to `WithLogger`, or could we use slog to mark a message or part of a message as PII and leave handling that up to the application?

apps/internal/slog/logger_121_below.go
```diff
@@ -0,0 +1,35 @@
...
+type Logger struct{}
+
+const (
+	LevelDebug Level = iota
+	LevelInfo
+	LevelWarn
+	LevelError
+)
+
+type NopHandler struct{}
```
chlowell, line 20: not a blocker but consider defining this once, in a separate file e.g. `nop_handler.go` with no build constraint

apps/managedidentity/managedidentity_with_logger_test.go
```diff
@@ -0,0 +1,128 @@
...
+		return true
+	})
+	return nil
+}
+func (h *BufferHandler) WithAttrs(attrs []slog.Attr) slog.Handler { return h }
+func (h *BufferHandler) WithGroup(name string) slog.Handler       { return h }
+func (h *FilteredBufferHandler) Enabled(ctx context.Context, level slog.Level) bool {
+	return level == slog.LevelDebug
+}
+func (h *FilteredBufferHandler) Handle(ctx context.Context, record slog.Record) error {
```
chlowell, line 40: you could add a `level` field to BufferHandler instead of defining a second type

apps/managedidentity/managedidentity_with_logger_test.go
```diff
@@ -0,0 +1,128 @@
...
+		h.buf.WriteString(attr.Key + "=" + attr.Value.String() + " ")
+		return true
+	})
+	return nil
+}
+func (h *FilteredBufferHandler) WithAttrs(attrs []slog.Attr) slog.Handler { return h }
+func (h *FilteredBufferHandler) WithGroup(name string) slog.Handler       { return h }
+
+func TestClientLogging(t *testing.T) {
+	// Set up mock client
```
chlowell, line 52: please forgive nitpicking: I'd remove these comments because the code clearly says what it does

apps/managedidentity/managedidentity_with_logger_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+func TestClientLogging_NoLoggerProvided(t *testing.T) {
+	// Set up mock client
+	mockClient := mock.Client{}
+	headers := http.Header{}
+	headers.Set("www-authenticate", "Basic realm=/path/to/secret.key")
+	mockClient.AppendResponse(mock.WithHTTPStatusCode(http.StatusUnauthorized), mock.WithHTTPHeader(headers))
+
+	// Create a custom logger with BufferHandler to capture logs
+	bufferHandler := &BufferHandler{}
```
chlowell, line 86: I'd delete this test because it doesn't add any coverage and can't catch any logging bug. It doesn't mutate the handler or pass it to another function, so of course the handler has the same state at the end of the test

apps/managedidentity/managedidentity_with_funcs.go
```diff
@@ -0,0 +1,21 @@
...
+
+// WithLogger enables logging within the SDK
+func WithLogger(l *slog.Logger) ClientOption {
+	return func(o *ClientOptions) {
+		o.logger = l
+	}
+}
+
+// WithPiiLogging enables logging of Personally Identifiable Information (PII) within the SDK
+func WithPiiLogging() ClientOption {
```
bgavrilMS, line 17: I would prefer to have a `WithLogger(l *slog.Logger, piiEnabled bool)` single API because: 

1. It reduces the number of methods that IDEs "auto-complete" display. We had some developer feedback on other MSALs that we expose too many options already...
2. Consistency with other MSALs

AndyOHart: Updated

apps/internal/slog/nop_handler.go
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
+package slog
+
+type NopHandler struct{}
```
chlowell, line 6: I meant you can put this type and all its methods in the same file, built unconditionally, so you don't have to duplicate them for both version ranges. You would have to define some additional stub types for versions < 1.21. One uninteresting line doesn't deserve its own file, so if you don't want to define the stub types, I suggest simply duplicating this definition as you did previously

apps/internal/slog/logger_test.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func TestNewLogger_ValidSlogHandler(t *testing.T) {
+	// Test case where handler is a valid slog.Handler
+	var buf bytes.Buffer
+	handler := slog.NewJSONHandler(&buf, &slog.HandlerOptions{
+		Level: slog.LevelDebug,
+	})
+	logInstance := New(handler)
+	if logInstance == nil {
```
chlowell, line 64: The above test calls `New` with the same arguments and will panic when the return is nil. So, you may as well delete this test

apps/managedidentity/managedidentity_with_funcs.go
chlowell, line 1: options.go?

apps/managedidentity/managedidentity.go
```diff
@@ -307,8 +317,9 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
 		option(&o)
 	}
 	c.authParams.Scopes = []string{resource}
+	c.logger.Log(ctx, slog.LevelInfo, "Managed Identity", slog.String("source", string(c.source)))
```
chlowell, line 320: This will log the source upon every token request. While it's definitely helpful to log the source, you really only want to do so once because the value is immutable. I guess you placed this here instead of in `New` because `New` doesn't get context? This is pretty awkward 😆. I suppose the possible solutions are
- create a context (with timeout) in `New` for logging the source
- log this here but only once per process (see [sync.Once](https://pkg.go.dev/sync#Once))

Option 2 looks better to me

apps/managedidentity/managedidentity_with_logger_test.go
```diff
@@ -0,0 +1,80 @@
...
+	return nil
+}
+
+func (h *BufferHandler) WithAttrs(attrs []slog.Attr) slog.Handler { return h }
+func (h *BufferHandler) WithGroup(name string) slog.Handler       { return h }
+
+func TestClientLogging(t *testing.T) {
+	mockClient := mock.Client{}
+	headers := http.Header{}
+	headers.Set("www-authenticate", "Basic realm=/path/to/secret.key")
```
chlowell, line 40: Is this necessary?

apps/managedidentity/managedidentity_with_logger_test.go
```diff
@@ -0,0 +1,80 @@
...
+	_, _ = client.AcquireToken(context.Background(), "https://resource")
+
+	logOutput := bufferHandler.buf.String()
+	expectedLogMessage := "Managed Identity"
+	if !strings.Contains(logOutput, expectedLogMessage) {
+		t.Errorf("expected log message %q not found in output", expectedLogMessage)
+	}
+}
+
+func TestClientLogging_CustomHandler(t *testing.T) {
```
chlowell, line 59: This is nearly identical to the above test. Really it just verifies that `AcquireToken` logs its message at a level higher than Debug. Couldn't you do that more simply by having the above test also verify the level directly?

apps/managedidentity/managedidentity_with_logger_test.go
```diff
@@ -0,0 +1,71 @@
+//go:build go1.21
```
chlowell, line 1: license

apps/managedidentity/managedidentity.go
```diff
@@ -260,7 +269,14 @@ func New(id ID, options ...ClientOption) (Client, error) {
 		httpClient:         opts.httpClient,
 		retryPolicyEnabled: opts.retryPolicyEnabled,
 		source:             source,
+		logger:             opts.logger,
+		piiLogging:         opts.piiLogging,
 	}
+
+	logOnce.Do(func() {
+		client.logger.Log(context.Background(), slog.LevelInfo, "Managed Identity", slog.String("source", string(client.source)))
```
chlowell, line 277: logging may entail network I/O, so this could still block indefinitely depending on the handler implementation. It would be better to do this in `AcquireToken`

apps/internal/slog/nop_handler.go
AndyOHart: As discussed, reverting to how it was as this may be too messy

PR #537, opened 2024-11-28
App service support for MI

4gust:
Added support for App service 
Added support for logging in Managed Identity


apps/logger/logger_118.go
```diff
@@ -0,0 +1,20 @@
+//go:build go1.18 && !go1.20
+
+package logger
```
bgavrilMS, line 3: Can you please split the logger in a separate commit? It can even go directly to main.

apps/logger/logger_118.go
```diff
@@ -0,0 +1,20 @@
...
+
+import "fmt"
+
+// Logger struct for Go versions <= 1.20.
+type Logger struct {
+	logCallback CallbackFunc
```
bgavrilMS, line 9: I was under the impression that Azure SDK will not be using the callback for now so there is no need for it. When they upgrade to 1.20, they will start using slog

apps/managedidentity/managedidentity.go
```diff
@@ -378,7 +424,6 @@ func (c Client) retry(maxRetries int, req *http.Request) (*http.Response, error)
 	var err error
 	for attempt := 0; attempt < maxRetries; attempt++ {
 		tryCtx, tryCancel := context.WithTimeout(req.Context(), time.Second*15)
```
bgavrilMS, line 426: Why 15s? I recommend 60s to use the same default as .NET

apps/managedidentity/managedidentity.go
```diff
@@ -390,14 +435,17 @@ func (c Client) retry(maxRetries int, req *http.Request) (*http.Response, error)
 			retrylist = retryCodesForIMDS
 		}
 		if err == nil && !contains(retrylist, resp.StatusCode) {
+			tryCancel()
```
bgavrilMS, line 438: How do these changes relate to app service?

apps/managedidentity/managedidentity.go
```diff
@@ -446,21 +494,55 @@ func (c Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRespons
...
+	if err != nil {
+		return nil, err
+	}
+	c.log(logger.Debug, fmt.Sprintf("Creating azure arc managed identity source Endpoint URI: %s", identityEndpoint))
+	req.Header.Set("X-IDENTITY-HEADER", os.Getenv(identityHeaderEnvVar))
+	q := req.URL.Query()
+	q.Set("api-version", appServiceAPIVersion)
+	q.Set("resource", resource)
+	switch t := id.(type) {
+	case UserAssignedClientID:
```
bgavrilMS, line 509: These are general purpose params, not related to App Service. So instead of logging them here, can we instead have logging statements at the start of the MSI "AcquireToken" request, which display all config values?

bgavrilMS: And similarly, after the HTTP request is done ... log failures and in case of success log things like "got an access token", expires in etc. Don't log the actual access token.

apps/logger/logger_120.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	return nil, fmt.Errorf("invalid input for Go 1.21+; expected *slog.Logger")
+}
+
+// Log method for Go 1.21+ with full support for structured logging and multiple log levels.
+func (a *Logger) Log(level Level, message string, fields ...any) {
+	if a.logging == nil {
+		return
+	}
+	var slogLevel slog.Level
```
bgavrilMS, line 29: It's unpleaseant to have to do this on every logging call :(.

bgavrilMS (approved):
Approved with comments.

apps/managedidentity/managedidentity.go
4gust: Azure SDK has it at 15 seconds, I will update it to 60

chlowell (changes requested):
Blockers are:
- query parameter names ([docs](https://learn.microsoft.com/azure/app-service/overview-managed-identity#rest-endpoint-reference))
- returned token expiration time is always 0
- fake token responses in tests don't resemble actual responses

Expiration time is always 0 because App Service responses include only `expires_on` ([docs](https://learn.microsoft.com/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#connect-to-azure-services-in-app-code)) and [MSAL unmarshals only `expires_in`](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/06ce6ba7393712f696f5df08cb4962da71581680/apps/internal/oauth/ops/accesstokens/tokens.go#L176), using a [custom UnmarshalJSON](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/06ce6ba7393712f696f5df08cb4962da71581680/apps/internal/json/types/time/time.go#L63) to convert that duration to an instant. Unit tests pass despite this because their fake responses [always include `expires_in`](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/8c5b978037085f57463b175cd755f903ffe4917b/apps/managedidentity/managedidentity_test.go#L55)

apps/managedidentity/managedidentity.go
```diff
@@ -446,7 +465,32 @@ func (c Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRespons
...
+	}
+	req.Header.Set("X-IDENTITY-HEADER", os.Getenv(identityHeaderEnvVar))
+	q := req.URL.Query()
+	q.Set("api-version", appServiceAPIVersion)
+	q.Set("resource", resource)
+	switch t := id.(type) {
+	case UserAssignedClientID:
+		q.Set(miQueryParameterClientId, string(t))
+	case UserAssignedResourceID:
+		q.Set(miQueryParameterResourceId, string(t))
```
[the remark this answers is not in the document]
4gust: Is this change only for App service ?

apps/managedidentity/managedidentity.go
```diff
@@ -446,7 +465,32 @@ func (c Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRespons
...
+	case UserAssignedObjectID:
+		q.Set(miQueryParameterObjectId, string(t))
```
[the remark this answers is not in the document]
4gust: I don't see this in .net or other msal's
Is this changed for App service only?

apps/managedidentity/managedidentity.go
```diff
@@ -290,27 +294,40 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
...
+	case AppService:
+		return acquireTokenForAppService(ctx, c, resource)
 	default:
 		return base.AuthResult{}, fmt.Errorf("unsupported source %q", c.source)
 	}
 }
 
-func acquireTokenForIMDS(ctx context.Context, client Client, resource string) (base.AuthResult, error) {
-	req, err := createIMDSAuthRequest(ctx, client.miType, resource)
+func acquireTokenForAppService(ctx context.Context, c Client, resource string) (base.AuthResult, error) {
```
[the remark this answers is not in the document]
4gust: Updated this to other functions as well.

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
chlowell: Yes. That issue is about supporting ACI, which imitates IMDS and requires m**s**i_res_id, in accordance with IMDS docs. App Service requires mi_res_id. Every platform has its own managed identity implementation, so subtle differences like this are normal

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
chlowell: Yes, but it appears I'm out of date on this point, as [the docs](https://learn.microsoft.com/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#rest-endpoint-reference) now state the API will accept `object_id` as an alias for `principal_id`. 🤷 I'd still make this change because Azure SDK uses, and tests, `principal_id`, and these docs have been incorrect before (i.e. if you want to keep using `object_id`, make sure you test it)

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
bgavrilMS: We were hoping to use `msi_res_id` everywhere. Seems like AppService supports it. Thoughts?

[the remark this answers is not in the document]
4gust: I will test this for on Azure Function with all the different user assigned ways and update this comment on the findings.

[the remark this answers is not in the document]
4gust: I did some testing, 
I Created a  App service function as assigned a `UserAssigned` managed Identity and a `SystemAssigned` managed identity.
Then I tried to access a `KeyVault` using the token we acquired.
Created 4 different client one for each type

> `SystemAssgiened()` -- No Querry parameter value in the request.
> `UserAssignedClientID("")` -- `client_id` client id was assigned to this parameter
> `UserAssignedObjectID("")` -- `object_id` Object id was assigned to this parameter
> `UserAssignedResourceID("")` -- `msi_res_id` Resource id was assigned to this parameter

Outcome -
I was able to get `AccessToken` (verified the token on the jwt.ms) from all the types of the managed identities, using that token I was able to get the secret that was stored in the `KeyVault`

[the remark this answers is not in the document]
bgavrilMS: I tend to agree with @chlowell on this one though. The documented parameter for App Service (and for all other sources except IMDS) is `mi_res_id` - see https://learn.microsoft.com/en-us/azure/app-service/overview-managed-identity?tabs=portal%2Chttp#rest-endpoint-reference

It'd be safer to use that and use `msi_res_id` only for IMDS.

apps/managedidentity/managedidentity.go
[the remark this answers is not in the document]
4gust: I tested with `object_id` and it worked, would you recommend that I also test with `principal_id`?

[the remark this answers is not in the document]
chlowell: You should run live tests before merging in any case

apps/managedidentity/managedidentity.go
```diff
@@ -390,14 +408,17 @@ func (c Client) retry(maxRetries int, req *http.Request) (*http.Response, error)
 			retrylist = retryCodesForIMDS
 		}
 		if err == nil && !contains(retrylist, resp.StatusCode) {
+			tryCancel()
```
chlowell, line 411: You deferred this above. Why call it here?

apps/internal/oauth/ops/accesstokens/tokens.go
```diff
@@ -183,6 +183,64 @@ type TokenResponse struct {
 	scopesComputed bool
 }
 
+func (tr *TokenResponse) UnmarshalJSON(data []byte) error {
+	type Alias TokenResponse
+	aux := &struct {
+		ExpiresIn json.Number `json:"expires_in"`
+		ExpiresOn json.Number `json:"expires_on"`
```
chlowell, line 190: You could simplify this method by continuing to use `internalTime.DurationTime` to calculate ExpiresOn from `expires_in`:
```suggestion
		ExpiresOn           json.Number               `json:"expires_on,omitempty"`
		ExpiresOnCalculated internalTime.DurationTime `json:"expires_in,omitempty"`
```
then after unmarshaling, if `aux.ExpiresOn` is nonzero, you can convert it to a `time.Time` for the `TokenResponse`. If that field is zero, `ExpiresOnCalculated.T` is the `time.Time` you want

apps/internal/json/types/time/time.go
```diff
@@ -59,12 +60,48 @@ func (d DurationTime) MarshalJSON() ([]byte, error) {
...
+	// Try parsing as Unix timestamp (seconds since the Unix epoch)
+	if len(str) == 10 {
+		if unixTimestamp, err := strconv.ParseInt(str, 10, 64); err == nil {
+			t.T = time.Unix(unixTimestamp, 0)
+			return nil
+		}
+	}
+
+	// Try parsing as ISO 8601 format (e.g., "2024-10-18T19:51:37.0000000+00:00")
+	iso8601Layout := "2006-01-02T15:04:05.9999999-07:00"
```
chlowell, line 77: Why do we need to parse these other formats?

apps/internal/oauth/ops/accesstokens/tokens.go
```diff
@@ -170,13 +170,14 @@ type TokenResponse struct {
...
-	IDToken        IDToken                   `json:"id_token"`
-	ClientInfo     ClientInfo                `json:"client_info"`
-	ExpiresOn      internalTime.DurationTime `json:"expires_in"`
-	ExtExpiresOn   internalTime.DurationTime `json:"ext_expires_in"`
-	GrantedScopes  Scopes                    `json:"scope"`
-	DeclinedScopes []string                  // This is derived
+	FamilyID            string                    `json:"foci"`
+	IDToken             IDToken                   `json:"id_token"`
+	ClientInfo          ClientInfo                `json:"client_info"`
+	ExpiresOnCalculated internalTime.DurationTime `json:"expires_in,omitempty"`
```
chlowell, line 176: This exposes the `expires_in` vs. `expires_on` distinction to consumers and makes them responsible for choosing which value to use. The type is then harder to use correctly and the extra complexity has no benefit--consumers just want to know when the token expires, they don't care how that was determined. Why not use `UnmarshalJSON` to set `ExpiresOn`, like you were before?

apps/internal/json/types/time/time.go
chlowell: Where are these other formats coming from? I haven't seen anything like them in a token response, excepting those from an old version of App Service we don't support

apps/internal/oauth/ops/accesstokens/tokens.go
```diff
@@ -173,14 +174,80 @@ type TokenResponse struct {
 	FamilyID       string                    `json:"foci"`
 	IDToken        IDToken                   `json:"id_token"`
 	ClientInfo     ClientInfo                `json:"client_info"`
-	ExpiresOn      internalTime.DurationTime `json:"expires_in"`
+	ExpiresOn      internalTime.DurationTime `json:"-"`
```
chlowell, line 177: DurationTime exists to calculate ExpiresOn from `expires_in`. That doesn't help now that you want to consider a possible value of `expires_on` as well, and you're disabling unmarshaling for this field anyway. So, why not make the type here simply `time.Time`?

apps/internal/oauth/ops/accesstokens/tokens.go
```diff
@@ -173,14 +174,80 @@ type TokenResponse struct {
...
+	}{
+		Alias: (*Alias)(tr),
+	}
+
+	// Unmarshal the JSON data into the aux struct
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	parseDuration := func(num json.Number) (int64, error) {
```
chlowell, line 201: How is this better than `DurationTime.UnmarshalJSON` i.e., what do you gain by unmarshaling `expires_in` to `json.Number` instead of `DurationTime`?

apps/internal/oauth/ops/accesstokens/tokens.go
```diff
@@ -173,14 +174,80 @@ type TokenResponse struct {
...
+		if num == "" {
+			return 0, nil
+		}
+		return num.Int64()
+	}
+
+	// Function to parse different date formats
+	parseExpiresOn := func(expiresOn string) (time.Time, error) {
```
bgavrilMS, line 209: Pls add a comment pointing at https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4963 to explain when this can occur.

chlowell: Maybe I'm overlooking it, but that issue doesn't mention the platform returning the unsupported format

bgavrilMS: The CX who reported this didn't make it very clear, it seems to be container + App Service.

chlowell: It would be good to know what the linked PR added support for so that support is easier to maintain. Setting that aside, the linked PR added ISO 8601 parsing. Why does this PR have two additional formats?

bgavrilMS: We asked, the CX didn't answer. Ultimately, Azure SDK for .NET did not have this problem so it was clearly a regression. 

Not objecting to supporting only ISO 8601. The .NET code uses a general-purpose DateTime parser btw https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4964/files#diff-a7d2338916806e48fb637fd9caad6095fec8537a96d157bc2ea47ab64903fef9R93

I assume Azure.SDK for .NET did the same.

apps/internal/oauth/ops/accesstokens/tokens.go
4gust: updated

apps/internal/oauth/ops/accesstokens/tokens.go
4gust: shifted this to `internalTime.DurationTime`

AndyOHart (approved):
LGTM

apps/managedidentity/managedidentity.go
```diff
@@ -446,6 +465,31 @@ func (c Client) getTokenForRequest(req *http.Request) (accesstokens.TokenRespons
...
+	}
+	req.Header.Set("X-IDENTITY-HEADER", os.Getenv(identityHeaderEnvVar))
+	q := req.URL.Query()
+	q.Set("api-version", appServiceAPIVersion)
+	q.Set("resource", resource)
+	switch t := id.(type) {
+	case UserAssignedClientID:
+		q.Set(miQueryParameterClientId, string(t))
+	case UserAssignedResourceID:
+		q.Set(miQueryParameterPrincipalId, string(t))
```
chlowell, line 482:
```suggestion
		q.Set(miQueryParameterResourceId, string(t))
```

chlowell (approved):

PR #542, opened 2025-01-09
Support for refresh_in for ConfidentialClient

4gust:
### PR Description:

This update introduces support for a `refreshIn` field, which is calculated as half of the `expiresIn` value, but only if `expiresIn` is greater than 2 hours and if it was not received in the response. 

Key changes:
- **When storing data**: If `expiresIn` is more than 2 hours and not provided in the response, `refreshIn` will be set to half of `expiresIn`.
- **When retrieving from cache**: We now check the `refreshIn` value. If it has expired, a server call will be triggered on the main thread to refresh the data.

This ensures that cache data is refreshed proactively, reducing the need for repeated server calls while improving overall performance.


apps/internal/base/base.go
```diff
@@ -86,6 +86,7 @@ type AuthResult struct {
 	Account        shared.Account
 	IDToken        accesstokens.IDToken
 	AccessToken    string
+	RefreshIn      time.Time
```
bgavrilMS, line 89: I agree with adding this to the public API, but for consistency with `ExpiresOn` it should be added as `RefreshOn`

apps/internal/base/internal/storage/items.go
```diff
@@ -72,6 +72,7 @@ type AccessToken struct {
 	ClientID          string            `json:"client_id,omitempty"`
 	Secret            string            `json:"secret,omitempty"`
 	Scopes            string            `json:"target,omitempty"`
+	RefreshIn         internalTime.Unix `json:"refresh_in,omitempty"`
```
bgavrilMS, line 75: We need to keep the cache consistent between MSALs and the property for refresh in `RefreshOn` (similar to ExpiresOn)

apps/internal/oauth/ops/accesstokens/accesstokens_test.go
```diff
@@ -778,6 +778,50 @@ func TestTokenResponseUnmarshal(t *testing.T) {
 			},
 			jwtDecoder: jwtDecoderFake,
 		},
+		{
+			desc: "Success",
+			payload: `
+				{
+					"access_token": "secret",
```
bgavrilMS, line 785: Did you want to add refresh_in here?

apps/internal/oauth/ops/accesstokens/accesstokens_test.go
```diff
@@ -778,6 +778,50 @@ func TestTokenResponseUnmarshal(t *testing.T) {
...
+				GrantedScopes: Scopes{Slice: []string{"openid", "profile"}},
+				ClientInfo: ClientInfo{
+					UID:  "uid",
+					UTID: "utid",
+				},
+			},
+			jwtDecoder: jwtDecoderFake,
+		},
+		{
+			desc: "Success",
```
bgavrilMS, line 804: seems identical to the above?

apps/internal/oauth/ops/accesstokens/tokens.go
```diff
@@ -183,6 +184,32 @@ type TokenResponse struct {
...
+		*Alias
+	}{
+		Alias: (*Alias)(t),
+	}
+
+	if err := json.Unmarshal(data, &aux); err != nil {
+		return err
+	}
+
+	// If refresh_in is not set, compute it as (now - expires_on) / 2 if expires_on > 2 hours from now
```
bgavrilMS, line 199: This is wrong. Imagine: 

1. I get a token with expires_on = 4h from AAD. It's 1 PM.
2. I get later need that token, I get it from cache. It's 2 PM (so 3h expiration remaining). I now get refresh_in = 1.5h
3. I need the token later too. It's 4 PM (so 1h expiration remaining). I now get refresh_in = nil. 

What should actually happen is: 

- at step 1, save the computed refresh_in value in the cache. 
- at step 2 and step 3, evaluate if a new token from AAD is needed.

bgavrilMS (changes requested):
This only takes care (partially) of adding RefreshOn in the cache. However, the feature is also meant to provide proactive token refresh capabilities, meaning: 

- when app dev requests a token
- if RefreshOn exists AND it has elasped AND expiresON has not elapsed
- make a request to AAD for a new token
- and if that request fails AAD is unavailable (HTTP 5xx error or HTTP 429 error), return the old token

apps/internal/base/base.go
4gust: Changed it to `refreshOn`

apps/internal/mock/mock.go
```diff
@@ -94,6 +94,23 @@ func GetAccessTokenBody(accessToken, idToken, refreshToken, clientInfo string, e
 	return []byte(body)
 }
 
+func GetAccessTokenBodyWithRefreshIn(accessToken, idToken, refreshToken, clientInfo string, expiresIn int, refreshIn int) []byte {
```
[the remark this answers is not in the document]
4gust: I will change this

apps/internal/base/base.go
```diff
@@ -345,6 +350,26 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	if silent.Claims == "" {
 		ar, err = AuthResultFromStorage(storageTokenResponse)
 		if err == nil {
+			if shouldRefresh(storageTokenResponse.AccessToken.RefreshOn.T) {
+				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
bgavrilMS, line 354: Why 10 seconds instead of obeying the general context cancellation? Since refresh happens on the same thread as the request, it makes sense to follow normal rules there.

apps/internal/base/base.go
```diff
@@ -345,6 +350,26 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+				defer cancel()
+				if tr, er := b.Token.Credential(ctx, authParams, silent.Credential); er == nil {
+					return b.AuthResultFromToken(ctx, authParams, tr, true)
```
bgavrilMS, line 357: nit: Is there any code path where `AuthResultFromToken` is called with the "cacheWrite" flag set to false? If not, consider getting rid of that param.

apps/internal/base/base.go
```diff
@@ -345,6 +350,26 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+				} else if callErr, ok := er.(*errors.CallErr); ok {
+					// Check if the error is of type CallErr and matches the relevant status codes
+					switch callErr.Resp.StatusCode {
+					case http.StatusRequestTimeout, // 408
+						http.StatusTooManyRequests,     // 429
+						http.StatusInternalServerError, // 500
+						http.StatusBadGateway,          // 502
+						http.StatusServiceUnavailable,  // 503
+						http.StatusGatewayTimeout:      // 504
```
bgavrilMS, line 366: nit: this was a bit hard to follow for me. Copilot tells me that these switch cases will result in an execution continuation, i.e. line 373 gets executed, where ar is from the cache.

apps/internal/exported/exported.go
```diff
@@ -31,4 +31,6 @@ type TokenProviderResult struct {
 	AccessToken string
 	// ExpiresInSeconds is the lifetime of the token in seconds
 	ExpiresInSeconds int
+	// RefreshInSeconds indicates the suggested	time to refresh the token, if any
+	RefreshInSeconds int
```
bgavrilMS, line 35: nit: this extensiblity point is really for Azure SDK only, so that they can plug in their MSI logic and benefit from MSAL caching. We don't have to add this here, but it's ok either way.

apps/internal/oauth/ops/accesstokens/accesstokens_test.go
```diff
@@ -757,18 +757,66 @@ func TestTokenResponseUnmarshal(t *testing.T) {
...
+			},
+			jwtDecoder: jwtDecoderFake,
+		},
+		{
+			desc: "Success with different expires_in and refresh On",
+			payload: `
+				{
+					"access_token": "secret",
+					"expires_in": 3600,
+					"refresh_in": 43199,
```
bgavrilMS, line 789: To mimic eSTS behavior, I recommend using expires_in = 24h and refresh_in = 12h in tests.

apps/confidential/confidential_test.go
```diff
@@ -138,6 +138,7 @@ func TestAcquireTokenByCredential(t *testing.T) {
 		}
 		client, err := fakeClient(accesstokens.TokenResponse{
 			AccessToken:   token,
+			RefreshOn:     internalTime.DurationTime{T: time.Now().Add(1 * time.Hour)},
```
bgavrilMS, line 141: This is not a realistic refresh_in value. eSTS will set this at 1/2 expires_in value.

apps/internal/exported/exported.go
bgavrilMS: But if you do add it, we need tests for it. And you need to plug it in.

apps/confidential/confidential_test.go
bgavrilMS: Comment not resolved.

apps/confidential/confidential_test.go
```diff
@@ -255,7 +257,7 @@ func TestAcquireTokenOnBehalfOf(t *testing.T) {
 	// TODO: OBO does instance discovery twice before first token request https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/351
 	mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, tenant)))
 	mockClient.AppendResponse(mock.WithBody(mock.GetTenantDiscoveryBody(lmo, tenant)))
-	mockClient.AppendResponse(mock.WithBody(mock.GetAccessTokenBody(token, "", "rt", "", 3600)))
+	mockClient.AppendResponse(mock.WithBody(mock.GetAccessTokenBodyWithRefreshIn(token, "", "rt", "", 3600, 7400)))
```
bgavrilMS, line 260: not good values. expires In should be < refresh in

apps/confidential/confidential_test.go
```diff
@@ -744,6 +748,9 @@ func TestNewCredFromTokenProvider(t *testing.T) {
 	if v := int(time.Until(ar.ExpiresOn).Seconds()); v < expiresIn-2 || v > expiresIn {
 		t.Fatalf("expected ExpiresOn ~= %d seconds, got %d", expiresIn, v)
 	}
+	if v := int(time.Until(ar.Metadata.RefreshOn).Seconds()); v < refreshOn-2 || v > refreshOn {
```
bgavrilMS, line 751: Why do you need this tolerance here? And if one is needed, should it live in some helper to benefit all tests and to keep DRY?

apps/confidential/confidential_test.go
```diff
@@ -756,6 +763,89 @@ func TestNewCredFromTokenProvider(t *testing.T) {
 	}
 }
 
+func TestRefreshIn(t *testing.T) {
```
bgavrilMS, line 766: Nice test!

apps/internal/base/base.go
```diff
@@ -458,6 +481,15 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 	return ar, err
 }
 
+// This function wraps time.Now() and is used for refreshing the application
+// was created to test the function against refreshin
+var GetCurrentTime = time.Now
```
bgavrilMS, line 486: 1. Making sure this isn't part of the public API?
2. Does it make sense to keep it higher up?

apps/internal/base/base.go
```diff
@@ -458,6 +483,11 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 	return ar, err
 }
 
+// shouldRefresh returns true if the token should be refreshed.
+func shouldRefresh(t time.Time) bool {
+	return !t.IsZero() && t.Before(time.Now().Add(2*time.Hour))
```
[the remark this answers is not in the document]
4gust: removed this!

apps/internal/oauth/ops/accesstokens/accesstokens_test.go
4gust: done

apps/confidential/confidential_test.go
4gust: Updated

apps/internal/base/base.go
4gust: There is no other way to do that, I will consult with Charles on this again.

apps/internal/exported/exported.go
bgavrilMS: Sorry if I wasn't clear - pls remove it or add tests for this. But please reply to the comment to help me (as reviewer) know which path you chose.

apps/internal/mock/mock.go
```diff
@@ -46,6 +46,12 @@ func WithCallback(callback func(*http.Request)) responseOption {
 	})
 }
 
+func WithCode(code int) responseOption {
```
bgavrilMS, line 49: nit: WithHttpStatusCode

apps/internal/exported/exported.go
4gust: Apologies, I will be more prompt with the response on the comments here,
I am adding test for this.
We cannot remove this from exported as the same contract is used for `base.Client` as well as `Confidential.Client`

apps/internal/oauth/ops/accesstokens/accesstokens_test.go
4gust: Added

apps/confidential/confidential_test.go
```diff
@@ -756,6 +764,97 @@ func TestNewCredFromTokenProvider(t *testing.T) {
...
+				t.Fatal("should have returned from cache.")
+			}
+			if (ar.AccessToken == secondToken) != tt.shouldGetNewToken {
+				t.Fatalf("wanted %q, got %q", secondToken, ar.AccessToken)
+			}
+		})
+	}
+}
+
+func isTimeSame(t time.Time, expectedSeconds int) bool {
```
chlowell, line 850:
```
if A {
  return false
}
return true
```
simplifies to `return !A`

apps/internal/base/base.go
```diff
@@ -83,9 +84,10 @@ type AcquireTokenOnBehalfOfParameters struct {
 // AuthResult contains the results of one token acquisition operation in PublicClientApplication
 // or ConfidentialClientApplication. For details see https://aka.ms/msal-net-authenticationresult
 type AuthResult struct {
-	Account        shared.Account
-	IDToken        accesstokens.IDToken
-	AccessToken    string
+	Account     shared.Account
+	IDToken     accesstokens.IDToken
+	AccessToken string
+	//RefreshOn indicates the recommended time to request a new access token, or zero if no refresh time is suggested
```
chlowell, line 90: 👀

apps/internal/base/base.go
```diff
@@ -345,6 +350,24 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	if silent.Claims == "" {
 		ar, err = AuthResultFromStorage(storageTokenResponse)
 		if err == nil {
+			if shouldRefresh(storageTokenResponse.AccessToken.RefreshOn.T) {
+				if tr, er := b.Token.Credential(ctx, authParams, silent.Credential); er == nil {
+					return b.AuthResultFromToken(ctx, authParams, tr)
+				} else if callErr, ok := er.(*errors.CallErr); ok {
```
chlowell, line 356: The type assertion is probably okay but errors.As would be more robust. Also, you don't need the else, because you return when the preceding if evaluates true

apps/internal/base/base.go
```diff
@@ -345,6 +350,24 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+					// Check if the error is of type CallErr and matches the relevant status codes
```
chlowell, line 357: this says literally what the code does but not why the code does it, which is the interesting and perhaps mysterious part. Consider commenting instead in the empty case below with something like "refreshing failed with a non-fatal error; return the cached token"

apps/internal/base/base.go
```diff
@@ -345,6 +350,24 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+					switch callErr.Resp.StatusCode {
+					case http.StatusRequestTimeout, // 408
+						http.StatusTooManyRequests,     // 429
+						http.StatusInternalServerError, // 500
+						http.StatusBadGateway,          // 502
+						http.StatusServiceUnavailable,  // 503
+						http.StatusGatewayTimeout:      // 504
+					default:
+						// Handle non-retryable errors
+						return AuthResult{}, er
```
chlowell, line 367: Why not return the valid token from the cache?

apps/internal/base/base.go
chlowell: While this is exported i.e. public and therefore accessible from other packages, it's in an internal package. Application code can't reference any internal package

apps/internal/base/base.go
4gust: Updated

apps/internal/base/base.go
4gust: Updated

apps/internal/base/base.go
4gust: Removed

apps/internal/base/base.go
chlowell: @bgavrilMS can you please explain the reasoning for this behavior? What's the benefit of MSAL returning an error to the application when it has a valid token it could return instead?

apps/internal/base/base.go
```diff
@@ -345,6 +349,26 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+						switch callErr.Resp.StatusCode {
+						case http.StatusRequestTimeout, // 408
+							http.StatusTooManyRequests,     // 429
+							http.StatusInternalServerError, // 500
+							http.StatusBadGateway,          // 502
+							http.StatusServiceUnavailable,  // 503
+							http.StatusGatewayTimeout:      // 504
+						default:
+							// return empty token for non handable error
+							return AuthResult{}, er
```
bgavrilMS, line 367: We should never return empty token. Instead, we should throw the error associated with the failure.

chlowell: This is the idiomatic equivalent to throwing--return the error and the zero value of any other types the function returns. Callers are responsible for checking whether the returned error is nil and if it isn't, they should discard anything else the function returned.

I don't understand why we'd throw here though. We could return the valid token from the cache. The fact that a refresh failed is interesting only when the application really needs a new token.

bgavrilMS (approved):

apps/internal/base/base.go
bgavrilMS: Yes, sorry, my bad for forgetting the golang pattern. 

With regards to you question:

1. We don't want to keep retrying getting a token in all cases. For example if the client secret is bad, we shouldn't silently retry that. It would generate a lot of extra traffic for eSTS and the SDK's responsibility is to minimize that. refresh_in triggers only eSTS has a problem and they signal this by sending those status codes 
2. refresh_in is only issued by eSTS for some tenants where long lived tokens are enabled. So the token lifetime is much longer (~24h for app tokens) anyway, it won't cause a problem

chlowell: But this behavior will cause the application to fail when it could have succeeded given the cached token. Note that as written, the application gets an error when the authority doesn't respond, so the cache no longer makes the client resilient against transient network outages. There are other ways to avoid hammering the authority; we need backoff and synchronization here in any case. It will be rare for a client to get this far with an invalid secret, because in most usage the cache is in memory and begins empty, but we could handle that case by its error code.

apps/internal/base/base.go
```diff
@@ -345,6 +349,26 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	if silent.Claims == "" {
 		ar, err = AuthResultFromStorage(storageTokenResponse)
 		if err == nil {
+			if shouldRefresh(storageTokenResponse.AccessToken.RefreshOn.T) {
+				if tr, er := b.Token.Credential(ctx, authParams, silent.Credential); er == nil {
```
chlowell, line 353: This assumes the caller is a confidential client, but actually public clients call this method as well. Because they pass `silent.Credential == nil`, they panic when the cached token should be refreshed.

apps/internal/base/base.go
```diff
@@ -345,6 +349,26 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	if silent.Claims == "" {
 		ar, err = AuthResultFromStorage(storageTokenResponse)
 		if err == nil {
+			if shouldRefresh(storageTokenResponse.AccessToken.RefreshOn.T) {
```
chlowell, line 352: This must be synchronized so that only one goroutine at a time tries to refresh the token, and there should be some backoff--respecting any Retry-After header--between refresh attempts

chlowell (changes requested):

apps/internal/base/base.go
```diff
@@ -169,6 +170,8 @@ type Client struct {
 	AuthParams      authority.AuthParams // DO NOT EVER MAKE THIS A POINTER! See "Note" in New().
 	cacheAccessor   cache.ExportReplace
 	cacheAccessorMu *sync.RWMutex
+	canRefresh      *atomic.Value
+	refreshMu       *sync.Mutex
```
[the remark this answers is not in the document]
4gust: Noted, I will update

apps/internal/base/base_test.go
```diff
@@ -444,3 +443,40 @@ func TestAuthResultFromStorage(t *testing.T) {
 		}
 	}
 }
+
+// TestShouldRefresh tests the shouldRefresh function
```
AndyOHart, line 447: Not sure if we need this comment

AndyOHart (approved):
LGTM

apps/internal/base/base.go
```diff
@@ -351,12 +352,20 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
-				b.removeTenantFromCanRefresh(tenant)
-				defer b.refreshMu.Unlock()
-				if tr, er := b.Token.Credential(ctx, authParams, silent.Credential); er == nil {
-					return b.AuthResultFromToken(ctx, authParams, tr)
+				if _, exists := b.canRefresh[tenant]; !exists {
+					var empty atomic.Value
+					empty.Store(false)
+					b.canRefresh[tenant] = &empty
+				}
+				refreshValue := b.canRefresh[tenant]
```
chlowell, line 362: you can do this with one read
```suggestion
				refreshValue, ok := b.canRefresh[tenant]
				if !ok {
					refreshValue = &atomic.Value{}
					refreshValue.Store(false)
					b.canRefresh[tenant] = refreshValue
				}
```
(also, `ok` is the idiomatic name for the second variable in this pattern, which is called "comma ok")

apps/internal/base/base.go
```diff
@@ -473,31 +482,9 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
...
-	if !b.doesTenantExists(tId) {
-		b.refreshMu.RUnlock()
-		b.refreshMu.Lock()
-		defer b.refreshMu.Unlock()
-		b.addTenantIntoCanRefresh(tId)
-		return !t.IsZero() && t.Before(GetCurrentTime())
-	}
-	b.refreshMu.RUnlock()
-	return false
+func (b *Client) shouldRefresh(t time.Time) bool {
```
chlowell, line 486: I'd inline this because it's trivial. If you still don't want to inline it, consider making it a plain old function because it doesn't need to reference `Client`

apps/internal/mock/syncmock.go
chlowell, line 1: Consider instead adding a mutex to the existing mock. Its current raciness is not a feature 😄

apps/managedidentity/managedidentity.go
```diff
@@ -249,7 +249,8 @@ func New(id ID, options ...ClientOption) (Client, error) {
 	default:
 		return Client{}, fmt.Errorf("unsupported type %T", id)
 	}
-	var zero atomic.Int32
+	var zero atomic.Value = atomic.Value{}
```
chlowell, line 252:
```suggestion
	zero := atomic.Value{}
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -1131,3 +1132,118 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+						requestChecker = true
+					}
+				}
+			}()
+		}
+		// Waiting for all goroutines to finish
+		go func() {
+			wg.Wait()
+			if !requestChecker {
+				t.Error("Error should be called at least once")
```
chlowell, line 1205: why launch a goroutine to do this?

apps/managedidentity/managedidentity_test.go
```diff
@@ -1131,3 +1132,118 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+func TestRefreshInMultipleRequests(t *testing.T) {
+	firstToken := "first token"
+	secondToken := "new token"
+	refreshIn := 43200
+	expiresIn := 86400
+	resource := "https://resource/.default"
+	miType := SystemAssigned()
+	setEnvVars(t, CloudShell)
+
+	t.Run("Test for refresh multiple request", func(t *testing.T) {
```
chlowell, line 1145: `t.Run` doesn't add any value because there are no other subtests

apps/internal/mock/mock.go
```diff
@@ -90,11 +90,19 @@ func (c *Client) Do(req *http.Request) (*http.Response, error) {
...
+func GetAccessTokenBody(accessToken, idToken, refreshToken, clientInfo string, expiresIn, refreshIn int) []byte {
+	// Start building the body with the common fields
 	body := fmt.Sprintf(
 		`{"access_token": "%s","expires_in": %d,"expires_on": %d,"token_type": "Bearer"`,
 		accessToken, expiresIn, time.Now().Add(time.Duration(expiresIn)*time.Second).Unix(),
 	)
+
+	// Conditionally add the "refresh_in" field if refreshIn is provided
+	if refreshIn > 0 {
+		body += fmt.Sprintf(`, "refresh_in": %d`, refreshIn)
```
chlowell, line 102: is `refresh_in` always a number, or could it be a string?

apps/internal/base/base.go
4gust: Noted, will use `ok`

apps/internal/base/base.go
4gust: Updated

apps/internal/mock/mock.go
4gust: Always a number, same as `expires_in`

apps/internal/mock/syncmock.go
4gust: Removed :), updated our client

apps/managedidentity/managedidentity.go
4gust: updated

apps/managedidentity/managedidentity_test.go
4gust: Removed

apps/managedidentity/managedidentity_test.go
4gust: Removed

bgavrilMS (approved):

apps/internal/mock/mock.go
chlowell: expires_in can be a string though. I'm referring to JSON value types here i.e., is refresh_in always a number like `42` or can it be a string like `"42"`

apps/managedidentity/managedidentity_test.go
```diff
@@ -1131,3 +1132,113 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+		go func() {
+			defer wg.Done()
+			ar, err := client.AcquireToken(context.Background(), resource)
+			if err != nil {
+				t.Error(err)
+				return
+			}
+			if ar.AccessToken == secondToken && ar.Metadata.TokenSource == base.IdentityProvider {
+				if requestChecker {
+					t.Error("Error can only call this only once")
```
chlowell, line 1193: a downside of allowing every goroutine to output an error from `go test` is that you could get 9,999 error messages and only the first would be interesting. Consider using a channel to limit the number of reported errors, [for example](https://github.com/Azure/azure-sdk-for-go/blob/01e85d2ec16d40200652aabd3e931a776deb26d8/sdk/azidentity/cache/new_test.go#L50-L68)

apps/internal/mock/mock.go
4gust: Updated, tested it and it works fine both ways :)

apps/confidential/confidential_test.go
```diff
@@ -805,6 +814,213 @@ func TestNewCredFromTokenProvider(t *testing.T) {
...
+				}
+			}
+		}()
+	}
+	// Wait for all goroutines in a separate goroutine
+	go func() {
+		wg.Wait()
+		close(done)
+		close(ch)
+	}()
```
chlowell, line 924: Why do this in a separate goroutine, and with a `done` channel? You want the goroutine executing the test to wait for the others--it can just call `wg.Wait()` as before. Maybe this will help explain the mechanics: https://play.golang.com/p/hWAE3Cenyd8

4gust: I created the go routine, so that we can safely close the channels and create a `wait()`
I understand what i missed in there
I updated it

apps/confidential/confidential_test.go
```diff
@@ -805,6 +814,207 @@ func TestNewCredFromTokenProvider(t *testing.T) {
...
+			defer wg.Done()
+			ar, err := client.AcquireTokenSilent(context.Background(), tokenScope, WithTenantID("firstTenant"))
+			if err != nil {
+				select {
+				case ch <- err:
+				default:
+				}
+				return
+			}
+			if ar.AccessToken == secondToken+"firstTenant" && ar.Metadata.TokenSource == base.IdentityProvider {
```
chlowell, line 892: 🤔 this test is unclear and maybe incorrect. How to clarify or fix it depends on what behaviors you want to verify. I suppose it's a subset of these?
- proactive refreshes in general don't race
- client refreshes different tenants concurrently
- concurrent refreshes in particular don't race

apps/managedidentity/managedidentity.go
```diff
@@ -434,6 +464,11 @@ func authResultFromToken(authParams authority.AuthParams, token accesstokens.Tok
 	if err != nil {
 		return base.AuthResult{}, err
 	}
+	// if refreshOn is not set, set it to half of the time until expiry if expiry is more than 2 hours away
+	timeUntilExpiry := time.Until(token.ExpiresOn)
+	if token.RefreshOn.T.IsZero() && timeUntilExpiry > (time.Hour*2) {
+		token.RefreshOn.T = time.Now().Add(timeUntilExpiry / 2)
+	}
```
chlowell, line 471: please forgive me nitpicking on something with negligible practical impact but this unnecessarily computes and throws away `timeUntilExpiry` when `!token.RefreshOn.T.IsZero()`
```suggestion
	if token.RefreshOn.T.IsZero() {
		if lifetime := time.Until(token.ExpiresOn); lifetime > 2*time.Hour {
			token.RefreshOn.T = time.Now().Add(lifetime / 2)
		}
	}
```

Also, we should test this behavior

apps/internal/base/base.go
```diff
@@ -457,6 +478,15 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 	return ar, err
 }
 
+// This function wraps time.Now() and is used for refreshing the application
+// was created to test the function against refreshin
+var GetCurrentTime = time.Now
+
+// shouldRefresh returns true if the token should be refreshed.
+func shouldRefresh(t time.Time) bool {
+	return !t.IsZero() && t.Before(GetCurrentTime())
```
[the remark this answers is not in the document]
4gust: Removed the method, but we need the `isZero` check.

apps/managedidentity/managedidentity.go
```diff
@@ -291,6 +296,21 @@ func GetSource() (Source, error) {
...
+// shouldRefresh returns true if the token should be refreshed.
+func (b Client) shouldRefresh(t time.Time) bool {
+	if t.IsZero() || t.After(getCurrentTime()) {
+		return false
+	}
+	if !b.canRefresh.CompareAndSwap(false, true) {
+		return false
+	}
+	return true
+}
```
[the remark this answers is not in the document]
4gust: Removed the method, but we need the `isZero` check.
As, when we are setting this with `timeRemaining/2` that is only when its less then 2 hours.

apps/managedidentity/managedidentity.go
4gust: updated

apps/internal/base/base.go
```diff
@@ -457,6 +478,15 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 	return ar, err
 }
 
+// This function wraps time.Now() and is used for refreshing the application
+// was created to test the function against refreshin
+var GetCurrentTime = time.Now
```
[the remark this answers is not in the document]
4gust: updated

apps/confidential/confidential_test.go
```diff
@@ -916,22 +916,16 @@ func TestRefreshInMultipleRequests(t *testing.T) {
...
 	case err := <-ch:
 		t.Fatal(err)
 	default:
 	}
-	// Wait for all goroutines to complete
-	<-done
 	if !secondTenantChecker && !firstTenantChecker {
 		t.Error("Error should be called at least once")
 	}
+	close(ch)
```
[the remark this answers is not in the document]
4gust: Noted and updated

apps/managedidentity/managedidentity_test.go
```diff
@@ -1131,3 +1132,121 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+			ar, err := client.AcquireToken(context.Background(), resource)
+			if err != nil {
+				select {
+				case ch <- err:
+				default:
+				}
+				return
+			}
+			if ar.AccessToken == secondToken && ar.Metadata.TokenSource == base.IdentityProvider {
+				if requestChecker {
```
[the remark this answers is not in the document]
4gust: This checks two things 

1. Have it token updated.
2. The updated token we got was not from cache.

In these scenarios should only happen once. If we don't check this then how do, we know that the request to the `mockClient` has been made or not. As, this is concurrent we can get the updated token from cache before the token from `Identity`. This check is useful. 
The request can only be made once, as there is only one request in queue.

apps/confidential/confidential_test.go
```diff
@@ -925,7 +914,96 @@ func TestRefreshInMultipleRequests(t *testing.T) {
...
+	)
+
+	wg.Add(1)
+	wgComeplete.Add(1)
+	go func() {
+		ar, err := client.AcquireTokenSilent(context.Background(), tokenScope, WithTenantID("firstTenant"))
+		if err != nil {
+			t.Error("Unexpected error", err)
+		}
+		wg.Wait()
```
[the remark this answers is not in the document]
4gust: Updated the code

apps/internal/mock/mock.go
```diff
@@ -62,10 +63,13 @@ func WithHTTPStatusCode(statusCode int) responseOption {
 
 // Client is a mock HTTP client that returns a sequence of responses. Use AppendResponse to specify the sequence.
 type Client struct {
+	mu   sync.Mutex
```
chlowell, line 66: A field with a Mutex value (i.e. `sync.Mutex` not `*sync.Mutex`) is safe only when instances of the struct are never copied because copying the struct would copy the Mutex but not its internal state i.e., its lockedness. This can create subtle bugs. For example, it means that methods with value receivers wouldn't synchronize with other methods. Client has no such methods and I doubt it's copied anywhere because that would be weird, but this type now comes with a footgun that deserves a comment like "Clients must never be copied because they have a Mutex value"

apps/confidential/confidential_test.go
```diff
@@ -967,43 +967,53 @@ func TestConcurrentRequests(t *testing.T) {
...
-		mock.WithBody([]byte(fmt.Sprintf(`{"access_token":%q,"expires_in":%d,"refresh_in":%d,"token_type":"Bearer"}`, secondToken+"secondTenant", expiresIn, refreshIn+44200))),
-	)
-
-	mockClient.AppendResponse(
-		mock.WithBody([]byte(fmt.Sprintf(`{"access_token":%q,"expires_in":%d,"refresh_in":%d,"token_type":"Bearer"}`, secondToken+"firstTenant", expiresIn, refreshIn+44200))),
-		mock.WithCallback(func(r *http.Request) {
-			time.Sleep(2 * time.Second)
-		}),
-	)
+	present := make(chan string, 2)
```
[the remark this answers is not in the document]
4gust: Making it fix sized

apps/confidential/confidential_test.go
```diff
@@ -805,6 +814,295 @@ func TestNewCredFromTokenProvider(t *testing.T) {
...
+		return fixedTime
+	}
+	var wg sync.WaitGroup
+	ch := make(chan error, 1)
+	firstTenantChecker := false
+	secondTenantChecker := false
+	mockClient.AppendResponse(
+		mock.WithBody([]byte(fmt.Sprintf(`{"access_token":%q,"expires_in":%d,"refresh_in":%d,"token_type":"Bearer"}`, secondToken+"firstTenant", expiresIn, refreshIn+44200))))
+	mockClient.AppendResponse(
+		mock.WithBody([]byte(fmt.Sprintf(`{"access_token":%q,"expires_in":%d,"refresh_in":%d,"token_type":"Bearer"}`, secondToken+"secondTenant", expiresIn, refreshIn+44200))))
```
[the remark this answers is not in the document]
4gust: interesting, 
but are we not blocking the response until the `secondTenant` is served ?
```go
mock.WithCallback(func(r *http.Request) {
    if s := strings.SplitN(r.URL.Path, "/", 3); len(s) == 3 {
	    // send the tenant from the URL, which uniquely identifies the goroutine
	    present <- s[1]
	    <-proceed
    } else {
	    t.Error("unexpected token request URL: " + r.URL.String())
    }
  }),
```
where we are sending the result to present only then we let the `proceed` go in

apps/internal/mock/mock.go
4gust: Updating this.

apps/confidential/confidential_test.go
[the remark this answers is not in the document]
4gust: Updated the test with this, as this is simpler.

apps/confidential/confidential_test.go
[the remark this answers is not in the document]
chlowell: what do you mean? This buffer enables up to two non-blocking sends between receives; a hypothetical third send would block. How is that useful?

apps/confidential/confidential_test.go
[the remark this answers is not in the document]
4gust: I miss read the comment, i understand what the problem was, Fixing it

chlowell (commented):
blocker is exporting `managedidentity.Now`

apps/managedidentity/managedidentity.go
```diff
@@ -291,6 +296,15 @@ func GetSource() (Source, error) {
 	return DefaultToIMDS, nil
 }
 
+// This function wraps time.Now() and is used for refreshing the application
+// was created to test the function against refreshin
+var Now = time.Now
```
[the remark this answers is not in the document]
4gust: updating it.

apps/internal/mock/mock.go
```diff
@@ -62,10 +63,16 @@ func WithHTTPStatusCode(statusCode int) responseOption {
 
 // Client is a mock HTTP client that returns a sequence of responses. Use AppendResponse to specify the sequence.
 type Client struct {
+	mu   *sync.Mutex
 	resp []response
 }
 
+func NewClient() *Client {
+	return &Client{mu: &sync.Mutex{}}
+}
```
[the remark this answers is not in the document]
4gust: Updated

chlowell (approved):

PR #544, opened 2025-01-16
CloudShell Service

AndyOHart:
- Introduced support for Cloud Shell service in the Microsoft Authentication Library for Go.
- Refactored existing code to improve readability and maintainability.
- Added unit tests to ensure the new functionality works as expected.

### DIFF
--- a/apps/internal/mock/mock.go
+++ b/apps/internal/mock/mock.go
@@ -74,7 +74,6 @@ func (c *Client) AppendResponse(opts ...responseOption) {
 }
 
 func (c *Client) Do(req *http.Request) (*http.Response, error) {
-	println("In Mock Do")
 	if len(c.resp) == 0 {
 		panic(fmt.Sprintf(`no response for "%s"`, req.URL.String()))
 	}
--- a/apps/internal/mock/mock.go
+++ b/apps/internal/mock/mock.go
@@ -74,6 +74,7 @@ func (c *Client) AppendResponse(opts ...responseOption) {
 }
 
 func (c *Client) Do(req *http.Request) (*http.Response, error) {
+	println("In Mock Do")
 	if len(c.resp) == 0 {
 		panic(fmt.Sprintf(`no response for "%s"`, req.URL.String()))
 	}
--- a/apps/managedidentity/cloud_shell.go
+++ b/apps/managedidentity/cloud_shell.go
@@ -11,32 +11,23 @@ import (
 	"net/url"
 	"os"
 	"strings"
-
-	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base"
 )
-
-func acquireTokenForCloudShell(ctx context.Context, client Client, resource string) (base.AuthResult, error) {
-	req, err := createCloudShellAuthRequest(ctx, resource)
-	if err != nil {
-		return base.AuthResult{}, err
-	}
-	tokenResponse, err := client.getTokenForRequest(req)
-	if err != nil {
-		return base.AuthResult{}, err
-	}
-	return authResultFromToken(client.authParams, tokenResponse)
-}
 
 func createCloudShellAuthRequest(ctx context.Context, resource string) (*http.Request, error) {
 	msiEndpoint := os.Getenv(msiEndpointEnvVar)
 	msiEndpointParsed, err := url.Parse(msiEndpoint)
+
+	println("createCloudShellAuthRequest - msiEndpoint " + msiEndpoint)
+	println("createCloudShellAuthRequest - msiEndpointParsed " + msiEndpoint)
+	println("createCloudShellAuthRequest - resource " + resource)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't parse %q: %s", msiEndpoint, err)
 	}
 
-	msiParameters := msiEndpointParsed.Query()
-	msiParameters.Set(resourceQueryParameterName, resource)
-	msiDataEncoded := msiParameters.Encode()
+	data := url.Values{}
+	data.Set(resourceQueryParameterName, resource)
+	println("createCloudShellAuthRequest - dataGet " + data.Get(resourceQueryParameterName))
+	msiDataEncoded := data.Encode()
 	body := ioutil.NopCloser(strings.NewReader(msiDataEncoded))
 
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, msiEndpointParsed.String(), body)
--- a/apps/managedidentity/cloud_shell.go
+++ b/apps/managedidentity/cloud_shell.go
@@ -6,7 +6,7 @@ package managedidentity
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"net/url"
 	"os"
@@ -28,15 +28,18 @@ func createCloudShellAuthRequest(ctx context.Context, resource string) (*http.Re
 	data.Set(resourceQueryParameterName, resource)
 	println("createCloudShellAuthRequest - dataGet " + data.Get(resourceQueryParameterName))
 	msiDataEncoded := data.Encode()
-	body := ioutil.NopCloser(strings.NewReader(msiDataEncoded))
+	body := io.NopCloser(strings.NewReader(msiDataEncoded))
 
+	println("createCloudShellAuthRequest - did body stuff " + resource)
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, msiEndpointParsed.String(), body)
 	if err != nil {
+		println("createCloudShellAuthRequest - error for request " + resource)
 		return nil, fmt.Errorf("error creating http request %s", err)
 	}
 
 	req.Header.Set(metaHTTPHeaderName, "true")
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
+	println("createCloudShellAuthRequest - end of func " + resource)
 	return req, nil
 }
--- a/apps/managedidentity/cloud_shell.go
+++ b/apps/managedidentity/cloud_shell.go
@@ -16,30 +16,22 @@ import (
 func createCloudShellAuthRequest(ctx context.Context, resource string) (*http.Request, error) {
 	msiEndpoint := os.Getenv(msiEndpointEnvVar)
 	msiEndpointParsed, err := url.Parse(msiEndpoint)
-
-	println("createCloudShellAuthRequest - msiEndpoint " + msiEndpoint)
-	println("createCloudShellAuthRequest - msiEndpointParsed " + msiEndpoint)
-	println("createCloudShellAuthRequest - resource " + resource)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't parse %q: %s", msiEndpoint, err)
 	}
 
 	data := url.Values{}
 	data.Set(resourceQueryParameterName, resource)
-	println("createCloudShellAuthRequest - dataGet " + data.Get(resourceQueryParameterName))
 	msiDataEncoded := data.Encode()
 	body := io.NopCloser(strings.NewReader(msiDataEncoded))
 
-	println("createCloudShellAuthRequest - did body stuff " + resource)
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, msiEndpointParsed.String(), body)
 	if err != nil {
-		println("createCloudShellAuthRequest - error for request " + resource)
 		return nil, fmt.Errorf("error creating http request %s", err)
 	}
 
 	req.Header.Set(metaHTTPHeaderName, "true")
 	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
 
-	println("createCloudShellAuthRequest - end of func " + resource)
 	return req, nil
 }
--- a/apps/managedidentity/cloudshell_service.go
+++ b/apps/managedidentity/cloudshell_service.go
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation.
+// Licensed under the MIT license.
+
 package managedidentity
 
 import (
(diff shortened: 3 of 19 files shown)


apps/managedidentity/managedidentity_test.go
```diff
@@ -454,6 +454,122 @@ func TestIMDSAcquireTokenReturnsTokenSuccess(t *testing.T) {
...
+	}{
+		{resource: resource, miType: SystemAssigned()},
+		{resource: resourceDefaultSuffix, miType: SystemAssigned()},
+		{resource: resource, miType: UserAssignedClientID("clientId")},
+		{resource: resourceDefaultSuffix, miType: UserAssignedResourceID("resourceId")},
+		{resource: resourceDefaultSuffix, miType: UserAssignedObjectID("objectId")},
+	}
+	for _, testCase := range testCases {
+		t.Run(string(CloudShell)+"-"+testCase.miType.value(), func(t *testing.T) {
+			endpoint := os.Getenv(msiEndpointEnvVar)
```
chlowell, line 470: you need to set this, not get it. Otherwise, this test can only pass in Cloud Shell. At the moment this test actually panics when this variable isn't set

apps/managedidentity/managedidentity_test.go
```diff
@@ -454,6 +454,122 @@ func TestIMDSAcquireTokenReturnsTokenSuccess(t *testing.T) {
...
+func TestCloudShellAcquireTokenReturnsTokenSuccess(t *testing.T) {
+	testCases := []struct {
+		resource string
+		miType   ID
+	}{
+		{resource: resource, miType: SystemAssigned()},
+		{resource: resourceDefaultSuffix, miType: SystemAssigned()},
+		{resource: resource, miType: UserAssignedClientID("clientId")},
+		{resource: resourceDefaultSuffix, miType: UserAssignedResourceID("resourceId")},
+		{resource: resourceDefaultSuffix, miType: UserAssignedObjectID("objectId")},
```
chlowell, line 466: these are not success cases, and they all fail because `New` returns an error

apps/tests/devapps/managedidentity/managedidentity_sample.go
```diff
@@ -19,7 +19,6 @@ func runIMDSSystemAssigned() {
 		log.Fatal(err)
 	}
 	fmt.Println("token expire at : ", result.ExpiresOn)
-
```
chlowell, line 22: 👀

apps/managedidentity/cloudshell_service.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+func createCloudShellAuthRequest(ctx context.Context, resource string) (*http.Request, error) {
+	msiEndpoint := os.Getenv(msiEndpointEnvVar)
+	msiEndpointParsed, err := url.Parse(msiEndpoint)
+	if err != nil {
+		return nil, fmt.Errorf("couldn't parse %q: %s", msiEndpoint, err)
+	}
+
+	msiParameters := msiEndpointParsed.Query()
```
chlowell, line 34: this is weird because it looks like you're trying to add query parameters to the URL
```suggestion
	data := url.Values{}
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -454,6 +454,122 @@ func TestIMDSAcquireTokenReturnsTokenSuccess(t *testing.T) {
...
+				t.Fatal(err)
+			}
+
+			if localUrl == nil || !strings.HasPrefix(localUrl.String(), endpoint) {
+				t.Fatalf("url request is not on %s got %s", endpoint, localUrl)
+			}
+
+			query := localUrl.Query()
+
+			if query.Get(resourceQueryParameterName) != strings.TrimSuffix(testCase.resource, "/.default") {
```
chlowell, line 510: the resource should be in the body and not the query string

chlowell (changes requested):
Please be sure your tests pass before requesting reviews

apps/managedidentity/managedidentity.go
```diff
@@ -72,11 +74,15 @@ const (
...
 	imdsEndVar                          = "IMDS_ENDPOINT"
 	msiEndpointEnvVar                   = "MSI_ENDPOINT"
+	msiSecretEnvVar                     = "MSI_SECRET"
 	identityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
 
 	defaultRetryCount = 3
 )
 
+// UnsupportedSources is a list of sources that don't support user-assigned managed identities.
+var UnsupportedSources = []Source{ServiceFabric, AzureArc, AzureML, CloudShell}
```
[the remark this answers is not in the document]
AndyOHart: 1. Nah
2. Ah ok I saw in azure sdk go, logic for throwing and error for them but I understand what you mean, theres a difference with it. however in azure sdk there is still err thrown when these are detected, but with more detailed information, so i guess having them here isn't the worst if we have an error message explaining to go along with it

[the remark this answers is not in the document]
chlowell: A correct error for each platform is important but don't forget the different behavior--AML does support user-assigned identities specified at runtime. The easiest solution is copying the relevant azidentity code

apps/internal/mock/mock.go
```diff
@@ -74,6 +74,7 @@ func (c *Client) AppendResponse(opts ...responseOption) {
 }
 
 func (c *Client) Do(req *http.Request) (*http.Response, error) {
+	println("In Mock Do")
```
bgavrilMS, line 77: delete?

apps/managedidentity/cloud_shell.go
```diff
@@ -0,0 +1,45 @@
...
+		return nil, fmt.Errorf("couldn't parse %q: %s", msiEndpoint, err)
+	}
+
+	data := url.Values{}
+	data.Set(resourceQueryParameterName, resource)
+	println("createCloudShellAuthRequest - dataGet " + data.Get(resourceQueryParameterName))
+	msiDataEncoded := data.Encode()
+	body := io.NopCloser(strings.NewReader(msiDataEncoded))
+
+	println("createCloudShellAuthRequest - did body stuff " + resource)
```
bgavrilMS, line 33: ?

apps/managedidentity/cloud_shell.go
```diff
@@ -0,0 +1,45 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, msiEndpointParsed.String(), body)
+	if err != nil {
+		println("createCloudShellAuthRequest - error for request " + resource)
```
bgavrilMS, line 36: ?

apps/managedidentity/managedidentity.go
```diff
@@ -72,11 +74,20 @@ const (
...
+	msiSecretEnvVar                     = "MSI_SECRET"
 	identityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
 
 	defaultRetryCount = 3
 )
 
+// sourceRestrictions is a list of sources that have some form of restriction, such as user-assigned managed identities not being supported, or being supported through different means
+var sourceRestrictions = map[Source]string{
+	ServiceFabric: "Service Fabric API doesn't support specifying a user-assigned identity at runtime. The identity is determined by cluster resource configuration. See https://aka.ms/servicefabricmi",
+	AzureArc:      "Azure Arc doesn't support user-assigned managed identities at runtime",
```
bgavrilMS, line 86:
```suggestion
	AzureArc:      "Azure Arc doesn't support user-assigned managed identities",
```

apps/managedidentity/managedidentity.go
```diff
@@ -72,11 +74,20 @@ const (
...
 	msiEndpointEnvVar                   = "MSI_ENDPOINT"
+	msiSecretEnvVar                     = "MSI_SECRET"
 	identityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
 
 	defaultRetryCount = 3
 )
 
+// sourceRestrictions is a list of sources that have some form of restriction, such as user-assigned managed identities not being supported, or being supported through different means
+var sourceRestrictions = map[Source]string{
+	ServiceFabric: "Service Fabric API doesn't support specifying a user-assigned identity at runtime. The identity is determined by cluster resource configuration. See https://aka.ms/servicefabricmi",
```
bgavrilMS, line 85:
```suggestion
	ServiceFabric: "Service Fabric API doesn't support specifying a user-assigned identity. The identity is determined by cluster resource configuration. See https://aka.ms/servicefabricmi",
```

apps/managedidentity/managedidentity.go
```diff
@@ -72,11 +74,20 @@ const (
...
+	AzureArc:      "Azure Arc doesn't support user-assigned managed identities at runtime",
+	AzureML:       "Azure ML supports specifying a user-assigned managed identity by client ID only",
+	CloudShell:    "Cloud Shell doesn't support user-assigned managed identities at runtime",
```
bgavrilMS, line 88:
```suggestion
	CloudShell:    "Cloud Shell doesn't support user-assigned managed identities",
```

apps/managedidentity/managedidentity.go
```diff
@@ -72,11 +74,20 @@ const (
...
 	imdsEndVar                          = "IMDS_ENDPOINT"
 	msiEndpointEnvVar                   = "MSI_ENDPOINT"
+	msiSecretEnvVar                     = "MSI_SECRET"
 	identityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
 
 	defaultRetryCount = 3
 )
 
+// sourceRestrictions is a list of sources that have some form of restriction, such as user-assigned managed identities not being supported, or being supported through different means
+var sourceRestrictions = map[Source]string{
```
bgavrilMS, line 84: This assumes there is 1 restriction per source. Wouldn't it be simpler to use strings everywhere?

apps/managedidentity/cloud_shell.go
AndyOHart: Was for testing and for some reason visual studio code commited this file when it was not meant to

apps/managedidentity/cloud_shell.go
AndyOHart: Was for testing and for some reason visual studio code commited this file when it was not meant to

apps/managedidentity/managedidentity.go
```diff
@@ -199,11 +210,11 @@ func New(id ID, options ...ClientOption) (Client, error) {
...
 	}
 
-	// If source is Azure Arc return an error, as Azure Arc allow accepts System Assigned managed identities.
-	if source == AzureArc {
+	// If source has some restriction, return an error, as some sources allow System Assigned managed identities, but not user assigned, or user assigned through another method.
+	if errMsg, unsupported := sourceRestrictions[source]; unsupported {
 		switch id.(type) {
 		case UserAssignedClientID, UserAssignedResourceID, UserAssignedObjectID:
-			return Client{}, errors.New("azure Arc doesn't support user assigned managed identities")
+			return Client{}, errors.New(errMsg)
```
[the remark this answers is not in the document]
AndyOHart: I get your point, however it will be messier code if we have to do if elses or switches here, when we could just use this map. Since the errMsg corresponds to the mapping we have of the Source type, we should be easily able to see the issue came from here and which one threw the issue I would think

apps/managedidentity/managedidentity.go
AndyOHart: This code is just for replicating: https://github.com/Azure/azure-sdk-for-go/blob/82a8eb20c3a2fd20d5a972303d33311c129d81b0/sdk/azidentity/managed_identity_client.go#L150

But in a way that matches our design. Would renaming it to userManagedSourceRestrictions? Or we could leave as sourceRestrictions, and if we have more restrictions down the line line we can change it to be a list that holds all restrictions per source

bgavrilMS (approved):

apps/managedidentity/managedidentity.go
```diff
@@ -325,13 +351,25 @@ func (c Client) acquireTokenForIMDS(ctx context.Context, resource string) (base.
...
 	}
-	tokenResponse, err := c.getTokenForRequest(req)
+	tokenResponse, err := c.getTokenForRequest(req, resource)
 	if err != nil {
 		return base.AuthResult{}, err
 	}
 	return authResultFromToken(c.authParams, tokenResponse)
 }
 
+func (c Client) acquireTokenForCloudShell(ctx context.Context, client Client, resource string) (base.AuthResult, error) {
```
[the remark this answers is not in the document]
AndyOHart: Ah forgot to clean it up after merging Nileshs pr, thanks

4gust (approved):
LGTM

apps/managedidentity/managedidentity_test.go
```diff
@@ -465,6 +465,105 @@ func TestIMDSAcquireTokenReturnsTokenSuccess(t *testing.T) {
...
+}
+
+func TestCloudShellOnlySystemAssignedSupported(t *testing.T) {
+	setEnvVars(t, CloudShell)
+	mockClient := mock.Client{}
+
+	for _, testCase := range []ID{
+		UserAssignedClientID("client"),
+		UserAssignedObjectID("ObjectId"),
+		UserAssignedResourceID("resourceid")} {
```
chlowell, line 554: this is more conventional, though gofmt doesn't require it
```suggestion
		UserAssignedResourceID("resourceid"),
	} {
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -465,6 +465,105 @@ func TestIMDSAcquireTokenReturnsTokenSuccess(t *testing.T) {
 	}
 }
 
+func TestCloudShellAcquireTokenReturnsTokenSuccess(t *testing.T) {
+	resource := "https://resource"
```
chlowell, line 469: if you're testing only one format, it's best to use the interesting one
```suggestion
	resource := "https://resource/.default"
```

chlowell (approved):

PR #545, opened 2025-01-24
Fix Bug: Prevent Empty Region in WithAzureRegion from Overriding MSAL_FORCE_REGION

4gust:
### **Description:**

This PR addresses a bug where passing an empty region (`""`) to the `WithAzureRegion` function would unintentionally override the value of the `MSAL_FORCE_REGION` environment variable, even when `MSAL_FORCE_REGION` was explicitly set. The correct behavior is that if `MSAL_FORCE_REGION` is defined, it should always take precedence over any empty region passed through `WithAzureRegion`.

### **Bug Details:**

- **Issue:** If the `MSAL_FORCE_REGION` environment variable is set, and an empty region (`""`) is passed to `WithAzureRegion`, the empty value would override the region specified by `MSAL_FORCE_REGION`. This is not the expected behavior.
- **Expected Behavior:** The value of `MSAL_FORCE_REGION` should not be overridden by an empty string passed via `WithAzureRegion`. If `MSAL_FORCE_REGION` is set, it should always take precedence.
- **Cause:** The current logic did not properly check if `WithAzureRegion` not being checked if it was assigning empty string or not. This resulted in the empty string incorrectly taking precedence over the environment variable.

### **Solution:**

The fix ensures that if `MSAL_FORCE_REGION` is set in the environment, it will always be used, and the region passed through `WithAzureRegion` will only be considered if it is not empty. This prevents accidental overrides when `MSAL_FORCE_REGION` is explicitly set.

### **Changes:**

- **Logic Update:** Updated the logic in `WithAzureRegion` to first check if value passed via `WithAzureRegion` is empty or not.
- **Unit Test:** Added tests to cover scenarios where both `MSAL_FORCE_REGION` and an empty region are provided, ensuring that the environment variable always takes precedence.

### DIFF
--- a/apps/confidential/confidential_test.go
+++ b/apps/confidential/confidential_test.go
@@ -235,13 +235,8 @@ func TestRegionAutoEnable_SpecifiedEmptyRegion_EnvRegion(t *testing.T) {
 				t.Fatal(err)
 			}
 			if test.envRegion != "" {
-				err = os.Setenv("MSAL_FORCE_REGION", test.envRegion)
-				if err != nil {
-					t.Fatal(err)
-				}
-				defer os.Unsetenv("MSAL_FORCE_REGION")
+				t.Setenv("MSAL_FORCE_REGION", test.envRegion)
 			}
-
 			lmo := "login.microsoftonline.com"
 			tenant := "tenant"
 			mockClient := mock.Client{}
@@ -256,7 +251,6 @@ func TestRegionAutoEnable_SpecifiedEmptyRegion_EnvRegion(t *testing.T) {
 				if client.base.AuthParams.AuthorityInfo.Region != "" {
 					t.Fatalf("wanted %q, got %q", test.resultRegion, client.base.AuthParams.AuthorityInfo.Region)
 				}
-				return
 			} else {
 				if client.base.AuthParams.AuthorityInfo.Region != test.resultRegion {
 					t.Fatalf("wanted %q, got %q", test.resultRegion, client.base.AuthParams.AuthorityInfo.Region)
--- a/apps/confidential/confidential_test.go
+++ b/apps/confidential/confidential_test.go
@@ -247,17 +247,10 @@ func TestRegionAutoEnable_SpecifiedEmptyRegion_EnvRegion(t *testing.T) {
 			mockClient := mock.Client{}
 
 			var client Client
-			// if test.region != "" {
 			client, err = New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient), WithAzureRegion(test.region))
 			if err != nil {
 				t.Fatal(err)
 			}
-			// } else {
-			// 	client, err = New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient))
-			// 	if err != nil {
-			// 		t.Fatal(err)
-			// 	}
-			// }
 
 			if test.resultRegion == "DisableMsalForceRegion" {
 				if client.base.AuthParams.AuthorityInfo.Region != "" {
--- a/apps/confidential/confidential_test.go
+++ b/apps/confidential/confidential_test.go
@@ -241,8 +241,7 @@ func TestRegionAutoEnable_SpecifiedEmptyRegion_EnvRegion(t *testing.T) {
 			tenant := "tenant"
 			mockClient := mock.Client{}
 
-			var client Client
-			client, err = New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient), WithAzureRegion(test.region))
+			client, err := New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient), WithAzureRegion(test.region))
 			if err != nil {
 				t.Fatal(err)
 			}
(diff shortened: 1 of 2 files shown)


AndyOHart (approved):
Goods good! Thanks

apps/confidential/confidential_test.go
```diff
@@ -195,50 +195,81 @@ func TestRegionAutoEnable_EmptyRegion_EnvRegion(t *testing.T) {
...
-	}
-}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			cred, err := NewCredFromSecret(fakeSecret)
+			if err != nil {
+				t.Fatal(err)
+			}
+			if test.envRegion != "" {
+				err = os.Setenv("MSAL_FORCE_REGION", test.envRegion)
```
chlowell, line 238: [t.Setenv](https://pkg.go.dev/testing#T.Setenv)

apps/confidential/confidential_test.go
```diff
@@ -195,50 +195,81 @@ func TestRegionAutoEnable_EmptyRegion_EnvRegion(t *testing.T) {
...
+			}
 
-	if client.base.AuthParams.AuthorityInfo.Region != "" {
-		t.Fatalf("wanted empty, got %q", client.base.AuthParams.AuthorityInfo.Region)
+			if test.resultRegion == "DisableMsalForceRegion" {
+				if client.base.AuthParams.AuthorityInfo.Region != "" {
+					t.Fatalf("wanted %q, got %q", test.resultRegion, client.base.AuthParams.AuthorityInfo.Region)
+				}
+				return
+			} else {
```
chlowell, line 267: you don't need this, because you exit the test when the `if` evaluates true

apps/confidential/confidential_test.go
```diff
@@ -195,50 +195,81 @@ func TestRegionAutoEnable_EmptyRegion_EnvRegion(t *testing.T) {
...
-	lmo := "login.microsoftonline.com"
-	tenant := "tenant"
-	mockClient := mock.Client{}
-	testRegion := "DisableMsalForceRegion"
-	client, err := New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient), WithAzureRegion(testRegion))
-	if err != nil {
-		t.Fatal(err)
-	}
+			var client Client
+			if test.region != "" {
```
chlowell, line 250: just a suggestion for tidying this up
```go
opts := []Option{WithHTTPClient(&mockClient)}
if test.region != "" {
	opts = append(opts, WithAzureRegion(test.region))
}
client, err := New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, opts...)
```

Having written that though, I realize this test doesn't verify the fix. The bug is:
1. MSAL_FORCE_REGION set
2. application calls `New(..., WithAzureRegion(""))`
3. Client uses non-regional authority i.e., ignores the value of MSAL_FORCE_REGION

This test doesn't do step 2 (try reverting the fix; does this test still pass?)

chlowell (changes requested):
Needs test coverage

apps/confidential/confidential_test.go
```diff
@@ -195,50 +195,81 @@ func TestRegionAutoEnable_EmptyRegion_EnvRegion(t *testing.T) {
...
-	if err != nil {
-		t.Fatal(err)
-	}
+			var client Client
+			// if test.region != "" {
+			client, err = New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient), WithAzureRegion(test.region))
+			if err != nil {
+				t.Fatal(err)
+			}
+			// } else {
```
bgavrilMS, line 255: Dead code?

4gust: Removed 👍🏽

apps/confidential/confidential_test.go
4gust: Updated the test to do that.

apps/confidential/confidential_test.go
```diff
@@ -195,50 +195,68 @@ func TestRegionAutoEnable_EmptyRegion_EnvRegion(t *testing.T) {
...
-	lmo := "login.microsoftonline.com"
-	tenant := "tenant"
-	mockClient := mock.Client{}
-	testRegion := "DisableMsalForceRegion"
-	client, err := New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient), WithAzureRegion(testRegion))
-	if err != nil {
-		t.Fatal(err)
-	}
+			var client Client
+			client, err = New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient), WithAzureRegion(test.region))
```
chlowell, line 245:
```suggestion
			client, err := New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient), WithAzureRegion(test.region))
```

apps/confidential/confidential_test.go
```diff
@@ -195,50 +195,68 @@ func TestRegionAutoEnable_EmptyRegion_EnvRegion(t *testing.T) {
...
+				t.Fatal(err)
+			}
 
-	if client.base.AuthParams.AuthorityInfo.Region != "" {
-		t.Fatalf("wanted empty, got %q", client.base.AuthParams.AuthorityInfo.Region)
+			if test.resultRegion == "DisableMsalForceRegion" {
+				if client.base.AuthParams.AuthorityInfo.Region != "" {
+					t.Fatalf("wanted %q, got %q", test.resultRegion, client.base.AuthParams.AuthorityInfo.Region)
+				}
+			} else {
```
chlowell, line 254: `else if`

apps/confidential/confidential_test.go
```diff
@@ -195,50 +195,68 @@ func TestRegionAutoEnable_EmptyRegion_EnvRegion(t *testing.T) {
...
+			var client Client
+			client, err = New(fmt.Sprintf(authorityFmt, lmo, tenant), fakeClientID, cred, WithHTTPClient(&mockClient), WithAzureRegion(test.region))
+			if err != nil {
+				t.Fatal(err)
+			}
 
-	if client.base.AuthParams.AuthorityInfo.Region != "" {
-		t.Fatalf("wanted empty, got %q", client.base.AuthParams.AuthorityInfo.Region)
+			if test.resultRegion == "DisableMsalForceRegion" {
+				if client.base.AuthParams.AuthorityInfo.Region != "" {
```
chlowell, line 251: I suppose the implementation is probably stable enough to depend on but note this is testing implementation details of the feature, not the feature itself, which is a mapping from the public API to HTTP requests. If I were testing this, I'd instead verify the actual authority URL of those requests

chlowell (approved):

apps/confidential/confidential_test.go
4gust: Updated

apps/confidential/confidential_test.go
4gust: Updated

PR #546, opened 2025-01-28
Azure ML Support

AndyOHart:
Feature - Add support for AzureML (Machine Learning)

Description:
This PR adds the ability for AzureML to be used as a Managed Identity source.

Changes:
* Update logic for GetSource so we can detect when AzureML should be used
* Add AuthRequest of making an AzureML request. Similar to AppService, but with the addition to having the msiSecretEnvVar in the request header
* Updates GetSource tests to allow AzureML and AppService to be detected (was missing before)
* Adds general tests for AzureML for getting a successful response and also returning error when resource and object id are used

### DIFF
--- a/apps/managedidentity/azure_ml.go
+++ b/apps/managedidentity/azure_ml.go
@@ -5,7 +5,6 @@ package managedidentity
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 	"os"
 )
@@ -20,13 +19,9 @@ func createAzureMLAuthRequest(ctx context.Context, id ID, resource string) (*htt
 	q := req.URL.Query()
 	q.Set(apiVersionQueryParameterName, azureMLAPIVersion)
 	q.Set(resourceQueryParameterName, resource)
-
-	switch t := id.(type) {
-	case UserAssignedClientID:
-		q.Set("clientid", string(t))
-	case systemAssignedValue:
-	default:
-		return nil, fmt.Errorf("unsupported type %T", id)
+	q.Set("clientid", os.Getenv("DEFAULT_IDENTITY_CLIENT_ID"))
+	if cid, ok := id.(UserAssignedClientID); ok {
+		q.Set("clientid", string(cid))
 	}
 	req.URL.RawQuery = q.Encode()
 	return req, nil
--- a/apps/managedidentity/azure_ml.go
+++ b/apps/managedidentity/azure_ml.go
@@ -18,16 +18,12 @@ func createAzureMLAuthRequest(ctx context.Context, id ID, resource string) (*htt
 
 	req.Header.Set("secret", os.Getenv(msiSecretEnvVar))
 	q := req.URL.Query()
-	q.Set(apiVersionQueryParameterName, azureMlApiVersion)
+	q.Set(apiVersionQueryParameterName, azureMLAPIVersion)
 	q.Set(resourceQueryParameterName, resource)
 
 	switch t := id.(type) {
 	case UserAssignedClientID:
-		q.Set(miQueryParameterClientId, string(t))
-	case UserAssignedResourceID:
-		return nil, fmt.Errorf("unsupported type %T", id)
-	case UserAssignedObjectID:
-		return nil, fmt.Errorf("unsupported type %T", id)
+		q.Set("clientid", string(t))
 	case systemAssignedValue:
 	default:
 		return nil, fmt.Errorf("unsupported type %T", id)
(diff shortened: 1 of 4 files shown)


bgavrilMS (approved):

4gust (approved):

apps/managedidentity/azure_ml.go
```diff
@@ -0,0 +1,33 @@
...
+	q.Set(apiVersionQueryParameterName, azureMLAPIVersion)
+	q.Set(resourceQueryParameterName, resource)
+
+	switch t := id.(type) {
+	case UserAssignedClientID:
+		q.Set("clientid", string(t))
+	case systemAssignedValue:
+	default:
+		return nil, fmt.Errorf("unsupported type %T", id)
+	}
```
chlowell, line 30: Sorry I forgot to point this out before now, but `clientid` must always be set, and the platform provides a default value. So instead of the switch, you need something like this:
```go
q.Set("clientid", os.Getenv("DEFAULT_IDENTITY_CLIENT_ID"))
if cid, ok := id.(UserAssignedClientID); ok {
	q.Set("clientid", string(cid))
}
```

apps/managedidentity/managedidentity_test.go
```diff
@@ -656,6 +659,112 @@ func TestAppServiceAcquireTokenReturnsTokenSuccess(t *testing.T) {
...
+			}
+			if testCase.miType.value() == "clientId" {
+				if query.Get("clientid") != "clientId" {
+					t.Fatalf("expected clientid to be set to clientId, got %s", query.Get("clientid"))
+				}
+			} else {
+				if query.Get("clientid") != os.Getenv("DEFAULT_IDENTITY_CLIENT_ID") {
+					t.Fatalf("expected clientid to be set to default identity client id, got %s", query.Get("clientid"))
+				}
+			}
```
chlowell, line 726: There are 2 cases to test here:
- DEFAULT_IDENTITY_CLIENT_ID = A & SystemAssigned() -> clientid=A
- DEFAULT_IDENTITY_CLIENT_ID = A & UserAssignedClientID("B") -> clientid=B

This doesn't cover them because you don't set the variable, and both `Get` and `Getenv` return `""` when the given key has no value

apps/managedidentity/managedidentity_test.go
```diff
@@ -661,14 +661,20 @@ func TestAppServiceAcquireTokenReturnsTokenSuccess(t *testing.T) {
 }
 
 func TestAzureMLAcquireTokenReturnsTokenSuccess(t *testing.T) {
+	// Set the DEFAULT_IDENTITY_CLIENT_ID environment variable
+	defaultClientID := "A"
+	os.Setenv("DEFAULT_IDENTITY_CLIENT_ID", defaultClientID)
+	defer os.Unsetenv("DEFAULT_IDENTITY_CLIENT_ID")
```
chlowell, line 667: This has a side effect because `os.*env()` operate at process level: after this test, the variable is unset for all subsequent tests. You should instead restore the value set before the test, if any. [t.Setenv](https://pkg.go.dev/testing#T.Setenv) will do that for you

apps/managedidentity/managedidentity_test.go
```diff
@@ -661,14 +661,20 @@ func TestAppServiceAcquireTokenReturnsTokenSuccess(t *testing.T) {
 }
 
 func TestAzureMLAcquireTokenReturnsTokenSuccess(t *testing.T) {
+	// Set the DEFAULT_IDENTITY_CLIENT_ID environment variable
```
chlowell, line 664: please forgive my nitpicking, but this code doesn't warrant a comment. You can assume readers understand `Setenv()`

apps/managedidentity/managedidentity_test.go
```diff
@@ -715,15 +721,10 @@ func TestAzureMLAcquireTokenReturnsTokenSuccess(t *testing.T) {
...
-			if testCase.miType.value() == "clientId" {
-				if query.Get("clientid") != "clientId" {
-					t.Fatalf("expected clientid to be set to clientId, got %s", query.Get("clientid"))
-				}
-			} else {
-				if query.Get("clientid") != os.Getenv("DEFAULT_IDENTITY_CLIENT_ID") {
-					t.Fatalf("expected clientid to be set to default identity client id, got %s", query.Get("clientid"))
-				}
+			if query.Get("clientid") != testCase.expectedClientID {
+				t.Fatalf("expected clientid to be set to %s, got %s", testCase.expectedClientID, query.Get("clientid"))
```
chlowell, line 725: just a suggestion for tidying and reducing the chance of typo bugs
```suggestion
			if actual := query.Get("clientid"); actual != testCase.expectedClientID {
				t.Fatalf("expected clientid to be set to %s, got %s", testCase.expectedClientID, actual)
```

chlowell (approved):

PR #551, opened 2025-02-12
Remove redundant ClientOptions type

chlowell:
ClientOptions is a subset of Client so instead of setting ClientOptions fields and having New copy their values onto Client, options funcs can simply set Client fields directly


4gust (approved):

AndyOHart (approved):
lgtm!

PR #557, opened 2025-03-04
Added a FIC test

4gust:
This PR introduces an end-to-end (E2E) test for Federated Identity Credentials. The test ensures the full flow of federated identity credential issuance, authentication, and verification works seamlessly across the system. By adding this test, we improve test coverage, enhance the reliability of federated identity handling, and help identify any issues early in the development cycle, ensuring smoother future updates and better system stability.

### DIFF
--- a/apps/tests/integration/integration_test.go
+++ b/apps/tests/integration/integration_test.go
@@ -21,6 +21,8 @@ import (
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/errors"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/managedidentity"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
 )
 
@@ -53,7 +55,9 @@ func httpRequest(ctx context.Context, url string, query url.Values, accessToken 
 		return nil, fmt.Errorf("failed to build new http request: %w", err)
 	}
 	req.Header.Set("Authorization", "Bearer "+accessToken)
-	req.URL.RawQuery = query.Encode()
+	if querry := query.Encode(); querry != "" {
+		req.URL.RawQuery = querry
+	}
 
 	resp, err := httpClient.Do(req)
 	if err != nil {
@@ -139,6 +143,18 @@ func (l *labClient) labAccessToken() (string, error) {
 		}
 	}
 	return result.AccessToken, nil
+}
+
+func (l *labClient) getLabResponse(url string) (string, error) {
+	accessToken, err := l.labAccessToken()
+	if err != nil {
+		return "", fmt.Errorf("problem getting lab access token: %w", err)
+	}
+	responseBody, err := httpRequest(context.Background(), url, nil, accessToken)
+	if err != nil {
+		return "", err
+	}
+	return string(responseBody), nil
 }
 
 func (l *labClient) user(ctx context.Context, query url.Values) (user, error) {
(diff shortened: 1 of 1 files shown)


bgavrilMS (approved):

apps/tests/integration/integration_test.go
```diff
@@ -480,3 +496,112 @@ func TestAccountFromCache(t *testing.T) {
...
+		t.Fatalf("Failed to acquire token: %v", err)
+	}
+	if token.AccessToken == "" {
+		t.Fatal("Expected non-empty access token")
+	}
+	cred := confidential.NewCredFromAssertionCallback(func(ctx context.Context, opt confidential.AssertionRequestOptions) (string, error) {
+		token, err := msiClient.AcquireToken(ctx, resource)
+		if err != nil {
+			t.Fatalf("Failed to acquire token: %v", err)
+			return "", err
```
chlowell, line 575: this return is redundant because `t.Fatalf` halts execution and you already have an unconditional return to satisfy the compiler

apps/tests/integration/integration_test.go
```diff
@@ -480,3 +496,112 @@ func TestAccountFromCache(t *testing.T) {
...
+	}
+
+	authResult, err := confidentialClient.AcquireTokenByCredential(ctx, []string{"https://msidlabs.vault.azure.net/.default"})
+	if err != nil {
+		t.Fatalf("Failed to acquire token by credential: %v", err)
+	}
+	if authResult.AccessToken == "" {
+		t.Fatal("Expected non-empty access token")
+	}
+	if authResult.Metadata.TokenSource != base.IdentityProvider {
```
chlowell, line 594: not related to this PR: should `base.TokenSource` be available to developers?

apps/tests/integration/integration_test.go
```diff
@@ -53,7 +55,9 @@ func httpRequest(ctx context.Context, url string, query url.Values, accessToken
 		return nil, fmt.Errorf("failed to build new http request: %w", err)
 	}
 	req.Header.Set("Authorization", "Bearer "+accessToken)
-	req.URL.RawQuery = query.Encode()
+	if querry := query.Encode(); querry != "" {
```
chlowell, line 58: `url.Values{}.Encode()` is okay, it just returns an empty string, which is the zero value of `url.RawQuery` anyway

apps/tests/integration/integration_test.go
4gust: Yes, we added this a couple of versions before.
Its in `authResult.Metadata.TokenSource`

apps/tests/integration/integration_test.go
```diff
@@ -480,3 +496,112 @@ func TestAccountFromCache(t *testing.T) {
...
 }
+
+type urlModifierTransport struct {
+	base       http.RoundTripper
+	modifyFunc func(*http.Request)
+}
+
+// RoundTrip implements the http.RoundTripper interface
+func (t *urlModifierTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	// Create a copy of the request to avoid modifying the original
```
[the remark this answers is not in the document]
4gust: Updated

apps/tests/integration/integration_test.go
```diff
@@ -480,3 +496,112 @@ func TestAccountFromCache(t *testing.T) {
...
+		t.Fatalf("Failed to get resource env variable: %v", err)
+	}
+	var result map[string]string
+	err = json.Unmarshal([]byte(response), &result)
+	if err != nil {
+		t.Fatalf("Failed to unmarshal response: %v", err)
+	}
+
+	for key, value := range result {
+		test := "https://service.msidlab.com/MSIToken?azureresource=WebApp&uri=http://127.0.0.1:414882/msi/token/"
```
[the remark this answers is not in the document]
4gust: Its in the URL that was provided by the LabAuth,
Now using it from the value it self.

apps/tests/integration/integration_test.go
4gust: Updated.

apps/tests/integration/integration_test.go
```diff
@@ -480,3 +496,112 @@ func TestAccountFromCache(t *testing.T) {
...
+	}
+	// Replace your existing http.Client with this one
+	httpClient := http.Client{
+		Transport: &urlModifierTransport{
+			base: http.DefaultTransport,
+			modifyFunc: func(req *http.Request) {
+				req.URL.Host = "service.msidlab.com"
+				req.URL.Path = "/MSIToken"
+				req.URL.Scheme = "https"
+				req.URL.RawQuery = "azureresource=WebApp&uri=http%3A%2F%2F127.0.0.1%3A41488%2Fmsi%2Ftoken%2F%3Fapi-version%3D2019-08-01%26resource%3Dapi%3A%2F%2Fazureadtokenexchange%26client_id%3D4b7a4b0b-ecb2-409e-879a-1e21a15ddaf6"
```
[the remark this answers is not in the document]
4gust: I created a variable, if now then how will we create a MSI client with client id

apps/tests/integration/integration_test.go
chlowell: That value isn't useful by itself because the related constants aren't in the public API. So instead of the line of code you have here, an application developer has to write this:
```go
if authResult.Metadata.TokenSource != 1 {
```

apps/tests/integration/integration_test.go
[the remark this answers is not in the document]
chlowell: it would be best to get this value from whatever automation runs the test, say via an environment variable, so you don't need a PR when the client ID changes

apps/tests/integration/integration_test.go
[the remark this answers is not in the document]
chlowell: it's still hard-coded in the query string

apps/tests/integration/integration_test.go
```diff
@@ -480,3 +492,109 @@ func TestAccountFromCache(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("Failed to unmarshal response: %v", err)
+	}
+
+	for key, value := range result {
+		if key == "IDENTITY_ENDPOINT" {
+			value = "https://service.msidlab.com/MSIToken?azureresource=WebApp&uri=" + value
+		}
+		os.Setenv(key, value)
+		defer os.Unsetenv(key)
```
chlowell, line 533: `t.Setenv` is better because it restores the pre-test value i.e., it prevents side-effects on other tests

apps/managedidentity/managedidentity_test.go
```diff
@@ -1131,3 +1132,95 @@ func TestCreatingIMDSClient(t *testing.T) {
 		})
 	}
 }
+
+func BenchmarkAcquireTokenSilentLatency(b *testing.B) {
```
chlowell, line 1136: this isn't related to FIC; do you want to add it in a separate PR?

apps/managedidentity/managedidentity_test.go
```diff
@@ -1131,3 +1132,95 @@ func TestCreatingIMDSClient(t *testing.T) {
...
+
+	sort.Slice(latencies, func(i, j int) bool { return latencies[i] < latencies[j] })
+
+	p50 := latencies[len(latencies)*50/100]
+	p90 := latencies[len(latencies)*90/100]
+	p99 := latencies[len(latencies)*99/100]
+
+	b.ReportMetric(float64(p50.Microseconds()), "p50_us")
+	b.ReportMetric(float64(p90.Microseconds()), "p90_us")
+	b.ReportMetric(float64(p99.Microseconds()), "p99_us")
```
chlowell, line 1181: Are these values interesting? The benchmark covers getting a token from the cache i.e., it measures how long it took to acquire an uncontested lock or two and find a value in a map that never grows. I imagine the distribution is therefore quite flat

4gust: I will create a separate PR for performance test.

apps/tests/integration/integration_test.go
4gust: Updated

apps/tests/integration/integration_test.go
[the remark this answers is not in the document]
4gust: these values are also created as constants in MSAL.net and others as well

apps/tests/integration/integration_test.go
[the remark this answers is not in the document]
4gust: Is there any other way we can do that for proxying one request from one source to another?
when I tried 1:1 replacement the `http.Client` return error as the source done match the response.

PR #561, opened 2025-03-13
Exposing necessary types from each client

4gust:
**PR Title:** Expose `TokenSource` and Related Values Across Packages in `AuthResult`

**PR Description:**

This PR exposes the `TokenSource` type and its associated values (`TokenSourceUnknown`, `TokenSourceCache`, `TokenSourceIdentityProvider`) from the `base` package to the `confidential`, `managedidentity`, and `public` packages.

### Changes:
- `TokenSource`, `TokenSourceUnknown`, `TokenSourceCache`, and `TokenSourceIdentityProvider` are now accessible from the `confidential`, `managedidentity`, and `public` packages.

### DIFF
--- a/apps/confidential/confidential.go
+++ b/apps/confidential/confidential.go
@@ -65,6 +65,12 @@ type AuthenticationScheme = authority.AuthenticationScheme
 
 type Account = shared.Account
 
+type TokenSource = base.TokenSource
+
+const TokenSourceUnknown = base.TokenSourceUnknown
+const TokenSourceIdentityProvider = base.TokenSourceIdentityProvider
+const TokenSourceCache = base.TokenSourceCache
+
 // CertFromPEM converts a PEM file (.pem or .key) for use with [NewCredFromCert]. The file
 // must contain the public certificate and the private key. If a PEM block is encrypted and
 // password is not an empty string, it attempts to decrypt the PEM blocks using the password.
--- a/apps/confidential/confidential_test.go
+++ b/apps/confidential/confidential_test.go
@@ -1097,7 +1097,7 @@ func TestRefreshIn(t *testing.T) {
 			if err != nil {
 				t.Fatal(err)
 			}
-			if ar.Metadata.TokenSource != base.Cache && !tt.shouldGetNewToken {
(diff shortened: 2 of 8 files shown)


apps/confidential/confidential.go
```diff
@@ -65,6 +65,12 @@ type AuthenticationScheme = authority.AuthenticationScheme
 
 type Account = shared.Account
 
+type TokenSource = base.TokenSource
+
+const TokenSourceUnknown = base.TokenSourceUnknown
```
bgavrilMS, line 70: I know we need some default value, but can we instead default to "IdentityProvider".

bgavrilMS (approved):

apps/confidential/confidential.go
4gust: The default value for `int` in go is 0.
If we have a empty object it will still have a default `0` that might indicate  `TokenSourceIdentityProvider `which we want to avoid.

chlowell: That's what Bogdan is suggesting, if I understand correctly. Seems reasonable to me because the original source of all tokens is the identity provider; TokenSourceIdentityProvider can never be incorrect, strictly speaking

4gust: I agree, we had a discussion about this in the original PR 
https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/498#discussion_r1725666156
Here is the link to the discussion 

To prevent this we have this default value
```go
ar := AuthResult{}
ar.Metadata.TokenSource == IdentityProvider // true
```

chlowell: With that comment I was trying to underline the importance of the constant assigned 0: if someone forgets to assign a struct field of this type, it has that value. I didn't mean to imply a faux enum *must* have a special constant for 0. (I do think it's crucial to assign 0 to a constant so that the type's zero value falls within the enum though.)

The question in this thread is different: is there any reason to make TokenSourceUnknown part of the public API? I'm in favor of dropping it because MSAL doesn't use it today and you can add it later if needed.

chlowell: Oops, let me try that question again: is there any reason for TokenSourceUnknown to exist?

chlowell (approved):

PR #565, opened 2025-03-28
Added example for ManagedIdentity for go package.

4gust:
**PR Description:**

This PR introduces the `example_test.go` file for the `managedidentity` package. The file is designed to demonstrate examples under the Go documentation for the package functions. Key points of this PR include:

- **Provides Example**: Includes example tests that show how to use functions within the `managedidentity` package.

These examples will improve both the documentation and testing for the `managedidentity` package.

### DIFF
--- a/apps/managedidentity/example_test.go
+++ b/apps/managedidentity/example_test.go
@@ -1,0 +1,38 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package managedidentity_test
+
+import (
+	"context"
+	"fmt"
+
+	mi "github.com/AzureAD/microsoft-authentication-library-for-go/apps/managedidentity"
+)
+
+func ExampleNew() {
+	miClient, err := mi.New(mi.SystemAssigned())
+	if err != nil {
+		fmt.Println("Error while making client", err)
+	}
+	token, err := miClient.AcquireToken(context.Background(), "resouce")
+	if err != nil {
+		fmt.Println("Error while fetching token", err)
+	}
+	fmt.Println("Token expires at:", token.ExpiresOn)
+	// Output:
+}
+
+func ExampleClient_AcquireToken() {
+	miClient, err := mi.New(mi.SystemAssigned())
+	if err != nil {
+		fmt.Println("Error while making client", err)
+	}
+	token, err := miClient.AcquireToken(context.Background(), "resouce")
+	if err != nil {
+		fmt.Println("Error while fetching token", err)
+	}
+	fmt.Println("Token expires at:", token.ExpiresOn)
+	// Output:
+
+}
--- a/apps/managedidentity/example_test.go
+++ b/apps/managedidentity/example_test.go
@@ -22,15 +22,13 @@ func ExampleNew() {
 	_ = miSystemassignedClient
 
 	// User assigned Client
-	clientId := "ClientId" // TODO: replace the correct Client id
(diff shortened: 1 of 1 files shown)


apps/managedidentity/example_test.go
```diff
@@ -0,0 +1,59 @@
...
+
+func ExampleGetSource() {
+	source, err := mi.GetSource()
+	if err != nil {
+		// TODO: Handle error
+	}
+	print("source is :", source)
+
+	// Output: source is : <source>
+}
```
chlowell, line 59: I'd delete these because they don't improve the API documentation--they just demonstrate how to call ordinary functions

apps/managedidentity/example_test.go
```diff
@@ -0,0 +1,59 @@
...
+	// User assigned Client
+	clientId := "ClientId" // TODO: replace the correct Client id
+
+	miClientIdAssignedClient, err := mi.New(mi.UserAssignedClientID(clientId))
+	if err != nil {
+		// TODO: Handle error
+	}
+	_ = miClientIdAssignedClient
+
+	// Output: -
```
chlowell, line 33: Suggest removing this because it doesn't improve the example--there's no output to help readers understand how the API works (to be clear, this example doesn't need any)--or test coverage (note that `go test` always compiles examples, so `// Output` isn't required; it just makes `go test` execute an example)

apps/managedidentity/example_test.go
```diff
@@ -0,0 +1,59 @@
...
+	// Output: Token expires at: <datetime>
+
+}
+
+func ExampleGetSource() {
+	source, err := mi.GetSource()
+	if err != nil {
+		// TODO: Handle error
+	}
+	print("source is :", source)
```
chlowell, line 56: The conventional `fmt.Print*` comes with a compatibility guarantee whereas the builtin `print` [does not](https://pkg.go.dev/builtin#print)

chlowell (dismissed):

apps/managedidentity/example_test.go
```diff
@@ -0,0 +1,59 @@
...
+func ExampleNew() {
+	// System assigned Client
+	miSystemassignedClient, err = mi.New(mi.SystemAssigned())
+	if err != nil {
+		// TODO: Handle error
+	}
+	_ = miSystemassignedClient
+
+	// User assigned Client
+	clientId := "ClientId" // TODO: replace the correct Client id
```
bgavrilMS, line 25:
```suggestion
	clientId := "ClientId" // TODO: replace with your Managed Identity Id
```

apps/managedidentity/example_test.go
```diff
@@ -0,0 +1,59 @@
...
+var miSystemassignedClient mi.Client
+var err error
+
+func ExampleNew() {
+	// System assigned Client
+	miSystemassignedClient, err = mi.New(mi.SystemAssigned())
+	if err != nil {
+		// TODO: Handle error
+	}
+	_ = miSystemassignedClient
```
bgavrilMS, line 22: Is this needed to stop some static analyzer from spewing a warning ? Can we disable the analyzer instead for the entire file?

4gust: Yea, or else the compiler wont compile as the variable is not used.

apps/managedidentity/example_test.go
4gust: Updated

apps/managedidentity/example_test.go
4gust: We also remove the `AcquireToken` ?

PR #567, opened 2025-04-03
Added support for Token revocation support

4gust:
**PR Title:** Add Token Revocation Handling and Enhanced Testing for App Service in Managed Identity

**Description:**  
This PR introduces a new feature to handle token revocation scenarios in App Service managed identity and enhances testing to ensure robust behavior. The changes focus on handling claims challenges, bad access tokens from the cache, and validating client capabilities.

---

#### **Feature Implementation:**

1. **Token Revocation Handling:**
   - Ensures that the client retries the request with the `claims` parameter included in the URL to obtain a fresh token.

2. **Bad Access Token Handling:**
   - Introduced logic to detect bad access tokens retrieved from the cache when claims are passed and ensures that a fresh token is acquired when necessary.

---

#### **New Unit Tests:**

1. **TestAppServiceWithClaimsAndBadAccessToken**
   - Validates the scenario when claims are passed, and a bad access token is retrieved from the cache.
   - Ensures that the client retries with the claims parameter and successfully obtains a valid token.

2. **TestAppServiceWithClientCapabilities**
   - Tests the inclusion of client capabilities in the request.
   - Verifies that the `client_capabilities` parameter is correctly passed and processed.


apps/managedidentity/managedidentity.go
```diff
@@ -192,14 +195,34 @@ type AcquireTokenOption func(o *AcquireTokenOptions)
...
-		o.claims = claims
+		if claims != "" {
+			o.claims = claims
+		}
+	}
+}
+
+// WithClientCapabilities sets the client capabilities to be used in the request.
+// This is used to enable specific features or behaviors in the token request.
+// The capabilities are passed as a slice of strings, and empty strings are filtered out.
```
bgavrilMS, line 206: The docs should point to CAE https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-continuous-access-evaluation

apps/managedidentity/managedidentity.go
```diff
@@ -323,28 +346,30 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
...
-		ar, err := base.AuthResultFromStorage(stResp)
-		if err == nil {
+	stResp, err := cacheManager.Read(ctx, c.authParams)
+	if err != nil {
+		return AuthResult{}, err
+	}
+	ar, err := base.AuthResultFromStorage(stResp)
+	if err == nil {
+		if o.claims != "" {
+			// When the claims are set, we need to passon bad/old token
```
bgavrilMS, line 356:
```suggestion
			// When the claims are set, we need to pass on revoked token to MSIv1 (AppService, ServiceFabric)
```

4gust: Updated the comment

apps/managedidentity/managedidentity.go
```diff
@@ -323,28 +346,30 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
...
+	stResp, err := cacheManager.Read(ctx, c.authParams)
+	if err != nil {
+		return AuthResult{}, err
+	}
+	ar, err := base.AuthResultFromStorage(stResp)
+	if err == nil {
+		if o.claims != "" {
+			// When the claims are set, we need to pass on revoked token to MSIv1 (AppService, ServiceFabric)
+			return c.getToken(ctx, resource, ar.AccessToken)
+		} else {
```
chlowell, line 358: the else is unnecessary because the if block returns

apps/managedidentity/managedidentity.go
```diff
@@ -323,28 +346,30 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
...
 			if !stResp.AccessToken.RefreshOn.T.IsZero() && !stResp.AccessToken.RefreshOn.T.After(now()) && c.canRefresh.CompareAndSwap(false, true) {
 				defer c.canRefresh.Store(false)
-				if tr, er := c.getToken(ctx, resource); er == nil {
+				if tr, er := c.getToken(ctx, resource, o.claims); er == nil {
```
chlowell, line 361:
```suggestion
				if tr, er := c.getToken(ctx, resource, ""); er == nil {
```
this has no functional change because `o.claims == ""` here, however it's confusing to see `o.claims` passed as `badToken` because that may imply the argument is misnamed or the above call should pass `o.claims` instead of `ar.AccessToken`

apps/managedidentity/managedidentity.go
```diff
@@ -192,14 +195,34 @@ type AcquireTokenOption func(o *AcquireTokenOptions)
...
+		if claims != "" {
+			o.claims = claims
+		}
+	}
+}
+
+// WithClientCapabilities sets the client capabilities to be used in the request.
+// For details see https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-continuous-access-evaluation
+// The capabilities are passed as a slice of strings, and empty strings are filtered out.
+func WithClientCapabilities(capabilities []string) ClientOption {
```
chlowell, line 207: As a caller wanting to enable graceful revocation, should I specify `CP1`?

chlowell (dismissed):

apps/managedidentity/managedidentity.go
bgavrilMS: +1 - an example of what a caller should do E2E would be great (cp1, claims parsing, token acquisition retry)

apps/managedidentity/managedidentity.go
```diff
@@ -323,28 +346,30 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
...
 			ar.AccessToken, err = c.authParams.AuthnScheme.FormatAccessToken(ar.AccessToken)
 			return ar, err
 		}
 	}
-	return c.getToken(ctx, resource)
+	return c.getToken(ctx, resource, "")
 }
 
-func (c Client) getToken(ctx context.Context, resource string) (AuthResult, error) {
+func (c Client) getToken(ctx context.Context, resource string, badToken string) (AuthResult, error) {
```
bgavrilMS, line 372: Call it "revokedToken" pls.

apps/managedidentity/managedidentity.go
```diff
@@ -569,16 +594,27 @@ func (c Client) getTokenForRequest(req *http.Request, resource string) (accessto
...
 	}
 	req.Header.Set("X-IDENTITY-HEADER", os.Getenv(identityHeaderEnvVar))
+
 	q := req.URL.Query()
 	q.Set("api-version", appServiceAPIVersion)
 	q.Set("resource", resource)
+
+	if badToken != "" {
+		hash := sha256.Sum256([]byte(badToken))
+		q.Set("token_sha256_to_refresh", hex.EncodeToString(hash[:]))
```
bgavrilMS, line 611: How do you know that this is the "right" format? Can you pls update the spec to include a test with an example

apps/managedidentity/managedidentity.go
4gust: I will attach an example to this

apps/managedidentity/managedidentity_test.go
```diff
@@ -1209,3 +1211,74 @@ func TestRefreshInMultipleRequests(t *testing.T) {
...
+			localUrl = r.URL
+		}))
+	// Reset cache for clean test
+	before := cacheManager
+	defer func() { cacheManager = before }()
+	cacheManager = storage.New(nil)
+
+	client, err := New(SystemAssigned(),
+		WithHTTPClient(mockClient),
+		WithClientCapabilities([]string{"c1", "c2"}))
```
bgavrilMS, line 1243: Can you have a E2E unit test that looks more like how the user would do this? With just "cp1"

apps/managedidentity/managedidentity.go
```diff
@@ -192,14 +195,34 @@ type AcquireTokenOption func(o *AcquireTokenOptions)
...
 	return func(o *AcquireTokenOptions) {
-		o.claims = claims
+		if claims != "" {
+			o.claims = claims
+		}
+	}
+}
+
+// WithClientCapabilities sets the client capabilities to be used in the request.
+// For details see https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-continuous-access-evaluation
```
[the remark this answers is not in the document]
4gust: updated

apps/managedidentity/managedidentity.go
4gust: Updated

apps/managedidentity/managedidentity_test.go
4gust: Added a example that shows that

apps/managedidentity/managedidentity.go
4gust: Updated

apps/managedidentity/managedidentity.go
4gust: Removed

apps/managedidentity/managedidentity_test.go
```diff
@@ -1209,3 +1211,74 @@ func TestRefreshInMultipleRequests(t *testing.T) {
...
+	}
+	if r := localUrlQuerry.Get(resourceQueryParameterName); strings.HasSuffix(r, "/.default") {
+		t.Fatal("suffix /.default was not removed.")
+	}
+	if localUrlQuerry.Get("xms_cc") != "c1,c2" {
+		t.Fatalf("Expected client capabilities %q, got %q", "c1,c2", localUrlQuerry.Get("xms_cc"))
+	}
+	hash := sha256.Sum256([]byte(token))
+	if localUrlQuerry.Get("token_sha256_to_refresh") != hex.EncodeToString(hash[:]) {
+		t.Fatalf("Expected token_sha256_to_refresh %q, got %q", hex.EncodeToString(hash[:]), localUrlQuerry.Get("token_sha256_to_refresh"))
```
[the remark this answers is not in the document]
4gust: Added a function do to the conversion and added a test with the example string `"test_token"` and output as `zAr5codUO2XaLH4UdkJgIYJsqxZvHgY+0BK4Vf+BllY=`

apps/managedidentity/managedidentity.go
4gust: Added a test

apps/managedidentity/example_test.go
```diff
@@ -0,0 +1,26 @@
...
+	miSystemassignedClient, err := mi.New(mi.SystemAssigned())
+	if err != nil {
+		// TODO: Handle error
+	}
+	_ = miSystemassignedClient
+
+	// User assigned Client
+	clientId := "ClientId" // TODO: replace with your Managed Identity Id
+
+	miClientIdAssignedClient, err := mi.New(mi.UserAssignedClientID(clientId), mi.WithClientCapabilities([]string{"cp1"}))
```
chlowell, line 21: I doubt this is going to work well as a code example. Readers know how to call functions and generally won't benefit from copying this code, so the content isn't really helpful. The interesting parts that need explanation are conceptual, like the difference between system- and user-assigned identity, and what "cp1" is about. As written, this implies that "cp1" is somehow related to, or required for, user-assigned identity, and overlooks the crucial functional implications of specifying it. A higher-level MSAL doc covering these concepts would be more helpful--does one of those exist?

apps/managedidentity/managedidentity.go
```diff
@@ -178,6 +180,7 @@ type Client struct {
 	authParams         authority.AuthParams
 	retryPolicyEnabled bool
 	canRefresh         *atomic.Value
+	clientCapabilities []string
```
chlowell, line 183: you can get a small perf improvement by saving the joined string instead

apps/managedidentity/example_test.go
4gust: I have updated the example, you thing this is explains these ?

chlowell: I think you want something like https://aka.ms/msal-net-managed-identity, and a separate document for cp1 (which may be only for 1p; do we recommend or expect 3p customers to specify capabilities?). Ideally, MSAL would have a single document on the Learn site explaining all the relevant concepts, with code samples for each supported language.

A function-level example is good for showing, with a little code the reader can copy, how to use an API when that isn't clear. This API doesn't need that because it's tiny and conventional. The conceptual explanation doesn't fit in as comments in the example because they overwhelm the code and make it hard to copy, and are themselves hard to use--the links aren't even clickable. If you have to explain concepts here because there's no better doc to link to, move that explanation into a comment on the example function so it's more readable and its links are clickable. And any explanation of cp1 belongs somewhere else because it's unrelated to system vs. user-assigned identity and entails claims handling.

chlowell (approved):
I approve of the Go content. Someone from the MSAL side should review the doc content

apps/managedidentity/servicefabric_test.go
```diff
@@ -96,6 +98,78 @@ func TestServiceFabricAcquireTokenReturnsTokenSuccess(t *testing.T) {
 		})
 	}
 }
+
+// TestAppServiceWithClaimsAndBadAccessToken tests the scenario where claims are passed
```
gladjohn, line 102:
```suggestion
// TestServiceFabricWithClaimsAndBadAccessToken tests the scenario where claims are passed
```

apps/managedidentity/managedidentity.go
```diff
@@ -192,14 +195,34 @@ type AcquireTokenOption func(o *AcquireTokenOptions)
...
+// The capabilities are passed as a slice of strings, and empty strings are filtered out.
+func WithClientCapabilities(capabilities []string) ClientOption {
+	return func(o *Client) {
+		var filteredCapabilities []string
+		for _, cap := range capabilities {
+			if cap != "" {
+				filteredCapabilities = append(filteredCapabilities, cap)
+			}
+		}
+		o.clientCapabilities = strings.Join(filteredCapabilities, ",")
```
gladjohn, line 215: allowing strings with leading/trailing spaces (" cp1 ") may cause AAD to ignore the header?

apps/managedidentity/managedidentity_test.go
```diff
@@ -1209,3 +1211,101 @@ func TestRefreshInMultipleRequests(t *testing.T) {
...
+			localUrl = r.URL
+		}))
+	// Reset cache for clean test
+	before := cacheManager
+	defer func() { cacheManager = before }()
+	cacheManager = storage.New(nil)
+
+	client, err := New(SystemAssigned(),
+		WithHTTPClient(mockClient),
+		WithClientCapabilities([]string{"c1", "c2"}))
```
gladjohn, line 1243: nit: Could we switch the test from c1,c2 → cp1,cp2? Functionality is identical, this just aligns the test data with the official capability names.

apps/managedidentity/managedidentity_test.go
```diff
@@ -1209,3 +1211,101 @@ func TestRefreshInMultipleRequests(t *testing.T) {
...
+		t.Fatalf("Expected access token %q, got %q", token, result.AccessToken)
+	}
+
+	// Call AcquireToken which should trigger token revocation flow
+	result, err = client.AcquireToken(context.Background(), resource, WithClaims("dummyClaims"))
+	if err != nil {
+		t.Fatalf("AcquireToken failed: %v", err)
+	}
+
+	localUrlQuerry := localUrl.Query()
```
gladjohn, line 1265: nit: localUrlQuery

apps/managedidentity/example_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Client Capabilities ("cp1", etc.)
+// 'cp1' is a capability that enables specific client behaviors — for example,
+// supporting Conditional Access policies that require additional client capabilities.
+// This is mostly relevant in scenarios where the identity is used to access resources
+// protected by policies like MFA or device compliance.
+// In most cases, you won't need to set this unless required by your Azure AD configuration.
+//
+// Learn more:
+// Client capabilities: https://learn.microsoft.com/en-us/entra/msal/python/advanced/client-capabilities
```
gladjohn, line 52:
```suggestion
// Client capabilities: https://learn.microsoft.com/entra/msal/python/advanced/client-capabilities
```

gladjohn:
@4gust this looks good, did not approve it because of the app service api-version update. anyway we can guard it?

apps/managedidentity/example_test.go
4gust: updated

apps/managedidentity/managedidentity_test.go
4gust: Updated

apps/managedidentity/managedidentity_test.go
4gust: Changed it to `cp1` and `cp2`

apps/managedidentity/managedidentity.go
4gust: Added a check and a test for the same

apps/managedidentity/servicefabric_test.go
```diff
@@ -96,6 +98,78 @@ func TestServiceFabricAcquireTokenReturnsTokenSuccess(t *testing.T) {
...
+	// Reset cache for clean test
+	before := cacheManager
+	defer func() { cacheManager = before }()
+	cacheManager = storage.New(nil)
+
+	client, err := New(SystemAssigned(),
+		WithHTTPClient(mockClient),
+		WithClientCapabilities([]string{"c1", "c2"}))
+	if err != nil {
+		t.Fatal(err)
```
Robbie-Microsoft, line 132: Nit (follow-up to gladjohn's comment on the managed identity test): this was updated to `cp1`/`cp2` in `managedidentity_test.go` but the Service Fabric test still uses `"c1"` and `"c2"`. The assertion on line ~529 also checks for `"c1,c2"`. Please apply the same rename here for consistency.

4gust:
- [ ]

PR #568, opened 2025-05-19
IMDS retry exponential backoff 

4gust:
### Description

This pull request introduces enhancements to the IMDS (Instance Metadata Service) retry logic in the Microsoft Authentication Library for Go (MSAL Go). The updates aim to improve the reliability and resilience of token acquisition when interacting with Azure IMDS endpoints, particularly under transient network failures or service unavailability scenarios.

**Key changes:**
- Adjusted backoff intervals and retry limits for improved fault tolerance.
- Added or updated corresponding unit tests to ensure correct retry behavior.

These improvements help ensure more consistent authentication experiences for applications running in Azure environments that rely on managed identities.

---

**Please review the changes and provide feedback or suggestions for further improvements.**

### DIFF
--- a/apps/managedidentity/managedidentity.go
+++ b/apps/managedidentity/managedidentity.go
@@ -21,6 +21,7 @@ import (
 	"runtime"
 	"strings"
 	"sync/atomic"
+	"math"
 	"time"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/errors"
@@ -508,8 +509,18 @@ func (c Client) retry(maxRetries int, req *http.Request) (*http.Response, error)
 		if err == nil && !contains(retrylist, resp.StatusCode) {
 			return resp, nil
 		}
+		// For IMDS, use exponential backoff based on attempt number
+		var waitTime time.Duration
+		if c.source == DefaultToIMDS {
+			// Exponential backoff with base of 1 second: 1s, 2s, 4s, 8s, etc.
+			waitTime = time.Second * time.Duration(1<<uint(attempt))
+		} else {
+			// For non-IMDS sources, use the fixed 1 second delay
+			waitTime = time.Second
+		}
+		
 		select {
-		case <-time.After(time.Second):
+		case <-time.After(waitTime):
 		case <-req.Context().Done():
 			err = req.Context().Err()
 			return resp, err
--- a/apps/managedidentity/managedidentity.go
+++ b/apps/managedidentity/managedidentity.go
@@ -21,7 +21,6 @@ import (
 	"runtime"
 	"strings"
 	"sync/atomic"
-	"math"
 	"time"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/errors"
@@ -518,7 +517,6 @@ func (c Client) retry(maxRetries int, req *http.Request) (*http.Response, error)
 			// For non-IMDS sources, use the fixed 1 second delay
 			waitTime = time.Second
 		}
-		
 		select {
 		case <-time.After(waitTime):
 		case <-req.Context().Done():
--- a/apps/managedidentity/managedidentity.go
+++ b/apps/managedidentity/managedidentity.go
@@ -99,6 +99,9 @@ const (
 	identityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
 
 	defaultRetryCount = 3
+
+	// maxRetryBackoff caps the exponential backoff wait between IMDS retries.
+	maxRetryBackoff = 60 * time.Second
 )
 
 var retryCodesForIMDS = []int{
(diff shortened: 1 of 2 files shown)


apps/managedidentity/managedidentity.go
```diff
@@ -508,8 +508,17 @@ func (c Client) retry(maxRetries int, req *http.Request) (*http.Response, error)
 		if err == nil && !contains(retrylist, resp.StatusCode) {
 			return resp, nil
 		}
+		// For IMDS, use exponential backoff based on attempt number
+		var waitTime time.Duration
+		if c.source == DefaultToIMDS {
+			// Exponential backoff with base of 1 second: 1s, 2s, 4s, 8s, etc.
```
chlowell, line 514: This is less backoff than recommended: https://learn.microsoft.com/entra/identity/managed-identities-azure-resources/how-to-use-vm-token#retry-guidance

apps/managedidentity/managedidentity_test.go
```diff
@@ -228,27 +229,78 @@ func TestRetryFunction(t *testing.T) {
...
+						actualDelays = append(actualDelays, time.Since(lastRequestTime))
+					}
+					lastRequestTime = time.Now()
+				}
+				// Apply callback only to retryable responses
+				if i < len(tt.mockResponses)-1 {
+					mockClient.AppendResponse(mock.WithBody(body.Bytes()), mock.WithHTTPStatusCode(resp.statusCode), mock.WithCallback(callback))
+				} else {
+					mockClient.AppendResponse(mock.WithBody(body.Bytes()), mock.WithHTTPStatusCode(resp.statusCode), mock.WithCallback(callback))
+				}
```
chlowell, line 285:
```go
if A {
    B
} else {
    B
}
```

apps/managedidentity/managedidentity.go
```diff
@@ -508,8 +508,17 @@ func (c Client) retry(maxRetries int, req *http.Request) (*http.Response, error)
...
+			waitTime = time.Second * time.Duration(1<<uint(attempt))
+		} else {
+			// For non-IMDS sources, use the fixed 1 second delay
+			waitTime = time.Second
+		}
 		select {
-		case <-time.After(time.Second):
+		case <-time.After(waitTime):
```
chlowell, line 521: This makes the tests slow. Consider enabling them to skip the wait, for example by adding a hook like this:
```go
var after = func(d time.Duration) <-chan time.Time {
	return time.After(d)
}

// in a test case
after = func(d time.Duration) <-chan time.Time {
    // TODO: validate d
    ch := make(chan time.Time, 1)
    ch <- time.Now()
    return ch
}
```
(has the additional benefit of preventing flakiness due to unpredictable scheduling)

chlowell (changes requested):

apps/managedidentity/managedidentity.go
4gust: Updated the time to reflect the time.

Robbie-Microsoft: Still not addressed. The current code starts at `1<<0 = 1s` for the first attempt (1s, 2s, 4s, 8s…). The [IMDS retry guidance](https://learn.microsoft.com/entra/identity/managed-identities-azure-resources/how-to-use-vm-token#retry-guidance) recommends starting at 2s (2^retry), so the series should be 2s, 4s, 8s, 16s…. Also worth noting: there's currently no cap on the wait duration — at attempt 10 the code would wait 1024s. Recommend something like:

```go
waitTime := time.Duration(2<<uint(attempt)) * time.Second
if waitTime > 60*time.Second {
    waitTime = 60 * time.Second
}
```

apps/managedidentity/managedidentity.go
Robbie-Microsoft: Still not addressed. The current diff still has `case <-time.After(waitTime):` with no overrideable hook. The new test cases actually sleep for real (1s + 2s + 4s = 7s for the IMDS exponential backoff case alone) and then validate wall-clock timing with ±500ms tolerance — exactly the flakiness concern raised here.

apps/managedidentity/managedidentity_test.go
Robbie-Microsoft: Still present in the updated diff. Lines 91–95 of the test have identical branches:

```go
if i < len(tt.mockResponses)-1 {
    mockClient.AppendResponse(..., mock.WithCallback(callback))
} else {
    mockClient.AppendResponse(..., mock.WithCallback(callback)) // identical
}
```

The condition does nothing — both branches call the same thing. Should just be:
```go
mockClient.AppendResponse(mock.WithBody(body.Bytes()), mock.WithHTTPStatusCode(resp.statusCode), mock.WithCallback(callback))
```

apps/managedidentity/managedidentity.go
4gust: Addressed in fb07b8b. Backoff now starts at 2s and follows 2^attempt (2s, 4s, 8s, 16s...) per the IMDS retry guidance, and is capped at 60s via a new maxRetryBackoff const.

apps/managedidentity/managedidentity.go
4gust: Addressed in fb07b8b. Added a package-private fter hook wrapping 	ime.After; the backoff select now uses fter(waitTime). Both TestRetryFunction and Test_RetryPolicy_For_AcquireToken override it to skip real sleeps, so the tests are deterministic and fast (no more wall-clock timing with tolerance).

apps/managedidentity/managedidentity_test.go
4gust: Addressed in fb07b8b. Collapsed the identical branches into a single AppendResponse call. Delay assertions now compare exact durations captured via the fter hook instead of measuring wall-clock time.

Issue #574, opened 2025-07-05
[Bug] login.partner.microsoftonline.cn does not work for China cloud


andyzhangx:
**Which version of MSAL Go are you using?**
Note that to get help, you need to run the latest version. 

related to:
https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/internal/oauth/ops/authority/authority.go#L51

can we revert the endpoint for china cloud to `login.chinacloudapi.cn`?

#### What happened:
When upgrading cloud provider azure from `v1.31.1` to` v1.31.2` on azure china environment , the cloud controller manager fails to initialize with the following error msg 

```bash
E0704 15:23:02.162092       1 azure.go:567] InitializeCloudFromConfig: failed to sync regional zones map for the first time: list zones: ClientSecretCredential: unable to resolve an endpoint: server response error:
 context deadline exceeded
F0704 15:23:02.162138       1 controllermanager.go:356] Cloud provider azure could not be initialized: could not init cloud provider azure: list zones: ClientSecretCredential: unable to resolve an endpoint: server response error:
 context deadline exceeded
```

#### What you expected to happen:
We expected that the cloud controller manager would be Ready and not fail during initializing stage
#### How to reproduce it (as minimally and precisely as possible):
Deploy at least the 1.31.2 version of the cloud provider azure addon with the azure environment `AzureChinaCloud` 
#### Anything else we need to know?:
We saw that starting from the v1.31.2 revision, the go package responsible for the azure AD authentication `AzureAD/microsoft-authentication-library-for-go`  was bumped to a newer version  that updated the trusted azureAD hosts

<img width="1410" height="240" alt="Image" src="https://github.com/user-attachments/assets/2d6c6d9a-8fc3-42d8-8027-b84ed77c2f62" />

The problem that we suspect is that the Environment variable for the china cloud was not adapted to target the new azureAD endpoint `login.partner.microsoftonline.cn` it is always set to `login.chinacloudapi.cn`

To confirm that, we did a quick test with the cloud provider azure v1.31.2 :
- changed the `aadTrustedHostList` map to set the old china login endpoint `login.chinacloudapi.cn`
- build the image only for the cloud controller manager
- Test the cloud provider azure v1.31.2 with our custom cloud controller manager image
Result: the cloud controller manager was initialized without issues   

#### Environment:

- Kubernetes version (use `kubectl version`): 1.31
[...]

PR #575, opened 2025-07-05
fix: support both login.partner.microsoftonline.cn and login.chinaclo…

Achiket123:
here problem(#574 ) is when we upgrad the azure thing and the ms auth thing for go, it break in china cloud. it not login good anymore. new lib only like login.partner.microsoftonline.cn but lot of old stuffs still use login.chinacloudapi.cn. so it not work and the cloud thingy not start.

i make this pr to fix it. now we say okay to both old and new login link in the aadTrustedHostList.

* add login.partner.microsoftonline.cn (new one)
* keep login.chinacloudapi.cn (old one lots people still use)

so now both new and old can login. no more break. i also write small note in code so we remember to check this again later maybe.

this help people move slowly to new thing without everything go boom. it only fix china cloud. other clouds is okay.

this fix the bug where login.partner.microsoftonline.cn no work in china cloud. it also related to the azure provider and the ms login thing change.

### DIFF
--- a/apps/internal/oauth/ops/authority/authority.go
+++ b/apps/internal/oauth/ops/authority/authority.go
@@ -46,9 +46,11 @@ type jsonCaller interface {
 	JSONCall(ctx context.Context, endpoint string, headers http.Header, qv url.Values, body, resp interface{}) error
 }
 
+// For backward compatibility, accept both old and new China endpoints for a transition period.
 var aadTrustedHostList = map[string]bool{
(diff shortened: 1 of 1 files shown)


andyzhangx:
@bgavrilMS @rayluo can you approve this PR? this is breaking AKS cloud provider in China cloud now.

bgavrilMS (approved):

Issue #411, opened 2023-04-14
[Feature Request] Support CIAM authorities


chlowell:
See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4054

PR #576, opened 2025-07-17
Add validation for issuer returned by OIDC endpoint

Avery-Dunn:
Adds new validation according to: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3268770

The issuer returned by the call to the OIDC endpoint (.well-known/openid-configuration) will now be validated by a few validation steps:
-If the issuer is a known Microsoft host it's valid, even if the authority doesn't match
-Otherwise, the host and scheme of the issuer must match the authority (the paths can be ignored)

(Special conditions for CIAM custom authorities were added in other MSALs, but it appears CIAM is not supported in MSAL Go so this was not added here: https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/411)

The behavior was added to `authority.go`, and a new test function was added to `authority_test.go` to cover it.

In addition, changes were made to other test files to fix issues exposed by the new validation. These appeared to be purely issues with the test configuration, not a bug in the actual library:
- `confidential_test.go`: A full URL was being passed to a mocked discovery response instead of a host like the function expected, causing the mocked discovery response to have badly formatted URLs 
  - For example, if the test used "https://some.authority/" it would create a response with URLs like "https://https://some.authority//v2.0"
- `public_test.go`: Same as above, except it was also using "localhost" instead of something like "login.microsoftonline.com"
  - E2E tests in other MSALs shows that the real discovery in that scenario would use "login.microsoftonline.com", and the Go test still worked when changed "login.microsoftonline.com"

### DIFF
--- a/apps/internal/oauth/ops/authority/authority_test.go
+++ b/apps/internal/oauth/ops/authority/authority_test.go
@@ -553,8 +553,14 @@ func TestTenantDiscoveryValidateIssuer(t *testing.T) {
 			expectError: false,
 		},
 		{
-			desc:        "issuer and authority have different hosts",
-			issuer:      "https://other.microsoftonline.com/tenant-id",
+			desc:        "custom authority with a non-matching Entra issuer",
+			issuer:      "https://login.microsoftonline.com/",
+			authority:   "https://contoso.com/tenant-id",
+			expectError: true,
+		},
+		{
+			desc:        "Entra authority with a non-matching custom issuer",
+			issuer:      "https://contoso.com/",
 			authority:   "https://login.microsoftonline.com/tenant-id",
 			expectError: true,
 		},


apps/internal/oauth/ops/authority/authority.go
```diff
@@ -98,6 +98,46 @@ func (r *TenantDiscoveryResponse) Validate() error {
...
+	}
+
+	// Even if it doesn't match the authority, issuers from known and trusted Microsoft hosts are valid
+	if TrustedHost(issuerURL.Host) {
+		return nil
+	}
+
+	// Check against other trusted Microsoft hosts
+	switch issuerURL.Host {
+	case loginMicrosoft, loginWindows, loginSTSWindows, defaultHost:
```
bgavrilMS: There are a bunch of aliases on soverigns too

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -98,6 +98,46 @@ func (r *TenantDiscoveryResponse) Validate() error {
...
+		return fmt.Errorf("TenantDiscoveryResponse: failed to parse issuer URL: %w", err)
+	}
+
+	// Even if it doesn't match the authority, issuers from known and trusted Microsoft hosts are valid
+	if TrustedHost(issuerURL.Host) {
+		return nil
+	}
+
+	// Check against other trusted Microsoft hosts
+	switch issuerURL.Host {
```
bgavrilMS: Can't this check be done after instance discovery? That would be future proof.

4gust (changes requested):
There are some test cases missing in both the Confidential and Public test suites. Specifically:

Error responses are not being tested, it would be good to cover scenarios where the input is invalid or triggers a known failure.

The current tests focus on the happy path (valid E2E flow), but they don't include negative cases or verify that the system returns the correct errors for invalid or unexpected input.

Adding these would improve coverage and confidence in the error-handling logic.

apps/internal/oauth/ops/authority/authority.go
4gust: The `openIDConfigurationEndpoint` in this `resp` response contains metadata and all the aliases, we can check this through issuer validation.

Avery-Dunn: @4gust : It looks like the `InstanceDiscoveryResponse` has the alias metadata, so are you talking about the `resp` down in the `AADInstanceDiscovery` function? How would I access the `InstanceDiscoveryResponse` from the new validation function?

4gust: You can add a new `Aliases` field to the `cache` struct inside `authorityEndpoint`. Include corresponding setter and getter methods.

These methods should update the `AADInstanceDiscovery` within the `openIDConfigurationEndpoint`.

This way, aliases can be managed per `authorityEndpoint` and accessed externally as needed, while keeping the logic scoped and consistent.

apps/internal/oauth/ops/authority/authority.go
Avery-Dunn: In the latest commit, it now uses aliases from instance discovery rather than the hardcoded list in the code.

apps/internal/oauth/ops/authority/authority.go
Avery-Dunn: @4gust : In the latest commit I refactored to match what you suggested:
- New `Aliases` field in `cache`
- `Aliases` are sent to the validation function by referencing `authorityEndpoint,cache` entry

Let me know if I'm missing anything with this new implementation. Also, I added new unit tests for this new behavior, and expanded the existing ones to include a few more failure cases.

Avery-Dunn: > There are a bunch of aliases on soverigns too

@bgavrilMS : Forgot to respond to this before the newest changes, but the sovereign clouds were covered by a different check a few lines above (`if TrustedHost(issuerURL.Host)`)

However, as of the latest commit that check was removed and replaced with a check on the actual aliases returned from  instance discovery.

apps/internal/oauth/ops/authority/authority_test.go
```diff
@@ -507,3 +507,115 @@ func TestMergeCapabilitiesAndClaims(t *testing.T) {
...
+			desc:        "issuer doesn't match any alias",
+			issuer:      "https://unknown.example.com/tenant-id",
+			authority:   "https://contoso.com/tenant-id",
+			aliases:     map[string]bool{"alias1.example.com": true, "alias2.example.com": true},
+			expectError: true,
+		},
+		{
+			desc:        "empty aliases map",
+			issuer:      "https://unknown.example.com/tenant-id",
+			authority:   "https://contoso.com/tenant-id",
```
bgavrilMS, line 605: I think @rayluo stated that there should be a test case like

`issuer: https://login.microsoftonline.com/tid`
`authority: https://contoso.com`

And that we expect this to fail.

bgavrilMS: (assuming regular aliases)

Avery-Dunn: Just added that in the latest commit, the reverse was covered (Entra authority/custom issuer) but not that case.

PR #578, opened 2025-08-28
Fix concurrent cache access in AcquireToken method

### LINKED ISSUE #569: [Bug] Concurrent managedidentity.Client.AcquireToken() calls cause multiple token requests to identity provider before caching
Latest in main

**Where is the issue?**
* Public client
    * [ ] Device code flow
    * [ ] Authorization code flow 
* Confidential client
    * [ ] Authorization code flow 
    * [ ] Client credentials:
        * [ ] client secret
        * [ ] client certificate
* Token cache serialization
     * [X] In-memory cache
* Other (please describe)
  * Managed identity client/token caching 

The app is in production, although the area with most likelihood of issue have the mitigation installed.

<pre>
$ go version
go version go1.24.3 linux/amd64
</pre>

<details><summary><code>go env</code> Output</summary><br><pre>
$ go env
</pre></details>

**Repro**
Adding this unit test to `managedidentity` package:
```go
package managedidentity

import (
	"context"
	"encoding/json"
	"fmt"
	"net/http"
	"sync"
	"testing"
	"time"

	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base/storage"
	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/mock"
)

func TestAcquireTokenConcurrency(t *testing.T) {
	resource := "https://management.azure.com"
	miType := SystemAssigned()
	setEnvVars(t, DefaultToIMDS)
	before := cacheManager
	defer func() { cacheManager = before }()
	cacheManager = storage.New(nil)

	// Track the number of HTTP requests made to IMDS, and the number of unique tokens received from AcquireToken
	// Optimally, both are supposed to be 1, as the first request should write to the cache, then become reusable
	var requestCount int32
	var requestCountMutex sync.Mutex
	var acquiredTokens []string
	var acquiredTokensMutex sync.Mutex

	tries := 10

	// Assume no caching on token provider server
	mockClient := mock.NewClient()
	for i := 0; i < tries; i++ {
		token := fmt.Sprintf("[%d]", i)
		responseBody, err := json.Marshal(SuccessfulResponse{
			AccessToken: token,
			ExpiresIn:   3600,
			ExpiresOn:   time.Now().Add(time.Hour).Unix(),
			Resource:    resource,
			TokenType:   "Bearer",
		})
		if err != nil {
			t.Fatal(err)
		}

		mockClient.AppendResponse(
			mock.WithHTTPStatusCode(http.StatusOK),
			mock.WithBody(responseBody),
			mock.WithCallback(func(r *http.Request) {
				requestCountMutex.Lock()
				t.Logf("token provider server providing token: %s", token)
				requestCount++
				requestCountMutex.Unlock()
			}),
		)
	}
	client, err := New(miType, WithHTTPClient(mockClient))
	if err != nil {
		t.Fatal(err)
	}

	// Launch multiple goroutines for AcquireToken() simultaneously
	numGoroutines := tries
	var wg sync.WaitGroup
	for i := 0; i < numGoroutines; i++ {
		wg.Add(1)
		go func(routineID int) {
			defer wg.Done()

			// Call AcquireToken in each goroutine
			result, err := client.AcquireToken(context.Background(), resource)
			if err != nil {
				t.Errorf("goroutine %d failed: %v", routineID, err)
				return
			}

			// Capture the token received
			acquiredTokensMutex.Lock()
			t.Logf("AcquireToken() caller receives token: %s", result.AccessToken)
			acquiredTokens = append(acquiredTokens, result.AccessToken)
			acquiredTokensMutex.Unlock()
		}(i)
	}
	wg.Wait()
	// Although, in the current implementation, we expect
	// - Multiple HTTP requests to IMDS until the cache is written and the subsequent call/goroutine sees the cached token
	// - Each new token received from the server overwrites the previous one in cache
	// - Different goroutines may receive different tokens

	uniqueTokens := make(map[string]bool)
	for _, token := range acquiredTokens {
		uniqueTokens[token] = true
	}

	// Log the results to show the race condition
	t.Logf("- Total HTTP requests made: %d", requestCount)
	t.Logf("- All tokens captured: %v", acquiredTokens)
	t.Logf("- Unique tokens received: %d", len(uniqueTokens))
}
```

**Expected behavior**
[AcquireToken(...)](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/managedidentity/managedidentity.go#L318) [sends the request to identity provider](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/managedidentity/managedidentity.go#L379-L383) once, [caches](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/managedidentity/managedidentity.go#L463), and subsequent (concurrent) calls to `AcquireToken(...)` [reuses the cache](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/managedidentity/managedidentity.go#L328-L340).
```
$ go test -v -run TestAcquireTokenConcurrency
=== RUN   TestAcquireTokenConcurrency
    race_condition_test.go:53: token provider server providing token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:98: - Total HTTP requests made: 1
    race_condition_test.go:99: - All tokens captured: [[0] [0] [0] [0] [0] [0] [0] [0] [0] [0]]
    race_condition_test.go:100: - Unique tokens received: 1
```

**Actual behavior**
[Cache read](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/managedidentity/managedidentity.go#L328) of the subsequent (concurrent) calls may occur before [cache write](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/b4b8bfc9569042572ccb82b648ea509075fadb74/apps/managedidentity/managedidentity.go#L463) by the previous call(s). This have resulted in these calls also sending requests to identity provider, when it could have used the previous one.
Examples:
```
=== RUN   TestAcquireTokenConcurrency
    race_condition_test.go:55: token provider server providing token: [0]
    race_condition_test.go:83: AcquireToken() caller receives token: [0]
    race_condition_test.go:55: token provider server providing token: [1]
    race_condition_test.go:55: token provider server providing token: [2]
    race_condition_test.go:83: AcquireToken() caller receives token: [0]
    race_condition_test.go:83: AcquireToken() caller receives token: [0]
    race_condition_test.go:83: AcquireToken() caller receives token: [0]
    race_condition_test.go:83: AcquireToken() caller receives token: [0]
    race_condition_test.go:83: AcquireToken() caller receives token: [0]
    race_condition_test.go:83: AcquireToken() caller receives token: [0]
    race_condition_test.go:83: AcquireToken() caller receives token: [0]
    race_condition_test.go:83: AcquireToken() caller receives token: [1]
    race_condition_test.go:83: AcquireToken() caller receives token: [2]
    race_condition_test.go:100: - Total HTTP requests made: 3
    race_condition_test.go:101: - All tokens captured: [[0] [0] [0] [0] [0] [0] [0] [0] [1] [2]]
    race_condition_test.go:102: - Unique tokens received: 3
```
```
=== RUN   TestAcquireTokenConcurrency
    race_condition_test.go:53: token provider server providing token: [0]
    race_condition_test.go:53: token provider server providing token: [1]
    race_condition_test.go:53: token provider server providing token: [2]
    race_condition_test.go:53: token provider server providing token: [3]
    race_condition_test.go:53: token provider server providing token: [4]
    race_condition_test.go:81: AcquireToken() caller receives token: [1]
    race_condition_test.go:81: AcquireToken() caller receives token: [2]
    race_condition_test.go:81: AcquireToken() caller receives token: [3]
    race_condition_test.go:81: AcquireToken() caller receives token: [3]
    race_condition_test.go:81: AcquireToken() caller receives token: [0]
    race_condition_test.go:81: AcquireToken() caller receives token: [4]
    race_condition_test.go:53: token provider server providing token: [5]
    race_condition_test.go:81: AcquireToken() caller receives token: [5]
    race_condition_test.go:53: token provider server providing token: [6]
    race_condition_test.go:81: AcquireToken() caller receives token: [6]
    race_condition_test.go:53: token provider server providing token: [7]
    race_condition_test.go:81: AcquireToken() caller receives token: [7]
    race_condition_test.go:53: token provider server providing token: [8]
    race_condition_test.go:81: AcquireToken() caller receives token: [8]
    race_condition_test.go:98: - Total HTTP requests made: 9
    race_condition_test.go:99: - All tokens captured: [[1] [2] [3] [3] [0] [4] [5] [6] [7] [8]]
    race_condition_test.go:100: - Unique tokens received: 9
```

**Possible solution**
The evaluation of whether to fetch the new token should be made after the cache is stabilized. Maybe wrap the logic of `AcquireToken()` with a lock?

**Additional context / logs / screenshots**
[One of its uses](https://github.com/Azure/azure-sdk-for-go/blob/8af5755b00fd6847a7a89dbc1996fc03350d73d7/sdk/azidentity/managed_identity_client.go#L189) in azure-sdk-for-go seems to [lock this](https://github.com/Azure/azure-sdk-for-go/blob/8af5755b00fd6847a7a89dbc1996fc03350d73d7/sdk/internal/temporal/resource.go#L78-L140) anyway. However, I don't think this lock requirement is intentional(?), and we might just get lucky that this use case locks this(?).
My use case is not through azure-sdk-for-go, so we have to install a lock like azure-sdk-for-go at the moment.

4gust:
#### **Summary**

This PR addresses a potential race condition in the `AcquireToken` method by properly synchronizing access to the cache using a mutex. Issue number #569

---

#### **Changes**

* Introduced `cacheAccessorMu.Lock()` / `defer cacheAccessorMu.Unlock()` around cache operations to ensure thread-safe access.
* Removed the `canRefresh` atomic variable, which was previously used to prevent concurrent refreshes which is no longer needed as whole cache is in mutex

---

#### **Testing**

* Validated using the existing `TestAcquireTokenConcurrency` unit test to ensure correct behavior under concurrent conditions.

---

#### **Breaking Changes**

* None. This is an internal, non-breaking implementation change that does not affect the public API.

### DIFF
--- a/apps/managedidentity/managedidentity.go
+++ b/apps/managedidentity/managedidentity.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -170,6 +171,7 @@ func SystemAssigned() ID {
 
 // cache never uses the client because instance discovery is always disabled.
 var cacheManager *storage.Manager = storage.New(nil)
+var cacheAccessorMu *sync.RWMutex = &sync.RWMutex{}
 
 type Client struct {
 	httpClient         ops.HTTPClient
@@ -177,7 +179,6 @@ type Client struct {
 	source             Source
 	authParams         authority.AuthParams
 	retryPolicyEnabled bool
-	canRefresh         *atomic.Value
 }
 
 type AcquireTokenOptions struct {
@@ -267,7 +268,6 @@ func New(id ID, options ...ClientOption) (Client, error) {
 		httpClient:         shared.DefaultClient,
 		retryPolicyEnabled: true,
 		source:             source,
-		canRefresh:         &zero,
 	}
 	for _, option := range options {
 		option(&client)
@@ -323,6 +323,8 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
 	}
 	c.authParams.Scopes = []string{resource}
 
+	cacheAccessorMu.Lock()
+	defer cacheAccessorMu.Unlock()
 	// ignore cached access tokens when given claims
 	if o.claims == "" {
 		stResp, err := cacheManager.Read(ctx, c.authParams)
@@ -331,8 +333,7 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
 		}
 		ar, err := base.AuthResultFromStorage(stResp)
 		if err == nil {
-			if !stResp.AccessToken.RefreshOn.T.IsZero() && !stResp.AccessToken.RefreshOn.T.After(now()) && c.canRefresh.CompareAndSwap(false, true) {
-				defer c.canRefresh.Store(false)
+			if !stResp.AccessToken.RefreshOn.T.IsZero() && !stResp.AccessToken.RefreshOn.T.After(now()) {
 				if tr, er := c.getToken(ctx, resource); er == nil {
 					return tr, nil
 				}
--- a/apps/managedidentity/managedidentity.go
+++ b/apps/managedidentity/managedidentity.go
@@ -21,7 +21,6 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/errors"
@@ -261,8 +260,6 @@ func New(id ID, options ...ClientOption) (Client, error) {
 	default:
 		return Client{}, fmt.Errorf("unsupported type %T", id)
 	}
-	zero := atomic.Value{}
-	zero.Store(false)
 	client := Client{
 		miType:             id,
 		httpClient:         shared.DefaultClient,
--- a/apps/managedidentity/managedidentity.go
+++ b/apps/managedidentity/managedidentity.go
@@ -170,7 +170,34 @@ func SystemAssigned() ID {
 
 // cache never uses the client because instance discovery is always disabled.
 var cacheManager *storage.Manager = storage.New(nil)
-var cacheAccessorMu *sync.RWMutex = &sync.RWMutex{}
+
+// refreshLocks deduplicates concurrent token refreshes by cache key. Each
+// (identity, resource) pair gets its own sync.Mutex, so refreshes for different
+// keys run in parallel while concurrent callers for the same key serialize and
+// share a single in-flight IMDS call (the waiters re-check the cache after
+// acquiring the lock and return the freshly written token).
+//
+// storage.Manager has its own internal locking, so cache Read/Write do not need
+// an additional outer mutex.
+var refreshLocks = struct {
+	mu sync.Mutex
+	m  map[string]*sync.Mutex
+}{m: map[string]*sync.Mutex{}}
+
+// refreshLockFor returns the per-key mutex for the given cache key, creating it
+// on first use. The number of unique (identity, resource) keys per process is
+// bounded by application configuration, so the map does not grow unboundedly in
+// practice.
+func refreshLockFor(key string) *sync.Mutex {
+	refreshLocks.mu.Lock()
+	defer refreshLocks.mu.Unlock()
+	m, ok := refreshLocks.m[key]
+	if !ok {
+		m = &sync.Mutex{}
+		refreshLocks.m[key] = m
+	}
+	return m
+}
 
 type Client struct {
 	httpClient         ops.HTTPClient
@@ -308,6 +335,30 @@ func GetSource() (Source, error) {
 // was created to test the function against refreshin
 var now = time.Now
 
+// cacheLookup classifies the cache state for the current authParams.
+//   - hit  == true : a fresh, formatted token is returned in result; caller may return immediately.
+//   - hit  == false && stale == true : result holds a still-valid (formatted) token whose
+//     RefreshOn has passed; caller should attempt a proactive refresh and fall back to result on failure.
+//   - hit  == false && stale == false : cache miss; result is unset, caller must refresh.
+//   - err  != nil : a real cache read error; caller should propagate.
+func (c Client) cacheLookup(ctx context.Context) (result AuthResult, hit, stale bool, err error) {
+	stResp, err := cacheManager.Read(ctx, c.authParams)
+	if err != nil {
+		return AuthResult{}, false, false, err
+	}
+	ar, err := base.AuthResultFromStorage(stResp)
+	if err != nil {
+		// no usable cached token; treat as miss, not as error
+		return AuthResult{}, false, false, nil
+	}
+	stale = !stResp.AccessToken.RefreshOn.T.IsZero() && !stResp.AccessToken.RefreshOn.T.After(now())
+	ar.AccessToken, err = c.authParams.AuthnScheme.FormatAccessToken(ar.AccessToken)
+	if err != nil {
+		return AuthResult{}, false, false, err
+	}
+	return ar, !stale, stale, nil
+}
+
 // Acquires tokens from the configured managed identity on an azure resource.
 //
 // Resource: scopes application is requesting access to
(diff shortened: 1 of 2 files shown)


bgavrilMS (approved):

apps/managedidentity/managedidentity.go
```diff
@@ -267,7 +268,6 @@ func New(id ID, options ...ClientOption) (Client, error) {
 		httpClient:         shared.DefaultClient,
 		retryPolicyEnabled: true,
 		source:             source,
-		canRefresh:         &zero,
```
chlowell, line 270: Also remove the declaration of `zero`

apps/managedidentity/managedidentity_test.go
```diff
@@ -1209,3 +1209,113 @@ func TestRefreshInMultipleRequests(t *testing.T) {
...
+		}(i)
+	}
+
+	wg.Wait()
+	close(errors)
+
+	// Check for any errors from goroutines
+	for err := range errors {
+		t.Error(err)
+	}
```
chlowell: Consider [this approach](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/a3dbb6d787f4218a51c0e08975337888aadc9899/apps/managedidentity/managedidentity_test.go#L1179-L1206) instead. If there were 100 identical errors, would you want to see all of them in `go test` output?

apps/managedidentity/managedidentity_test.go
```diff
@@ -1209,3 +1209,113 @@ func TestRefreshInMultipleRequests(t *testing.T) {
...
+		Resource:    resource,
+		TokenType:   "Bearer",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Mock client should only need to respond once if caching works correctly
+	mockClient := mock.NewClient()
+	// Add multiple responses in case caching fails (but we'll verify it doesn't)
```
chlowell: if you simply appended 1 response the mock would panic when it gets a second request and you wouldn't have to count the requests yourself

apps/managedidentity/managedidentity_test.go
```diff
@@ -1209,3 +1209,113 @@ func TestRefreshInMultipleRequests(t *testing.T) {
...
+			}
+
+			// Verify the token is correct
+			if result.AccessToken != expectedToken {
+				errors <- fmt.Errorf("goroutine %d: expected token %q, got %q",
+					routineID, expectedToken, result.AccessToken)
+				return
+			}
+
+			// Capture the token received
```
chlowell: why? There can only be one token because the mock client returns a static value, and you don't need this to know whether the goroutine succeeded because if it didn't it would have written an error to the channel

chlowell (changes requested):

apps/managedidentity/managedidentity.go
```diff
@@ -323,6 +320,8 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
 	}
 	c.authParams.Scopes = []string{resource}
 
+	cacheAccessorMu.Lock()
+	defer cacheAccessorMu.Unlock()
```
Robbie-Microsoft: The lock is held for the entire duration of `AcquireToken`, including the `c.getToken()` HTTP call on the refresh path. This means all goroutines calling `AcquireToken` are fully serialized globally — if IMDS is slow or times out (up to 30s), every other goroutine blocks.

The original `canRefresh` CAS was more surgical: it only gated concurrent refreshes; cache hits were still concurrent. Consider holding the lock only around cache reads and writes, not across the network call:

```go
cacheAccessorMu.RLock()
stResp, err := cacheManager.Read(...)
ar, err := base.AuthResultFromStorage(stResp)
cacheAccessorMu.RUnlock()

if needsRefresh {
    tr, er := c.getToken(ctx, resource) // no lock held during HTTP
    if er == nil {
        // re-acquire write lock to update cache
        return tr, nil
    }
}
```

apps/managedidentity/managedidentity.go
```diff
@@ -170,14 +170,14 @@ func SystemAssigned() ID {
 
 // cache never uses the client because instance discovery is always disabled.
 var cacheManager *storage.Manager = storage.New(nil)
+var cacheAccessorMu *sync.RWMutex = &sync.RWMutex{}
```
Robbie-Microsoft: `cacheAccessorMu` is declared as `*sync.RWMutex` but `Lock()` (exclusive write lock) is always called — `RLock()` is never used. This gives no benefit over a plain `sync.Mutex` and adds confusion. Either switch to `sync.Mutex`:

```go
var cacheAccessorMu sync.Mutex
```

or use `RLock`/`RUnlock` for the read-only cache lookup path and reserve `Lock` only for writes (which also addresses the HTTP-hold concern above).

Robbie-Microsoft:
**Design note: global mutex serializes across all Client instances**

Both `cacheManager` and the new `cacheAccessorMu` are package-level globals, so all `Client` instances — regardless of resource or identity — contend on the same lock. A caller that creates two clients for different resources will have their token acquisitions serialized against each other even though they have no shared state at the application level.

This is a pre-existing consequence of `cacheManager` being global, but the new mutex makes it more visible. It may be worth considering a per-client `sync.Mutex` paired with a per-client cache, or at minimum documenting the serialization behavior so callers know concurrent multi-resource clients don't actually run concurrently.

Issue #580, opened 2025-09-04
[Bug] panic during AcquireTokenSilent


weikanglim:
**Which version of MSAL Go are you using?**
1.4.2

**Where is the issue?**
* Public client
    * [ ] Device code flow
    * [x] Username/Password (ROPC grant)
    * [ ] Authorization code flow 
* Confidential client
    * [ ] Authorization code flow 
    * [ ] Client credentials:
        * [ ] client secret
        * [ ] client certificate
* Token cache serialization
     * [ ] In-memory cache
* Other (please describe)

**Is this a new or an existing app?**
The app is in production and I have upgraded to a new version of Microsoft Authentication Library for Go.

**What version of Go are you using (`go version`)?**
1.25

**What operating system and processor architecture are you using (`go env`)?**
Linux/amd64 (go env not available from user report)

**Repro**

https://github.com/Azure/azure-dev/issues/5685
 
User waited 2 days for auth token to expire, triggered `AcquireTokenSilent`, panic was observed:

```
goroutine 1 [running]:
github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth.(*Client).Credential(_, {_, _}, {{{0x40004fd868, 0x19}, {0x40004fd860, 0x47}, {0xaaaacdc9227d, 0x5}, 0x1, ...}, ...}, ...)
        github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/oauth/oauth.go:104 +0x48
github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base.Client.AcquireTokenSilent({0x400050d000, {0xaaaace730ad0, 0x400049f450}, {0xaaaace72a828, 0x400049f4a0}, {{{0x40004fd868, 0x19}, {0x40004fd860, 0x47}, {0xaaaacdc9227d, ...}, ...}, ...}, ...}, ...)
        github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/base/base.go:370 +0xa24
github.com/AzureAD/microsoft-authentication-library-for-go/apps/public.Client.AcquireTokenSilent({{0x400050d000, {0xaaaace730ad0, 0x400049f450}, {0xaaaace72a828, 0x400049f4a0}, {{{...}, {...}, {...}, 0x1, {...}, ...}, ...}, ...}}, ...)
[...]

PR #581, opened 2025-09-09
Fix panic in AcquireTokenSilent for public clients due to nil Credential

4gust:
**Fix panic in AcquireTokenSilent for public clients due to nil Credential**

### Description

This PR addresses a `nil` panic occurring during `AcquireTokenSilent` calls for **public clients** in the Microsoft Authentication Library for Go. #580 

### Root Cause

In `apps/internal/base/base.go`, the `Token.Credential(...)` function is called unconditionally within `AcquireTokenSilent`, even for **public clients**. This function expects a non-nil `*accesstokens.Credential` (specifically, `silent.Credential`), but for public clients, this value is `nil`.

The panic occurs in `apps/internal/oauth/oauth.go` at line 104:

```go
if cred.TokenProvider != nil {
```

Here, `cred` is `nil`, leading to a runtime panic.

### Fix

Added a conditional check to ensure that `Token.Credential(...)` is only called for confidential client flows (i.e., when `silent.RequestType == accesstokens.ATConfidential`), preventing the use of a `nil` credential for public clients.

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -367,8 +367,10 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 					// If the token is not same, we don't need to refresh it.
 					// Which means it refreshed.
 					if str, err := m.Read(ctx, authParams); err == nil && str.AccessToken.Secret == ar.AccessToken {
-						if tr, er := b.Token.Credential(ctx, authParams, silent.Credential); er == nil {
-							return b.AuthResultFromToken(ctx, authParams, tr)
+						if silent.RequestType == accesstokens.ATConfidential {
+							if tr, er := b.Token.Credential(ctx, authParams, silent.Credential); er == nil {
+								return b.AuthResultFromToken(ctx, authParams, tr)
+							}
 						}
 					}
 				}
--- a/apps/public/public_test.go
+++ b/apps/public/public_test.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/cache"
+	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/base"
 	internalTime "github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/json/types/time"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/mock"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/fake"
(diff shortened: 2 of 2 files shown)


apps/internal/base/base.go
```diff
@@ -367,8 +367,17 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 					// If the token is not same, we don't need to refresh it.
 					// Which means it refreshed.
 					if str, err := m.Read(ctx, authParams); err == nil && str.AccessToken.Secret == ar.AccessToken {
-						if tr, er := b.Token.Credential(ctx, authParams, silent.Credential); er == nil {
-							return b.AuthResultFromToken(ctx, authParams, tr)
+						switch silent.RequestType {
+						case accesstokens.ATConfidential:
```
bgavrilMS, line 371: There are 3 cases: 

1. PCA, AcquireTokenSilent means refres_token grant. No credential, only RT
2. CCA, web site scenario (user logs in to web site). AcquireTokenSilent - use RT and credential
3. CCA, service principal scenario, - use credential only (AcquireTokenSilent should not used for SP, but I'm not sure if MSAL Go does this)

But your code seems to look at case 1 and 3?

chlowell: 👍 I don't see a test covering case 2 e.g. redeeming a refresh token from OBO or auth code flows

4gust: Added test for the second case.

apps/public/public_test.go
```diff
@@ -1046,3 +1047,56 @@ func getNewClientWithMockedResponses(
...
+	))
+	lmo := "login.microsoftonline.com"
+	originalTime := base.Now
+	defer func() {
+		base.Now = originalTime
+	}()
+	mockClient := mock.NewClient()
+	mockClient.AppendResponse(mock.WithBody(mock.GetTenantDiscoveryBody(lmo, "common")))
+	mockClient.AppendResponse(mock.WithBody(mock.GetAccessTokenBody(accessToken, mock.GetIDToken(homeTenant, fmt.Sprintf(authorityFmt, lmo, homeTenant)), "rt", clientInfo, 36000, 1000)))
+	mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, homeTenant)))
```
chlowell: this doesn't look right. Why are we doing instance discovery after getting a token?

apps/public/public_test.go
```diff
@@ -1046,3 +1047,56 @@ func getNewClientWithMockedResponses(
...
+	mockClient.AppendResponse(mock.WithBody(mock.GetAccessTokenBody("accessToken", mock.GetIDToken(homeTenant, fmt.Sprintf(authorityFmt, lmo, homeTenant)), "rt", clientInfo, 36000, 1000)))
```
chlowell: consider doing something like this so it's obvious this is certain to be a different token
```suggestion
	mockClient.AppendResponse(mock.WithBody(mock.GetAccessTokenBody("not-"+accessToken, mock.GetIDToken(homeTenant, fmt.Sprintf(authorityFmt, lmo, homeTenant)), "rt", clientInfo, 36000, 1000)))
```

apps/confidential/confidential_test.go
```diff
@@ -432,6 +432,78 @@ func TestAcquireTokenByAuthCode(t *testing.T) {
 	}
 }
 
+func TestAcquireTokenByAuthCodeTokenExpiry(t *testing.T) {
```
chlowell, line 435: this test passes without the fix

apps/internal/base/base.go
```diff
@@ -367,8 +367,17 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 					// If the token is not same, we don't need to refresh it.
 					// Which means it refreshed.
 					if str, err := m.Read(ctx, authParams); err == nil && str.AccessToken.Secret == ar.AccessToken {
-						if tr, er := b.Token.Credential(ctx, authParams, silent.Credential); er == nil {
-							return b.AuthResultFromToken(ctx, authParams, tr)
+						switch silent.RequestType {
```
chlowell, line 370: This switch should be exhaustive; it needs either a `default` or a case for `accesstokens.ATUnknown`. At first glance it seems reasonable to error on seeing `accesstokens.ATUnknown` but I don't know whether there's a real scenario for that case

apps/confidential/confidential_test.go
4gust: Yes, because the fix is for the public client. Confidential client works fine.
Wrote this test for perform the **2** case in the list that Bogdan mentioned.

apps/internal/base/base.go
4gust: There shouldn't be the `accesstokens.ATUnknown` case possible as we control which client is calling the `base. Client` 
But, just in case added a error return when its `accesstokens.ATUnknown`.

apps/public/public_test.go
4gust: Actually it does, when we get a token using auth code, and then get a token silent which is expired, it will do the instance discovery again for fetching a new token.
Or did i understand that wrong ?

I updated the client to have instance discovery to false, now it does not calls.

chlowell: I guess this is the bug described by #351

apps/internal/base/base.go
```diff
@@ -367,8 +367,19 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
...
+								return b.AuthResultFromToken(ctx, authParams, tr)
+							}
+						case accesstokens.ATPublic:
+							token, err := b.Token.Refresh(ctx, silent.RequestType, authParams, silent.Credential, storageTokenResponse.RefreshToken)
+							if err != nil {
+								return ar, err
+							}
+							return b.AuthResultFromToken(ctx, authParams, token)
+						case accesstokens.ATUnknown:
+							return ar, errors.New("Silent request type cannot be ATUnknown")
```
chlowell: [staticcheck ST1005](https://staticcheck.dev/docs/checks/#ST1005)
```suggestion
							return ar, errors.New("silent request type cannot be ATUnknown")
```

apps/public/public_test.go
4gust: I understand

PR #584, opened 2025-09-25
Deprecate ROPC flow

Ugonnaak1:
[AB#3385955](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3385955)


apps/public/public.go
```diff
@@ -368,9 +368,8 @@ type AcquireByUsernamePasswordOption interface {
 	acquireByUsernamePasswordOption()
 }
 
+// Deprecated: This API will be removed in a future release. Use a more secure flow instead. Follow this migration guide: https://aka.ms/msal-ropc-migration
```
chlowell, line 371: This comment should be a separate paragraph to render correctly in docs (see [Go documentation](https://go.dev/doc/comment#deprecations))

apps/public/public_test.go
```diff
@@ -116,6 +116,7 @@ func TestAcquireTokenSilentHomeTenantAliases(t *testing.T) {
 }
 
 func TestAcquireTokenSilentWithTenantID(t *testing.T) {
+	t.Skip("Test skipped due to deprecated AcquireTokenByUsernamePassword API usage")
```
chlowell: Don't skip this test--it covers other auth flows and `AcquireTokenByUsernamePassword` is still supported

chlowell (changes requested):
There should be no test changes. Deprecated API is still useable.

PR #585, opened 2025-10-01
feat: add WithPrompt option for AuthCodeURL

szogoon:
This PR adds WithPrompt option, so that URL can be crafted to use prompt parameter.

### DIFF
--- a/apps/public/public.go
+++ b/apps/public/public.go
@@ -149,7 +149,7 @@ func New(clientID string, options ...Option) (Client, error) {
 
 // authCodeURLOptions contains options for AuthCodeURL
 type authCodeURLOptions struct {
-	claims, loginHint, tenantID, domainHint string
+	claims, loginHint, tenantID, domainHint, prompt string
 }
 
 // AuthCodeURLOption is implemented by options for AuthCodeURL
@@ -159,7 +159,7 @@ type AuthCodeURLOption interface {
 
 // AuthCodeURL creates a URL used to acquire an authorization code.
 //
-// Options: [WithClaims], [WithDomainHint], [WithLoginHint], [WithTenantID]
+// Options: [WithClaims], [WithDomainHint], [WithLoginHint], [WithTenantID], [WithPrompt]
 func (pca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string, scopes []string, opts ...AuthCodeURLOption) (string, error) {
 	o := authCodeURLOptions{}
 	if err := options.ApplyOptions(&o, opts); err != nil {
@@ -172,6 +172,7 @@ func (pca Client) AuthCodeURL(ctx context.Context, clientID, redirectURI string,
 	ap.Claims = o.claims
 	ap.LoginHint = o.loginHint
 	ap.DomainHint = o.domainHint
+	ap.Prompt = o.prompt
 	return pca.base.AuthCodeURL(ctx, clientID, redirectURI, scopes, ap)
 }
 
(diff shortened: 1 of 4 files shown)


rayluo (commented):
Yes, it makes sense to add WithPrompt option for PCA's interactive flow and CCA's auth code flow. I'll defer to my other teammates to review the implementation.

apps/public/public.go
```diff
@@ -674,7 +702,7 @@ func (pca Client) AcquireTokenInteractive(ctx context.Context, scopes []string,
 	authParams.LoginHint = o.loginHint
 	authParams.DomainHint = o.domainHint
 	authParams.State = uuid.New().String()
-	authParams.Prompt = "select_account"
+	authParams.Prompt = o.prompt
```
4gust: We’re now changing the default from `"select_account"` to an empty string `""`.
Will need to test this a bit more to ensure everything behaves as expected.

apps/public/public_test.go
```diff
@@ -43,16 +44,16 @@ func fakeBrowserOpenURL(authURL string) error {
...
+	// 	return errors.New("missing query param 'prompt")
+	// }
 	state := q.Get("state")
 	if state == "" {
 		return errors.New("missing query param 'state'")
 	}
 	redirect := q.Get("redirect_uri")
 	if redirect == "" {
-		return errors.New("missing query param 'redirect_uri'")
+		return errors.New(" 'redirect_uri'")
```
4gust: Providing a bit more detail in the error message helps with debugging just saying 'redirect_uri' is too short. Including what exactly is wrong (e.g., missing, invalid, mismatched) would be more helpful.
```suggestion
		return errors.New("missing redirect param 'redirect_uri'")
```
revert to original

apps/public/public_test.go
```diff
@@ -43,16 +44,16 @@ func fakeBrowserOpenURL(authURL string) error {
 	if m := q.Get("code_challenge_method"); m != "S256" {
 		return fmt.Errorf("unexpected code_challenge_method '%s'", m)
 	}
-	if q.Get("prompt") == "" {
-		return errors.New("missing query param 'prompt")
-	}
+	// if q.Get("prompt") == "" {
+	// 	return errors.New("missing query param 'prompt")
```
4gust: Commenting this out will fix the failing test, but as I mentioned earlier, changing the default value from `"select_account"` to `""` might introduce unintended behavior. 
We'll need to test this further, and once confirmed, re-enable it and update the test to assert the correct value based on the expected response.

4gust (changes requested):
This looks good to me. Just please update it with the couple of changes mentioned.

apps/public/public_test.go
szogoon: Maybe I should put "select_account" as a default for AcquireInteractiveOption?

apps/public/public.go
rayluo: I was not aware that the "select_account" was the default value. That seems to be debatable choice in the first place. The default value on the protocol level is empty or completely absent.

In any case, changing the default from "select_account" to empty "" shall not cause catastrophic result because, again that was the default value in the specs and presumably most other libraries do that all the time. The perceived behavior will change, though, as the account selector will not pop up as often; but I think that shall be a welcome change.

4gust: Should this be tested live ?

PrzemyslawMalolepszy-TomTom:
@4gust can I get another round of review?

PR #586, opened 2025-10-03
Adding support for FMI (Federated Managed Identity)

4gust:
### Summary
This PR adds support for **Federated Managed Identity (FMI)** in confidential clients, enabling secure token acquisition for service principals using federated credentials. To support FMI scenarios, we've also added cache extensibility features that allow custom cache key components and extra body parameters.

#### 1. FMI Support for Confidential Clients
Added `WithFMIPath` option for confidential clients to support federated managed identity scenarios:

```go
// Create confidential client with FMI support
cred := confidential.NewCredFromAssertionCallback(
    func(ctx context. Context, aro confidential.AssertionRequestOptions) (string, error) {
        //return GetFederatedToken(ctx) // get federated token
    },
)

app, err := confidential.New(authorityURL, clientID, cred, 
    confidential.WithCache(cacheAccessor))

// Acquire token with FMI path
result, err := app.AcquireTokenByCredential(ctx, scopes, 
    confidential.WithFMIPath("SomeFmiPath/FmiCredentialPath"))

```
#### Cache Key Extensibility
Added `withAdditionalCacheKeyComponents` to support custom cache partitioning, essential for FMI scenarios where tokens need to be isolated by credential path.

#### Extra Body Parameters
Added `withExtraBodyParameters` to send custom parameters in token requests, required for FMI authentication.

Both the method `withAdditionalCacheKeyComponents` and `withExtraBodyParameters` are for internal use only.

### DIFF
--- a/apps/confidential/confidential_test.go
+++ b/apps/confidential/confidential_test.go
@@ -1305,174 +1305,492 @@ func TestWithCache(t *testing.T) {
 	}
 }
 
-func TestWithClaims(t *testing.T) {
-	cred, err := NewCredFromSecret(fakeSecret)
-	if err != nil {
-		t.Fatal(err)
-	}
-	accessToken := "at"
-	lmo, tenant := "login.microsoftonline.com", "tenant"
-	authority := fmt.Sprintf(authorityFmt, lmo, tenant)
-	for _, test := range []struct {
-		capabilities     []string
-		claims, expected string
-	}{
-		{},
+// TestWithClaims1 verifies that claims and client capabilities are correctly
+// merged and sent in token requests across different authentication methods.
+func TestWithClaims1(t *testing.T) {
+	testScenarios := []claimsTestScenario{
 		{
+			name:     "no_claims_or_capabilities",
+			desc:     "Neither claims nor capabilities - no claims parameter should be sent",
+			expected: "",
+		},
+		{
+			name:         "capabilities_only",
+			desc:         "Only client capabilities - should be added to access_token claims",
 			capabilities: []string{"cp1"},
 			expected:     `{"access_token":{"xms_cc":{"values":["cp1"]}}}`,
 		},
 		{
+			name:     "claims_only",
+			desc:     "Only custom claims - should be sent as-is",
 			claims:   `{"id_token":{"auth_time":{"essential":true}}}`,
 			expected: `{"id_token":{"auth_time":{"essential":true}}}`,
 		},
 		{
+			name:         "claims_and_capabilities_merged",
+			desc:         "Both claims and capabilities - should merge capabilities into access_token",
 			capabilities: []string{"cp1", "cp2"},
 			claims:       `{"access_token":{"nbf":{"essential":true, "value":"42"}}}`,
 			expected:     `{"access_token":{"nbf":{"essential":true, "value":"42"}, "xms_cc":{"values":["cp1","cp2"]}}}`,
 		},
-	} {
-		var expected map[string]any
-		if err := json.Unmarshal([]byte(test.expected), &expected); err != nil && test.expected != "" {
-			t.Fatal("test bug: the expected result must be JSON or an empty string")
+	}
+
+	for _, scenario := range testScenarios {
+		t.Run(scenario.name, func(t *testing.T) {
+			// Test each authentication method with this scenario
+			testAuthCodeFlow(t, scenario)
+			testAuthCodeURLFlow(t, scenario)
+			testCredentialFlow(t, scenario)
+			testCredentialFlowWithAccount(t, scenario)
+			testOnBehalfOfFlow(t, scenario)
+		})
+	}
+}
+
+// claimsTestScenario defines a test case for claims and capabilities handling
+type claimsTestScenario struct {
+	name         string
+	desc         string
+	capabilities []string
+	claims       string
+	expected     string // expected JSON claims parameter in request
+}
+
+// claimsValidator validates that claims in HTTP requests match expectations
+type claimsValidator struct {
+	t        *testing.T
+	expected map[string]any
+}
+
+func newClaimsValidator(t *testing.T, expectedJSON string) claimsValidator {
+	t.Helper()
+	var expected map[string]any
+	if expectedJSON != "" {
+		if err := json.Unmarshal([]byte(expectedJSON), &expected); err != nil {
+			t.Fatalf("test bug: expected result must be valid JSON: %v", err)
 		}
-		validate := func(t *testing.T, v url.Values) {
-			if test.expected == "" {
-				if v.Has("claims") {
-					t.Fatal("claims shouldn't be set")
-				}
-				return
-			}
-			claims, ok := v["claims"]
-			if !ok {
-				t.Fatal("claims should be set")
-			}
-			if len(claims) != 1 {
-				t.Fatalf("expected 1 value for claims, got %d", len(claims))
-			}
-			var actual map[string]any
-			if err := json.Unmarshal([]byte(claims[0]), &actual); err != nil {
-				t.Fatal(err)
-			}
-			if diff := pretty.Compare(expected, actual); diff != "" {
-				t.Fatal(diff)
-			}
+	}
+	return claimsValidator{t: t, expected: expected}
+}
+
+func (v claimsValidator) validate(values url.Values) {
+	v.t.Helper()
+
+	if len(v.expected) == 0 {
+		if values.Has("claims") {
+			v.t.Fatal("claims parameter should not be present when no claims are expected")
 		}
-		for _, method := range []string{"authcode", "authcodeURL", "credential", "obo"} {
-			t.Run(method, func(t *testing.T) {
-				mockClient := mock.NewClient()
-				clientInfo, idToken, refreshToken := "", "", ""
-				if method == "obo" {
-					clientInfo = base64.RawStdEncoding.EncodeToString([]byte(`{"uid":"uid","utid":"utid"}`))
-					idToken = mock.GetIDToken(tenant, authority)
-					refreshToken = "rt"
-					// TODO: OBO does instance discovery twice before first token request https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/351
-					mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, tenant)))
-				}
-				if method == "credential" {
-					if test.claims == "" {
-						mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, tenant)))
-					}
-				}
-				mockClient.AppendResponse(mock.WithBody(mock.GetTenantDiscoveryBody(lmo, tenant)))
-				mockClient.AppendResponse(
-					mock.WithBody(mock.GetAccessTokenBody(accessToken, idToken, refreshToken, clientInfo, 3600, 0)),
-					mock.WithCallback(func(r *http.Request) {
-						if err := r.ParseForm(); err != nil {
-							t.Fatal(err)
-						}
-						validate(t, r.Form)
-					}),
-				)
-				if method != "obo" {
-					mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, tenant)))
-				}
-				client, err := New(authority, fakeClientID, cred, WithClientCapabilities(test.capabilities), WithHTTPClient(mockClient))
-				if err != nil {
+		return
+	}
+
+	claims := values["claims"]
+	if len(claims) == 0 {
+		v.t.Fatal("claims parameter should be present")
+	}
+	if len(claims) != 1 {
+		v.t.Fatalf("expected exactly 1 claims value, got %d", len(claims))
+	}
+
+	var actual map[string]any
+	if err := json.Unmarshal([]byte(claims[0]), &actual); err != nil {
+		v.t.Fatalf("failed to parse claims JSON: %v", err)
+	}
+
+	if diff := pretty.Compare(v.expected, actual); diff != "" {
+		v.t.Fatalf("claims mismatch (-expected +actual):\n%s", diff)
+	}
+}
+
+// testAuthCodeFlow tests AcquireTokenByAuthCode with claims
+func testAuthCodeFlow(t *testing.T, scenario claimsTestScenario) {
+	t.Run("auth_code", func(t *testing.T) {
+		const (
+			accessToken = "at"
+			lmo         = "login.microsoftonline.com"
+			tenant      = "tenant"
+		)
+		authority := fmt.Sprintf(authorityFmt, lmo, tenant)
+
+		cred, err := NewCredFromSecret(fakeSecret)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		mockClient := mock.NewClient()
+		validator := newClaimsValidator(t, scenario.expected)
+
+		// Setup mock responses
+		idToken := mock.GetIDToken(tenant, authority)
+		mockClient.AppendResponse(mock.WithBody(mock.GetTenantDiscoveryBody(lmo, tenant)))
+		mockClient.AppendResponse(
+			mock.WithBody(mock.GetAccessTokenBody(accessToken, idToken, "", "", 3600, 0)),
+			mock.WithCallback(func(r *http.Request) {
+				if err := r.ParseForm(); err != nil {
 					t.Fatal(err)
 				}
-				if _, err = client.AcquireTokenSilent(context.Background(), tokenScope, WithSilentAccount(shared.Account{})); err == nil {
-					t.Fatal("silent authentication should fail because the cache is empty")
-				}
-				ctx := context.Background()
-				var ar AuthResult
-				switch method {
-				case "authcode":
-					ar, err = client.AcquireTokenByAuthCode(ctx, "code", localhost, tokenScope, WithClaims(test.claims))
-				case "authcodeURL":
-					u := ""
-					if u, err = client.AuthCodeURL(ctx, fakeClientID, localhost, tokenScope, WithClaims(test.claims)); err == nil {
-						var parsed *url.URL
-						if parsed, err = url.Parse(u); err == nil {
-							validate(t, parsed.Query())
-							return // didn't acquire a token, no need for further validation
-						}
-					}
-				case "credential":
-					ar, err = client.AcquireTokenByCredential(ctx, tokenScope, WithClaims(test.claims))
-				case "obo":
-					ar, err = client.AcquireTokenOnBehalfOf(ctx, "assertion", tokenScope, WithClaims(test.claims))
-				case "password":
-					ar, err = client.AcquireTokenByUsernamePassword(ctx, tokenScope, "username", "password", WithClaims(test.claims))
-				default:
-					t.Fatalf("test bug: no test for %s", method)
-				}
-				if err != nil {
-					t.Fatal(err)
-				}
-				if ar.AccessToken != accessToken {
-					t.Fatalf(`unexpected access token "%s"`, ar.AccessToken)
-				}
-				// silent auth should now succeed, provided no claims are requested, because the client has cached an access token
-				if method == "obo" {
-					ar, err = client.AcquireTokenOnBehalfOf(ctx, "assertion", tokenScope)
-				} else if method == "credential" {
-					ar, err = client.AcquireTokenByCredential(ctx, tokenScope)
-				} else {
-					ar, err = client.acquireTokenSilentInternal(ctx,
-						base.AcquireTokenSilentParameters{Scopes: tokenScope,
-							Account:     shared.Account{},
-							IsAppCache:  true,
-							TenantID:    tenant,
-							AuthnScheme: client.base.AuthParams.AuthnScheme})
-				}
-				if err != nil {
-					t.Fatal(err)
-				}
-				if ar.AccessToken != accessToken {
-					t.Fatalf(`unexpected access token "%s"`, ar.AccessToken)
-				}
-				if test.claims != "" {
-					if _, err = client.AcquireTokenSilent(ctx, tokenScope, WithClaims(test.claims)); err == nil {
-						t.Fatal("AcquireTokenSilent should fail when given claims")
-					}
-					if method == "obo" {
-						// client has cached access and refresh tokens. When given claims, it should redeem a refresh token for a new access token.
-						newToken := "new-access-token"
-						mockClient.AppendResponse(
-							mock.WithBody(mock.GetAccessTokenBody(newToken, idToken, "", clientInfo, 3600, 0)),
-							mock.WithCallback(func(r *http.Request) {
-								if err := r.ParseForm(); err != nil {
-									t.Fatal(err)
-								}
-								// all token requests should include any specified claims
-								validate(t, r.Form)
-								if actual := r.Form.Get("refresh_token"); actual != refreshToken {
-									t.Fatalf(`unexpected refresh token "%s ,, %s"`, actual, refreshToken)
-								}
-							}),
-						)
-						ar, err = client.AcquireTokenOnBehalfOf(ctx, "assertion", tokenScope, WithClaims(test.claims))
-						if err != nil {
-							t.Fatal(err)
-						}
-						if ar.AccessToken != newToken {
-							t.Fatalf(`unexpected access token "%s"`, ar.AccessToken)
-						}
-					}
-				}
-			})
+				validator.validate(r.Form)
+			}),
+		)
+		mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, tenant)))
+
+		client, err := New(authority, fakeClientID, cred,
+			WithClientCapabilities(scenario.capabilities),
+			WithHTTPClient(mockClient))
+		if err != nil {
+			t.Fatal(err)
 		}
+
+		// Verify cache is empty
+		if _, err = client.AcquireTokenSilent(context.Background(), tokenScope,
+			WithSilentAccount(shared.Account{})); err == nil {
+			t.Fatal("silent authentication should fail with empty cache")
+		}
+
+		// Acquire token by auth code
+		ctx := context.Background()
+		ar, err := client.AcquireTokenByAuthCode(ctx, "code", localhost, tokenScope,
+			WithClaims(scenario.claims))
+		if err != nil {
+			t.Fatalf("AcquireTokenByAuthCode failed: %v", err)
+		}
+		if ar.AccessToken != accessToken {
+			t.Fatalf("unexpected access token: got %q, want %q", ar.AccessToken, accessToken)
+		}
+
+		// Verify silent authentication now succeeds (token is cached)
+		// Auth code flow returns an account, so use public API
+		verifySilentAuthSuccess(t, client, accessToken, &ar.Account)
+
+		// Verify that silent auth with claims fails even when providing the account
+		if scenario.claims != "" {
+			if _, err = client.AcquireTokenSilent(ctx, tokenScope,
+				WithSilentAccount(ar.Account),
+				WithClaims(scenario.claims)); err == nil {
+				t.Fatal("AcquireTokenSilent should fail when claims are provided")
+			}
+		}
+	})
+}
+
+// testAuthCodeURLFlow tests AuthCodeURL with claims
+func testAuthCodeURLFlow(t *testing.T, scenario claimsTestScenario) {
+	t.Run("auth_code_url", func(t *testing.T) {
+		const (
+			lmo    = "login.microsoftonline.com"
+			tenant = "tenant"
+		)
+		authority := fmt.Sprintf(authorityFmt, lmo, tenant)
+
+		cred, err := NewCredFromSecret(fakeSecret)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		mockClient := mock.NewClient()
+		validator := newClaimsValidator(t, scenario.expected)
+
+		// Setup mock responses for instance and tenant discovery
+		mockClient.AppendResponse(mock.WithBody(mock.GetTenantDiscoveryBody(lmo, tenant)))
+		mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, tenant)))
+
+		client, err := New(authority, fakeClientID, cred,
+			WithClientCapabilities(scenario.capabilities),
+			WithHTTPClient(mockClient))
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Generate auth code URL with claims
+		ctx := context.Background()
+		authURL, err := client.AuthCodeURL(ctx, fakeClientID, localhost, tokenScope,
+			WithClaims(scenario.claims))
+		if err != nil {
+			t.Fatalf("AuthCodeURL failed: %v", err)
+		}
+
+		// Parse and validate the URL contains expected claims
+		parsed, err := url.Parse(authURL)
+		if err != nil {
+			t.Fatalf("failed to parse auth URL: %v", err)
+		}
+		validator.validate(parsed.Query())
+	})
+}
+
+// testCredentialFlow tests AcquireTokenByCredential with claims
+func testCredentialFlow(t *testing.T, scenario claimsTestScenario) {
+	t.Run("credential", func(t *testing.T) {
+		const (
+			accessToken = "at"
+			lmo         = "login.microsoftonline.com"
+			tenant      = "tenant"
+		)
+		authority := fmt.Sprintf(authorityFmt, lmo, tenant)
+
+		cred, err := NewCredFromSecret(fakeSecret)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		mockClient := mock.NewClient()
+		validator := newClaimsValidator(t, scenario.expected)
+
+		// Setup mock responses
+		// Note: credential flow only does instance discovery when no claims are provided
+		if scenario.claims == "" {
+			mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, tenant)))
+		}
+		mockClient.AppendResponse(mock.WithBody(mock.GetTenantDiscoveryBody(lmo, tenant)))
+		mockClient.AppendResponse(
+			mock.WithBody(mock.GetAccessTokenBody(accessToken, "", "", "", 3600, 0)),
+			mock.WithCallback(func(r *http.Request) {
+				if err := r.ParseForm(); err != nil {
+					t.Fatal(err)
+				}
+				validator.validate(r.Form)
+			}),
+		)
+		mockClient.AppendResponse(mock.WithBody(mock.GetInstanceDiscoveryBody(lmo, tenant)))
+		mockClient.AppendResponse(
+			mock.WithBody(mock.GetAccessTokenBody(accessToken, "", "", "", 3600, 0)),
+			mock.WithCallback(func(r *http.Request) {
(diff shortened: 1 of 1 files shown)


apps/confidential/confidential.go
```diff
@@ -781,3 +787,45 @@ func (cca Client) Account(ctx context.Context, accountID string) (Account, error
...
+		AcquireByCredentialOption
+		AcquireSilentOption
+		options.CallOption
+	}{
+		CallOption: options.NewCallOption(
+			func(a any) error {
+				switch t := a.(type) {
+				case *acquireTokenByCredentialOptions:
+					t.extraBodyParameters = params
+				case *acquireTokenSilentOptions:
```
bgavrilMS: nit: is it difficult to add it to the other confidential client flows? OBO and auth_code?

apps/confidential/confidential.go
```diff
@@ -781,3 +787,45 @@ func (cca Client) Account(ctx context.Context, accountID string) (Account, error
...
+//
+//	params := map[string]func(context.Context) (string, error){
+//	    "custom_param": func(ctx context.Context) (string, error) {
+//	        return "custom_value", nil
+//	    },
+//	}
+//	result, err := client.AcquireTokenByCredential(ctx, scopes, confidential.WithExtraBodyParameters(params))
+func WithExtraBodyParameters(params map[string]func(context.Context) (string, error)) interface {
+	AcquireByCredentialOption
+	AcquireSilentOption
```
bgavrilMS: Will this affect public client? Can we not support this for public client?

apps/confidential/confidential.go
4gust: Not difficult, i can add it.

apps/confidential/confidential.go
4gust: This wont effect public client at all

apps/confidential/confidential_test.go
```diff
@@ -1862,3 +1862,299 @@ func TestAcquireTokenByCredentialFromDSTS(t *testing.T) {
...
+	// This should get token1 from cache
+	result1Again, err := client.AcquireTokenSilent(
+		context.Background(),
+		tokenScope,
+		WithExtraBodyParameters(params1),
+	)
+	if err != nil {
+		t.Fatalf("retrieving cached token1 failed: %v", err)
+	}
+	if result1Again.AccessToken != token1 {
```
bgavrilMS: nit: make use more of the "fromCache" property you exposed a while back.

apps/internal/base/storage/storage.go
```diff
@@ -291,26 +294,92 @@ func (m *Manager) aadMetadata(ctx context.Context, authorityInfo authority.Info)
...
 
-func (m *Manager) readAccessToken(homeID string, envAliases []string, realm, clientID string, scopes []string, tokenType, authnSchemeKeyID string) AccessToken {
+func (m *Manager) readAccessToken(homeID string, envAliases []string, realm, clientID string, scopes []string, tokenType, authnSchemeKeyID, extraBodyParamHashed string) AccessToken {
 	m.contractMu.RLock()
-	// TODO: linear search (over a map no less) is slow for a large number (thousands) of tokens.
-	// this shows up as the dominating node in a profile. for real-world scenarios this likely isn't
-	// an issue, however if it does become a problem then we know where to look.
-	for k, at := range m.contract.AccessTokens {
+
+	scopesStr := strings.Join(scopes, " ")
```
bgavrilMS: Hmm, why did this logic change so much?

apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -466,3 +481,21 @@ func addScopeQueryParam(queryParams url.Values, authParameters authority.AuthPar
 	scopes := AppendDefaultScopes(authParameters)
 	queryParams.Set("scope", strings.Join(scopes, " "))
 }
+
+// addExtraBodyParameters evaluates and adds extra body parameters to the request
+func addExtraBodyParameters(ctx context.Context, v url.Values, ap authority.AuthParams) error {
```
bgavrilMS: I would separate the 2 concepts: 

- extra token payload parameters i.e. extra params in the token cache
- extra body parameters i.e. extra params to the /token endpoint

apps/confidential/confidential_test.go
4gust: Updating it.

apps/internal/base/storage/storage.go
4gust: The implementation was liner before, i changed it.
Later did of checks for matching scopes, hash that we added now and expiry.

apps/confidential/confidential.go
```diff
@@ -781,3 +787,45 @@ func (cca Client) Account(ctx context.Context, accountID string) (Account, error
...
+//
+// Example:
+//
+//	params := map[string]func(context.Context) (string, error){
+//	    "custom_param": func(ctx context.Context) (string, error) {
+//	        return "custom_value", nil
+//	    },
+//	}
+//	result, err := client.AcquireTokenByCredential(ctx, scopes, confidential.WithExtraBodyParameters(params))
+func WithExtraBodyParameters(params map[string]func(context.Context) (string, error)) interface {
```
bgavrilMS: We discussed offline a few changes: 

- make this internal for now
- normal (sync) dictionary ok

apps/internal/base/storage/storage.go
bgavrilMS: We discussed offline about adding optimization as a separate PR.

apps/internal/oauth/ops/accesstokens/accesstokens.go
4gust: Updated and separated them,
Named them similarly to .net counterpart

apps/confidential/confidential.go
4gust: Updated made it internal and now accepts dict.

apps/internal/base/storage/storage.go
4gust: updated, now it only checks if the hash is present in the key as a additional check

apps/confidential/confidential.go
4gust: Added for `AcquireTokenOnBehalfOf` and `AcquireTokenByAuthCode`

apps/confidential/confidential.go
```diff
@@ -685,31 +690,34 @@ func WithChallenge(challenge string) interface {
 // AcquireTokenByAuthCode is a request to acquire a security token from the authority, using an authorization code.
 // The specified redirect URI must be the same URI that was used when the authorization code was requested.
 //
-// Options: [WithChallenge], [WithClaims], [WithTenantID]
+// Options: [WithChallenge], [WithClaims], [WithTenantID], [WithExtraBodyParameters]
```
bgavrilMS: Should this be lowercase w ?

docs/WithExtraBodyParameters.md
```diff
@@ -0,0 +1,356 @@
+# WithExtraBodyParameters API Documentation
+
+## Overview
+
+The `WithExtraBodyParameters` API allows developers to add custom body parameters to token requests in MSAL Go. This feature is particularly useful for scenarios where additional parameters need to be sent to the authorization server, such as custom authentication requirements or special token attributes.
```
bgavrilMS: Not currently public API.

bgavrilMS (approved):

apps/confidential/confidential.go
```diff
@@ -585,7 +587,7 @@ func WithSilentAccount(account Account) interface {
 
 // AcquireTokenSilent acquires a token from either the cache or using a refresh token.
 //
-// Options: [WithClaims], [WithSilentAccount], [WithTenantID]
+// Options: [WithClaims], [WithSilentAccount], [WithTenantID], [WithExtraBodyParameters], [WithAdditionalCacheKeyComponents]
```
chlowell: 1. these doc links don't work because they reference symbols that don't exist (W vs w)
2. you don't want to document these options anyway because (according to the PR description) they're supposed to be private
```suggestion
// Options: [WithClaims], [WithSilentAccount], [WithTenantID]
```

apps/confidential/confidential.go
```diff
@@ -756,18 +771,19 @@ type AcquireOnBehalfOfOption interface {
 // AcquireTokenOnBehalfOf acquires a security token for an app using middle tier apps access token.
 // Refer https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow.
 //
-// Options: [WithClaims], [WithTenantID]
+// Options: [WithClaims], [WithTenantID], [withExtraBodyParameters]
```
chlowell: This symbol isn't exported, so:
1. you can't link to its docs; it has no docs
2. there's no reason to document it anyway; applications can't use it
```suggestion
// Options: [WithClaims], [WithTenantID]
```

apps/confidential/confidential_test.go
```diff
@@ -1862,3 +1862,261 @@ func TestAcquireTokenByCredentialFromDSTS(t *testing.T) {
 		})
 	}
 }
+
+// TestWithExtraBodyParams tests the WithExtraBodyParams option
```
chlowell: Why? Applications can't use this option and it isn't used internally

docs/WithExtraBodyParameters.md
chlowell, line 1: Why document an API applications can't use?

docs/WithExtraBodyParameters.md
```diff
@@ -0,0 +1,356 @@
...
+        "custom_param": func(ctx context.Context) (string, error) {
+            return "custom_value", nil
+        },
+    }
+
+    // Acquire token with extra parameters
+    result, err := app.AcquireTokenByCredential(
+        context.Background(),
+        []string{"https://graph.microsoft.com/.default"},
+        confidential.WithExtraBodyParameters(extraParams),
```
chlowell: This doesn't compile

apps/confidential/confidential.go
```diff
@@ -736,6 +744,12 @@ func (cca Client) AcquireTokenByCredential(ctx context.Context, scopes []string,
 	if o.authnScheme != nil {
 		authParams.AuthnScheme = o.authnScheme
 	}
+	if o.extraBodyParameters != nil {
+		authParams.ExtraBodyParameters = o.extraBodyParameters
+	}
+	if o.cacheKeyComponents != nil {
+		authParams.CacheKeyComponents = o.cacheKeyComponents
+	}
```
chlowell: Nitpicking: you may as well skip the nil checks because
1. these assignments are a no-op when the map in question is zero (nil)
2. zero maps can be used safely: https://go.dev/play/p/A-gui_LvBkV

But in practice these fields will always be nil because an application can't set them

chlowell (changes requested):
This is adding dead code: there's no connection between the public API and the apparent feature code in `/apps/internal`. The description mentions a `WithFMIPath` option; where's that?

apps/confidential/confidential.go
4gust: Updated
Application will use it `withFMIPath`

docs/WithExtraBodyParameters.md
4gust: Updated

docs/WithExtraBodyParameters.md
4gust: Removed, will update the file after we have `withFMIPath` finalized.

apps/confidential/confidential_test.go
4gust: We have planned to keep this, and we might make it public.Its still under discussion

apps/confidential/confidential.go
4gust: removed

apps/internal/base/base.go
```diff
@@ -46,39 +46,43 @@ type accountManager interface {
...
+	Account             shared.Account
+	RequestType         accesstokens.AppType
+	Credential          *accesstokens.Credential
+	IsAppCache          bool
+	TenantID            string
+	UserAssertion       string
+	AuthorizationType   authority.AuthorizeType
+	Claims              string
+	AuthnScheme         authority.AuthenticationScheme
+	ExtraBodyParameters map[string]string
```
bgavrilMS, line 59: Can we keep this internal for now / outside the public API?

apps/confidential/confidential.go
```diff
@@ -729,7 +739,7 @@ type AcquireByCredentialOption interface {
 
 // AcquireTokenByCredential acquires a security token from the authority, using the client credentials grant.
 //
-// Options: [WithClaims], [WithTenantID]
+// Options: [WithClaims], [WithTenantID], [WithFMIPath]]
```
chlowell:
```suggestion
// Options: [WithClaims], [WithTenantID], [WithFMIPath]
```

apps/confidential/confidential.go
```diff
@@ -799,3 +826,105 @@ func (cca Client) Account(ctx context.Context, accountID string) (Account, error
...
+				default:
+					return fmt.Errorf("unexpected options type %T", a)
+				}
+				return nil
+			},
+		),
+	}
+}
+
+func WithFMIPath(path string) interface {
```
chlowell, line 825: Missing docs

apps/confidential/confidential.go
```diff
@@ -799,3 +826,105 @@ func (cca Client) Account(ctx context.Context, accountID string) (Account, error
...
+// WithExtraBodyParameters adds extra body parameters to the token request.
+// These parameters are added to the request body
+//
+// Example:
+//
+//	params := map[string]string{
+//	    "custom_param": "custom_value",
+//	}
+//	result, err := client.AcquireTokenByCredential(ctx, scopes, confidential.WithExtraBodyParameters(params))
+func withExtraBodyParameters(params map[string]string) interface {
```
chlowell: This and `withAdditionalCacheKeyComponents` are dead code--they're used only in tests. I suggest not adding them until you're sure you want to include them in the public API.

docs/federated_managed_identity.md
```diff
@@ -0,0 +1,133 @@
...
+- FMI tokens don't interfere with regular (non-FMI) tokens
+- Silent token acquisition works seamlessly with cached FMI tokens
+
+```go
+// First call - acquires from identity provider
+result, err := app.AcquireTokenByCredential(ctx, scopes,
+    confidential.WithFMIPath("YourFmiPath/CredentialPath"))
+
+// Second call - retrieves from cache automatically
+cachedResult, err := app.AcquireTokenSilent(ctx, scopes)
```
chlowell: will this ever succeed i.e. return a nil error?

docs/federated_managed_identity.md
```diff
@@ -0,0 +1,133 @@
...
+    )
+    if err != nil {
+        return "", err
+    }
+
+    return result.AccessToken, nil
+}
+```
+
+### Step 2: Use FMI with Assertion Callback
```
chlowell: "step 2" implies this code should run after "step 1" but in fact they're alternatives

docs/federated_managed_identity.md
```diff
@@ -0,0 +1,133 @@
...
+1. **Assertion Callback**: Your callback function retrieves an FMI credential from RMA
+2. **FMI Path**: The `WithFMIPath` option identifies the specific federated credential
+3. **Automatic Isolation**: MSAL handles cache isolation transparently based on the FMI path
+4. **Token Retrieval**: Silent authentication automatically finds the correct cached token
+
+## Important Notes
+
+- Use the special client ID `urn:microsoft:identity:fmi` for FMI scenarios
+- The FMI credential is typically obtained from an RMA (Resource Management API) service  
+- Cache isolation is handled automatically - no manual cache management needed
```
chlowell: consider moving this content to the overview above the example code

apps/tests/integration/fmi_integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+		t.Fatalf("TestFMIWithMultipleApps: AcquireTokenSilent() for app2 failed: %s", errors.Verbose(err))
+	}
+	if result2.AccessToken == "" {
+		t.Fatal("TestFMIWithMultipleApps: AcquireTokenSilent() for app2 returned empty AccessToken")
+	}
+
+	t.Log("TestFMIWithMultipleApps: Successfully acquired token for second app using FMI")
+}
+
+// TestFMIIntegration tests the Family Multiple Identity functionality
```
chlowell: I guess you want
```suggestion
// TestFMIIntegration tests the Federated Managed Identity functionality
```

apps/tests/integration/fmi_integration_test.go
```diff
@@ -0,0 +1,256 @@
+// FMITestSuite contains a set of tests for Family of Multiple Identities functionality
```
chlowell: Missing license header, and consider deleting this comment because it seems to describe a different feature and FMITestSuite doesn't exist

apps/tests/integration/fmi_integration_test.go
chlowell: Should this be
```suggestion
// TestFMIIntegration tests the Federated Managed Identity functionality
```

apps/internal/base/base.go
```diff
@@ -473,7 +490,7 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 	}
 	ar, err := NewAuthResult(token, account)
 	if err == nil && b.cacheAccessor != nil {
-		err = b.cacheAccessor.Export(ctx, b.manager, cache.ExportHints{PartitionKey: key})
+		err = b.cacheAccessor.Export(ctx, m, cache.ExportHints{PartitionKey: key})
```
chlowell: What motivated this? It looks correct but I wonder whether it could introduce a new way to trigger #455

docs/federated_managed_identity.md
```diff
@@ -0,0 +1,133 @@
...
+
+- **Assertion-Based Authentication**: Uses callback functions to dynamically retrieve FMI credentials
+- **Cache Isolation**: Tokens acquired with different FMI paths are cached separately
+- **FMI Path Support**: The `WithFMIPath` option identifies the specific federated credential path
+
+## Basic Usage
+
+### Step 1: Create an FMI Credential Provider
+
+```go
```
chlowell, line 18: These code samples would be better as [testable examples](https://go.dev/blog/examples). Those appear in [the package docs](https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-library-for-go/apps/confidential#pkg-examples) (unlike this document) and are easier to maintain because `go test` compiles them

apps/tests/integration/fmi_integration_test.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	// Validate that we got the same token (proving cache was used)
+	if result.AccessToken != silentResult.AccessToken {
+		t.Fatalf("TestFMIBasicFunctionality: token comparison failed - tokens don't match, cache might not be working correctly")
+	}
+}
+
+// TestFMIWithMultipleApps tests sharing tokens between multiple applications
+// that belong to the same family using the FMI feature
```
chlowell: family?

apps/internal/base/storage/storage.go
```diff
@@ -291,26 +293,50 @@ func (m *Manager) aadMetadata(ctx context.Context, authorityInfo authority.Info)
...
-func (m *Manager) readAccessToken(homeID string, envAliases []string, realm, clientID string, scopes []string, tokenType, authnSchemeKeyID string) AccessToken {
+func (m *Manager) readAccessToken(homeID string, envAliases []string, realm, clientID string, scopes []string, tokenType, authnSchemeKeyID, extCacheKey string) AccessToken {
 	m.contractMu.RLock()
-	// TODO: linear search (over a map no less) is slow for a large number (thousands) of tokens.
-	// this shows up as the dominating node in a profile. for real-world scenarios this likely isn't
-	// an issue, however if it does become a problem then we know where to look.
-	for k, at := range m.contract.AccessTokens {
+
+	// Choose the appropriate map based on whether we have a hash
+	tokensToSearch := m.contract.AccessTokens
```
chlowell: Is it the comment or the code here that's incorrect?

4gust: There was some more code here, i missed to remove the comment

apps/internal/base/base.go
4gust: I missed that, reverted the change.

docs/federated_managed_identity.md
4gust: Added example

apps/tests/integration/fmi_integration_test.go
4gust: Updated.

apps/tests/integration/fmi_integration_test.go
4gust: Added license header

docs/federated_managed_identity.md
4gust: No, it wont. 
I updated the test it.

apps/confidential/confidential.go
4gust: Removed both the method. Initially it was planned that we will use both separate, but then now we will use `WithFMIPath` which will do things it needs.

apps/internal/base/base.go
4gust: As base is internal it will only show up in our code.
and they need to be public for our confidential client.

apps/confidential/confidential.go
```diff
@@ -695,31 +702,34 @@ func WithChallenge(challenge string) interface {
 // AcquireTokenByAuthCode is a request to acquire a security token from the authority, using an authorization code.
 // The specified redirect URI must be the same URI that was used when the authorization code was requested.
 //
-// Options: [WithChallenge], [WithClaims], [WithTenantID]
+// Options: [WithChallenge], [WithClaims], [WithTenantID]]
```
chlowell:
```suggestion
// Options: [WithChallenge], [WithClaims], [WithTenantID]
```

apps/confidential/confidential.go
```diff
@@ -763,7 +788,8 @@ func (cca Client) AcquireTokenByCredential(ctx context.Context, scopes []string,
 
 // acquireTokenOnBehalfOfOptions contains optional configuration for AcquireTokenOnBehalfOf
 type acquireTokenOnBehalfOfOptions struct {
-	claims, tenantID string
+	claims, tenantID    string
+	extraBodyParameters map[string]string
```
chlowell: Should external callers be able to set this?

apps/confidential/confidential_test.go
```diff
@@ -1428,7 +1428,12 @@ func TestWithClaims(t *testing.T) {
 				} else if method == "credential" {
 					ar, err = client.AcquireTokenByCredential(ctx, tokenScope)
 				} else {
-					ar, err = client.acquireTokenSilentInternal(ctx, tokenScope, shared.Account{}, "", tenant, client.base.AuthParams.AuthnScheme)
+					ar, err = client.acquireTokenSilentInternal(ctx,
```
chlowell: Tests should prefer calling public methods

apps/confidential/confidential.go
```diff
@@ -662,6 +668,7 @@ func (cca Client) AcquireTokenByUsernamePassword(ctx context.Context, scopes []s
 // acquireTokenByAuthCodeOptions contains the optional parameters used to acquire an access token using the authorization code flow.
 type acquireTokenByAuthCodeOptions struct {
 	challenge, claims, tenantID string
+	extraBodyParameters         map[string]string
```
chlowell: Should external callers be able to set this?

apps/confidential/confidential_test.go
4gust: Actually this is a special case, because we now have added that silent calls will return error on service principal calls .
So, we call this to just this test which tests claims

apps/confidential/confidential.go
4gust: No, they should not be, only though methods like `WithFMIPath`.

apps/confidential/confidential.go
4gust: Will remove this, this needs to be added in another PR where we enable/add the `WithExtraBodyParameters`

apps/confidential/confidential.go
```diff
@@ -601,21 +603,25 @@ func (cca Client) AcquireTokenSilent(ctx context.Context, scopes []string, opts
...
+		Scopes:              scopes,
+		Account:             o.account,
+		RequestType:         accesstokens.ATConfidential,
+		Credential:          cca.cred,
+		IsAppCache:          o.account.IsZero(),
+		TenantID:            o.tenantID,
+		AuthnScheme:         o.authnScheme,
+		Claims:              o.claims,
+		ExtraBodyParameters: o.extraBodyParameters,
+		CacheKeyComponents:  o.cacheKeyComponents,
```
chlowell: Should callers be able to set these fields (they can't)?

apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -466,3 +481,17 @@ func addScopeQueryParam(queryParams url.Values, authParameters authority.AuthPar
...
+	if len(ap.ExtraBodyParameters) == 0 {
+		return nil
+	}
+
```
chlowell: it's safe to range over an empty map
```suggestion
```

apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -466,3 +481,17 @@ func addScopeQueryParam(queryParams url.Values, authParameters authority.AuthPar
 	scopes := AppendDefaultScopes(authParameters)
 	queryParams.Set("scope", strings.Join(scopes, " "))
 }
+
+// addExtraBodyParameters evaluates and adds extra body parameters to the request
+func addExtraBodyParameters(ctx context.Context, v url.Values, ap authority.AuthParams) error {
```
chlowell: you can spare callers from writing some boilerplate because this always returns nil
```suggestion
func addExtraBodyParameters(ctx context.Context, v url.Values, ap authority.AuthParams) {
```

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -642,8 +649,47 @@ func (a *AuthParams) AssertionHash() string {
...
+	baseKey := ""
 	if a.AuthorityInfo.Tenant != "" {
-		return fmt.Sprintf("%s_%s_AppTokenCache", a.ClientID, a.AuthorityInfo.Tenant)
+		baseKey = fmt.Sprintf("%s_%s", a.ClientID, a.AuthorityInfo.Tenant)
+	} else {
+		baseKey = fmt.Sprintf("%s_", a.ClientID)
+	}
+
+	// Include extra body parameters in the cache key
+	if len(a.CacheKeyComponents) > 0 {
```
chlowell: you can call `CacheExtKeyGenerator()` unconditionally because it does the same check

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -642,8 +649,47 @@ func (a *AuthParams) AssertionHash() string {
 }
 
 func (a *AuthParams) AppKey() string {
+	baseKey := ""
 	if a.AuthorityInfo.Tenant != "" {
-		return fmt.Sprintf("%s_%s_AppTokenCache", a.ClientID, a.AuthorityInfo.Tenant)
+		baseKey = fmt.Sprintf("%s_%s", a.ClientID, a.AuthorityInfo.Tenant)
+	} else {
+		baseKey = fmt.Sprintf("%s_", a.ClientID)
+	}
```
chlowell, line 655: simpler and more efficient:
```go
	baseKey := a.ClientID + "_"
	if a.AuthorityInfo.Tenant != "" {
		baseKey += a.AuthorityInfo.Tenant
	}
```

apps/internal/oauth/ops/authority/authority_ext_cachekey_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	hash1 := ap1.CacheExtKeyGenerator()
+	hash2 := ap2.CacheExtKeyGenerator()
+
+	if hash1 == hash2 {
+		t.Errorf("expected different hashes for different parameter keys, both got %q", hash1)
+	}
+}
+
+func TestAppKeyWithCacheKeyComponent(t *testing.T) {
```
chlowell, line 104: this test will pass even if the implementation doesn't sort the keys

apps/confidential/confidential.go
4gust: Not yet, we added the functions earlier and then removed, we dont have them to be public yet.

apps/internal/oauth/ops/authority/authority_ext_cachekey_test.go
4gust: Will, `for k := range a.CacheKeyComponents` in go always return the same order ?
If no then how can we test for that random order ?

apps/confidential/confidential.go
chlowell: Then you should remove these fields (from `acquireTokenSilentOptions`) because they have no function

apps/confidential/confidential.go
```diff
@@ -585,7 +587,7 @@ func WithSilentAccount(account Account) interface {
 
 // AcquireTokenSilent acquires a token from either the cache or using a refresh token.
 //
-// Options: [WithClaims], [WithSilentAccount], [WithTenantID]
+// Options: [WithClaims], [WithSilentAccount], [WithTenantID], [WithFMIPath]
```
chlowell:
```suggestion
// Options: [WithClaims], [WithSilentAccount], [WithTenantID]
```

apps/internal/oauth/ops/authority/authority_ext_cachekey_test.go
chlowell: No, map iteration order is undefined. You can only test probabilistically whether the AppKey method returns a stable value. Try adding a few more pairs to the test `params` and a loop around AppKey calls

apps/confidential/confidential.go
4gust: Removed.

apps/internal/oauth/ops/authority/authority_ext_cachekey_test.go
4gust: Added one more test to have couple of more keys

apps/confidential/confidential.go
```diff
@@ -799,3 +818,72 @@ func (cca Client) Account(ctx context.Context, accountID string) (Account, error
...
 	return cca.base.RemoveAccount(ctx, account)
 }
+
+// WithFMIPath specifies the path to a federated managed identity.
+// The path should point to a valid FMI configuration file that contains the necessary
+// identity information for authentication.
+//
+// Example:
+//
+//	result, err := client.AcquireTokenByCredential(ctx, scopes, confidential.WithFMIPath("fmi/path"))
```
chlowell: suggest removing this example because the compiler won't check it and it's trivial (the hard thing here is knowing when to use this option and how to get a valid FMI configuration file, not calling the function)

apps/confidential/confidential.go
```diff
@@ -799,3 +818,72 @@ func (cca Client) Account(ctx context.Context, accountID string) (Account, error
...
+}
+
+// WithAttribute provies the attribute which is passed on as "xmc_attr" in the token request body.
+// This is used in conjunction with federated managed identities to specify the identity attribute
+//
+// Example:
+//
+// Using raw string literal (backticks) for easier formatting
+// attrs := `{"FavoriteColor": "Blue", "file:/c/users/foobar/documents/info.txt": "{\"permissions\":[\"read\",\"write\"]}"}`
+// result, err := client.AcquireTokenByCredential(ctx, scopes, confidential.WithAttribute(attrs))
```
chlowell: Is this a real scenario or should callers always pass `WithAttribute` alongside `WithFMIPath`?

apps/confidential/confidential_test.go
chlowell: Sorry, I don't understand. If `AcquireTokenByCredential` now does silent auth, why can't tests call that function instead of `acquireTokenSilentInternal`?

4gust: `AcquireTokenByCredential` now does silent calls, which is only for service principal.
every other API which takes account will be able to call `AcquireTokenSilent`.
In this test the account is passed with empty object with `AcquireTokenByAuthCode` and `AcquireTokenOnBehalfOf` which fails this test, by passing this interface will have no impact on the implementation.

apps/confidential/confidential.go
4gust: Yes, this is a valid scenario.

apps/confidential/confidential.go
4gust: Makes sense, removed this.

apps/confidential/confidential.go
```diff
@@ -799,3 +818,62 @@ func (cca Client) Account(ctx context.Context, accountID string) (Account, error
...
+					return fmt.Errorf("unexpected options type %T", a)
+				}
+				return nil
+			},
+		),
+	}
+}
+
+// WithAttribute provies the attribute which is passed on as "xmc_attr" in the token request body.
+// This is used in conjunction with federated managed identities to specify the identity attribute
```
chlowell: This implies `WithAttribute` isn't useable without `WithFMIPath`

apps/internal/oauth/ops/accesstokens/extrabodyparams_test.go
chlowell, line 1: Conventionally, these tests would be defined in `accesstokens_test.go`. Is there any particular reason to create a new file for them?

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -642,8 +649,44 @@ func (a *AuthParams) AssertionHash() string {
...
+		return ""
+	}
+
+	// Sort keys to ensure consistent hashing
+	keys := make([]string, 0, len(a.CacheKeyComponents))
+	for k := range a.CacheKeyComponents {
+		keys = append(keys, k)
+	}
+
+	// Sort the keys
```
chlowell: nitpicking: the comment isn't helpful because what the code does is obvious

apps/confidential/confidential.go
```diff
@@ -799,3 +818,62 @@ func (cca Client) Account(ctx context.Context, accountID string) (Account, error
...
+				default:
+					return fmt.Errorf("unexpected options type %T", a)
+				}
+				return nil
+			},
+		),
+	}
+}
+
+// WithAttribute provies the attribute which is passed on as "xmc_attr" in the token request body.
```
chlowell:
```suggestion
// WithAttribute provides the attribute which is passed on as "xmc_attr" in the token request body.
```

apps/confidential/confidential_test.go
chlowell: I'm sorry if I confused the issue by suggesting `AcquireTokenCredential` in this case. That isn't the right method here because if we reach this line, the client authenticated with a password. So, why can't this test call `AcquireTokenSilent` here? That's what an external caller would do and it should work, no?

apps/confidential/confidential.go
4gust: Updated comment to now correctly reflect what it does.

apps/internal/oauth/ops/accesstokens/extrabodyparams_test.go
4gust: Separate test files improve maintainability and follow the .NET library's structure.
Also, test files are too long so, decided to make a new file for specific feature test.

apps/confidential/confidential_test.go
4gust: let me rewrite this whole test, there are few things wrong her, It does not response with account.

4gust: Updated, made some changes to run individual cases for all scenarios

PR #598, opened 2026-02-28
Replace godebug with go-cmp

jfantinhardesty:
This PR refactors the codebase to replace the use of the `github.com/kylelemons/godebug` package with `github.com/google/go-cmp` package. Godebug is no longer maintained and has not been updated in 6 years and go-cmp is actively maintained. This mainly just replaces `pretty.Compare` with `cmp.Diff`. And the errors.Verbose() function now uses the httputil library to dump the requests and responses.


apps/errors/errors.go
```diff
@@ -82,7 +65,37 @@ func (e InvalidJsonErr) Error() string {
 func (e CallErr) Verbose() string {
 	e.Resp.Request = nil // This brings in a bunch of TLS crap we don't need
 	e.Resp.TLS = nil     // Same
-	return fmt.Sprintf("%s:\nRequest:\n%s\nResponse:\n%s", e.Err, prettyConf.Sprint(e.Req), prettyConf.Sprint(e.Resp))
+	return fmt.Sprintf("%s:\nRequest:\n%s\nResponse:\n%s", e.Err, dumpRequest(e.Req), dumpResponse(e.Resp))
```
bgavrilMS, line 66: This part needs some cleanup logic. Something like: 

For dumpRequest:

- Log the URL host and path only (no query string values)
- Log HTTP method
- Log header names only (not values — Authorization header leaks tokens)
- Log GET/POST parameter names only (not values — body leaks secrets/assertions)

For dumpResponse:

- On error responses (non-2xx): log field names + values (safe, it's just an error description)
- On success responses: log field names only, never values

apps/internal/base/storage/items_test.go
```diff
@@ -13,7 +13,7 @@ import (
 	internalTime "github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/json/types/time"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/oauth/ops/accesstokens"
 	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/internal/shared"
-	"github.com/kylelemons/godebug/pretty"
+	"github.com/google/go-cmp/cmp"
```
bgavrilMS, line 16: @chlowell - would appreciate your review on this PR.

apps/errors/errors.go
jfantinhardesty: I made the output more similar to how it worked before with godebug/pretty. I do believe the prior version with godebug/pretty could leak header tokens and secrets since there was no processing of those values, but that may have been intentional since it is a verbose log. I made it a bit more structured so it should be easier to decide what to print. Maybe you could help with that as that seems like there are a few edge cases and more like a separate PR, since the current code can leak these tokens.

apps/errors/errors.go
```diff
@@ -82,7 +65,80 @@ func (e InvalidJsonErr) Error() string {
...
+func dumpRequest(req *http.Request) string {
+	if req == nil {
+		return "nil"
+	}
+	var b strings.Builder
+	fmt.Fprintf(&b, "{Method:     %q,\n", req.Method)
+	fmt.Fprintf(&b, " URL:        {Scheme:   %q,\n", req.URL.Scheme)
+	fmt.Fprintf(&b, "              Host:     %q,\n", req.URL.Host)
+	fmt.Fprintf(&b, "              Path:     %q,\n", req.URL.Path)
+	fmt.Fprintf(&b, "              RawQuery: %q},\n", req.URL.RawQuery)
```
bgavrilMS: The rawQuery should not be printed. It may contain login hint, i.e. user emails.

apps/errors/errors.go
```diff
@@ -82,7 +65,80 @@ func (e InvalidJsonErr) Error() string {
...
+	fmt.Fprintf(&b, " Proto:      %q,\n", req.Proto)
+	fmt.Fprintf(&b, " ProtoMajor: %d,\n", req.ProtoMajor)
```
bgavrilMS: nit: Isn't `Proto` enough?

apps/errors/errors.go
```diff
@@ -82,7 +65,80 @@ func (e InvalidJsonErr) Error() string {
...
+	if bodyStr == "" {
+		fmt.Fprintf(&b, " Body:          {},\n")
+	} else {
+		fmt.Fprintf(&b, " Body:          %q,\n", bodyStr)
+	}
+	fmt.Fprintf(&b, " ContentLength: %d}", resp.ContentLength)
+	return b.String()
+}
+
+func formatHeaders(header http.Header, indent string) string {
```
bgavrilMS: We should not print header values here. It's ok to not print headers at all. We can decide to print specific headers in other layers of the library.

apps/errors/errors.go
```diff
@@ -82,7 +65,80 @@ func (e InvalidJsonErr) Error() string {
...
+		}
+		resp.Body = io.NopCloser(strings.NewReader(""))
+	}
+	var b strings.Builder
+	fmt.Fprintf(&b, "{Status:        %q,\n", resp.Status)
+	fmt.Fprintf(&b, " StatusCode:    %d,\n", resp.StatusCode)
+	fmt.Fprintf(&b, " Proto:         %q,\n", resp.Proto)
+	fmt.Fprintf(&b, " ProtoMajor:    %d,\n", resp.ProtoMajor)
+	fmt.Fprintf(&b, " ProtoMinor:    %d,\n", resp.ProtoMinor)
+	fmt.Fprintf(&b, " Header:        %s,\n", formatHeaders(resp.Header, "                 "))
```
bgavrilMS: As above, let's not print headers.

apps/errors/errors.go
```diff
@@ -82,7 +65,80 @@ func (e InvalidJsonErr) Error() string {
...
+	if bodyStr == "" {
```
bgavrilMS: body can contain tokens, better not print it at all.

apps/errors/errors.go
bgavrilMS: Agreed, let's keep it simple. Print minimal information only. From what I can see, app owner can choose to print full HTTP details on their side anyway.

apps/errors/errors.go
jfantinhardesty: Yes! godebug/pretty would print it, so I added that to the prints. But now it will just print Proto.

apps/errors/errors.go
jfantinhardesty: Removed rawQuery.

apps/errors/errors.go
jfantinhardesty: Removed printing headers.

apps/errors/errors.go
jfantinhardesty: Removed printing body.

go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/AzureAD/microsoft-authentication-library-for-go
 
-go 1.18
+go 1.21
```
chlowell, line 3: I doubt this will cause problems because Google stopped supporting 1.20 a couple years ago and upgrading to 1.21 should be easy, however I think it's important to note this is raising the library's minimum required version to enable a test change having no functional impact on the tests or customer applications. Telemetry, if there is any, would help evaluate this

apps/errors/errors.go
```diff
@@ -82,7 +63,34 @@ func (e InvalidJsonErr) Error() string {
...
+
+func dumpRequest(req *http.Request) string {
+	if req == nil {
+		return "nil"
+	}
+	var b strings.Builder
+	fmt.Fprintf(&b, "{Method:     %q,\n", req.Method)
+	fmt.Fprintf(&b, " URL:        {Scheme:   %q,\n", req.URL.Scheme)
+	fmt.Fprintf(&b, "              Host:     %q,\n", req.URL.Host)
+	fmt.Fprintf(&b, "              Path:     %q},\n", req.URL.Path)
```
chlowell, line 77: We should guard against `req.URL == nil`, if only to satisfy static analysis and LLMs

chlowell (changes requested):
Looks good overall, the blocker for me is the nil URL guard. Version upgrade policy is up to @bgavrilMS

go.mod
bgavrilMS: We don't have telemetry on Go.

4gust (approved):

PR #614, opened 2026-04-21
 Recover from panics in `json.Unmarshal` to prevent token-cache crashes (fixes #579)

### LINKED ISSUE #579: [Bug] Panic with "reflect: New of type that may not be allocated in heap (possibly undefined cgo C type)"
1.4.2

**Where is the issue?**
* Public client
    * [ ] Device code flow
    * [ ] Authorization code flow 
* Confidential client
    * [ ] Authorization code flow 
    * [ ] Client credentials:
        * [ ] client secret
        * [ ] client certificate
* Token cache serialization
     * [ ] In-memory cache
* Other (please describe)

1.24.1

Windows 11 AMD64

**Repro**

I don't know the steps to reproduce - this was captured by sentry.  I could guess that the input data was malformed somehow, but looking at the backtrace can probably provide a clue as to how to reproduce.

**Expected behavior**
The library code should not panic when unmarshalling.  Unmarshall returns an error so I expect it to error if there are issues with the input data.  The struct it is unmarshalling to is defined by the library, so if that is at fault it should be fixed.

**Actual behavior**
Panic:

reflect: New of type that may not be allocated in heap (possibly undefined cgo C type)

reflect/value.go in New at line 3019
github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/json/mapslice.go in newValue at line 332
github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/json/mapslice.go in (*mapWalk).storeStruct at line 123
github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/json/mapslice.go in (*mapWalk).run at line 37
github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/json/mapslice.go in unmarshalMap at line 19
github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/json/struct.go in (*decoder).storeValue at line 151
github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/json/struct.go in (*decoder).run at line 57
github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/json/struct.go in unmarshalStruct at line 38
github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/json/json.go in Unmarshal at line 68
github.com/AzureAD/microsoft-authentication-library-for-go@v1.4.2/apps/internal/base/storage/storage.go in (*Manager).Unmarshal at line 581

**Possible solution**

**Additional context / logs / screenshots**
Add any other context about the problem here, such as logs and screenshots.

4gust:
## Summary
 
 Fixes #579.
 
 `apps/internal/json.Unmarshal` could panic with `reflect: New of type that may
 not be allocated in heap (possibly undefined cgo C type)` (and similar
 runtime-level panics) when processing certain inputs. The panic originates
 deep in the custom reflect-based decoder
 (`mapWalk.storeStruct → newValue → reflect.New`) and escapes unhandled,
 crashing the caller — including any application using the in-memory token
 cache, since `storage.Manager.Unmarshal` is the entry point on the cache
 deserialization path.
 
 ## Fix
 
 `apps/internal/json/json.go`
 
 - Add a deferred `recover()` to `Unmarshal` (using a named return) that
   converts any escaping panic into an error of the form
   `json: panic during Unmarshal: <value>`. This restores the documented
   contract: callers get an error, not a process-level crash.
 - Update the doc comment to describe the new recover behavior and warn that
   on error the destination `i` may be partially populated, since the decoder
   writes fields sequentially. MSAL's own `storage.Manager.Unmarshal` already
   follows the safe decode-into-temporary-then-swap pattern, so the in-memory
   cache cannot end up half-populated from a panicking input.
 
 ## Tests
 
 `apps/internal/json/json_test.go`
 
 Added regression coverage that locks in `recover()` behavior on every
 reflect-bearing code path in the json package, so a future change cannot
 silently let a panic escape `Unmarshal` again.
 
 | Test | Code path exercised | Panic source |
 |---|---|---|
 | `TestUnmarshalPanicRecovery` | `mapWalk.storeStruct` (custom Unmarshaler) | `panic(string)` |
 | `TestUnmarshalRuntimePanicRecovery/map_of_struct` | `mapWalk.storeStruct` — **exact #579 stack frames** | `runtime.Error` (nil-pointer dereference) |
 | `TestUnmarshalRuntimePanicRecovery/slice_of_struct` | `sliceWalk.storeStruct` | `runtime.Error` |
 | `TestUnmarshalRuntimePanicRecovery/nested_struct_field` | `decoder.storeValue` (struct branch) | `runtime.Error` |
 
 A true `notinheap` `reflect.New` panic cannot be constructed from user code
 (the flag is only set on internal runtime types), so the regression tests
 trigger a `runtime.Error` (nil-pointer dereference) from a custom
 `UnmarshalJSON` — the same panic *category* as the original failure. With
 the `recover()` removed, the map-of-struct test fails with a stack trace
 identical to the one reported in #579, confirming the test exercises the
 real failure path.
 
 ## Verification
 
 - `go build ./apps/...` — clean
 - `go test ./apps/internal/...` — all packages pass
 - New tests pass with the fix; fail with the expected `runtime.Error` stack
   trace when the `recover()` is removed.

### DIFF
--- a/apps/internal/json/json_test.go
+++ b/apps/internal/json/json_test.go
@@ -6,6 +6,7 @@ package json
 import (
 	"encoding/json"
 	"fmt"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


bgavrilMS (approved):

apps/internal/json/json_test.go
```diff
@@ -202,3 +203,111 @@ func TestDelimIs(t *testing.T) {
...
+			name:   "slice of struct",
+			input:  `{"Inner":[{}]}`,
+			target: &runtimePanicSliceHolder{},
+		},
+		{
+			name:   "nested struct field",
+			input:  `{"Inner":{}}`,
+			target: &runtimePanicStructHolder{},
+		},
+	}
```
Robbie-Microsoft, line 300: Nit: the subtest closure captures the loop variable by reference. It's not a live bug here since the subtests aren't parallel (no `t.Parallel()`), but Go's `loopclosure` vet check flags this pattern, and it would silently break if `t.Parallel()` is added later. Easy fix — rebind before the `t.Run` call:

```go
for _, test := range tests {
    test := test // capture loop variable for subtest closure
    t.Run(test.name, func(t *testing.T) {
```

Robbie-Microsoft (approved):

Issue #613, opened 2026-04-20
[Bug] Remove or fix gzip decompression


bgavrilMS:
MSAL-Go's internal HTTP client decompresses gzip-encoded responses without any size limit. 

This exists in 

```go
func gzipDecompress(r io.Reader) io.Reader {
gzipReader, _ := gzip.NewReader(r)
pipeOut, pipeIn := io.Pipe()
go func() {
// decompression bomb would have to come from Azure services.
// If we want to limit, we should do that in comm.do().
_, err := io.Copy(pipeIn, gzipReader) // line 18 - UNBOUNDED
...

A simple fix would to limit the reader, e.g.

```go
// comm.go — readBody()
const maxResponseBody = 32 << 20 // 32 MiB

func (c *Client) readBody(resp *http.Response) ([]byte, error) {
    ....
    return io.ReadAll(io.LimitReader(reader, maxResponseBody))
}

```

Possible solutions: 

- other MSALs do not have this logic - should we just remove it? Check the other MSAL HttpClient to see if they do it on their own.   I don't think it's needed, I checked network traces of other MSALs.
- if we cannot remove, bound it as above

PR #616, opened 2026-04-23
Disable gzip compression #613

4gust:
# Description
This PR removes client‑side gzip support from the comm.Client HTTP stack and explicitly opts out of compression by setting Accept-Encoding: identity on all outbound requests. #613 

## Background
Previously, the client both advertised gzip support and automatically decompressed responses. This created a potential gzip decompression‑bomb vector, where a malicious or compromised endpoint could return a very small compressed payload that expands to unbounded memory usage.
Other MSAL SDKs (.NET, Java, Python, JS) already do not opt into compression for authentication endpoints, so this change also brings msal‑go into parity.

## Changes
Remove gzip handling and dependency from the HTTP client
Explicitly set Accept-Encoding: identity on all requests
Reject any response with a non‑identity Content-Encoding
Simplify response handling by reading bodies uncompressed
Add regression tests to lock in the new behaviour end‑to‑end

## Security impact

Closes a decompression‑bomb attack surface
Prevents unexpected transparent gzip handling at both SDK and net/http layers
Aligns with secure defaults and zero‑trust assumptions

## Testing

New unit tests verify headers and response handling
End‑to‑end test ensures JSONCall sends Accept-Encoding: identity
Existing tests continue to pass


apps/internal/oauth/ops/internal/comm/comm.go
```diff
@@ -284,25 +290,29 @@ func (c *Client) checkResp(v reflect.Value) error {
...
+	return io.ReadAll(resp.Body)
 }
 
 var testID string
 
 // addStdHeaders adds the standard headers we use on all calls.
 func addStdHeaders(headers http.Header) http.Header {
-	headers.Set("Accept-Encoding", "gzip")
+	// Explicitly opt out of compression. This both prevents a malicious endpoint from mounting a
+	// gzip decompression-bomb attack and disables Go's net/http transparent gzip handling (which
```
bgavrilMS: Let's please not use alarmist terms in comments.

4gust: okay i will update this in better terms

apps/internal/oauth/ops/internal/comm/comm.go
```diff
@@ -284,25 +290,29 @@ func (c *Client) checkResp(v reflect.Value) error {
...
+	// only kicks in when the caller has not set Accept-Encoding itself). Sibling MSAL SDKs
+	// (.NET / Java / Python / JS) also do not request compression on the auth endpoints.
+	headers.Set("Accept-Encoding", "identity")
```
bgavrilMS: Other MSALs don't set this. Do we to be explicit about it?

bgavrilMS (approved):
Approved with comments.

apps/internal/oauth/ops/internal/comm/comm.go
4gust: > **TL;DR** — Without this header, deleting our gzip code wouldn't actually fix the bug. Go's stdlib would silently take over compression handling and re-introduce the same decompression-bomb risk.
 
 ### The problem
 
 Go's `net/http` has a helpful default: if you don't set `Accept-Encoding` yourself, it:
 
 1. Adds `Accept-Encoding: gzip` to your request, and
 2. Transparently decompresses gzip responses behind the scenes.
 
 That decompression uses the same unbounded `gzip.Reader` we just removed — but now in stdlib code we can't audit, gate, or limit.
 
 ### The fix
 
 If we set **any** value for `Accept-Encoding`, Go steps back and lets us handle it. We pick `"identity"` — the official HTTP token for *"no encoding"*.
 
 This does two things:
 
 - Tells the **server**: don't compress.
 - Tells **Go's stdlib**: don't auto-decompress.
 
 Combined with `readBody` rejecting non-identity `Content-Encoding`, compressed responses can never reach a decoder.
 
 ### Why not `Transport.DisableCompression = true`?
 
 Users supply their own `*http.Client` via `WithHTTPClient`. We don't own their transport, so we can't flip that flag. The request header is the only opt-out under our control.
 
 ### References
 
 | Topic | Link |
 |---|---|
 | Go behavior we rely on | [`net/http.Transport` — `DisableCompression`](https://pkg.go.dev/net/http#Transport) |
 | Meaning of `identity` | [RFC 9110 §8.4.1](https://www.rfc-editor.org/rfc/rfc9110#section-8.4.1) |

bgavrilMS: Well presumably go stdlib handles this correctly. Do you have any evidence it doesn't ? We should just not be in the business of HTTP, except for resiliency purposes. Can we just not add any encoding header and not check any either?

4gust: Removed.

bgavrilMS (approved):

PR #625, opened 2026-05-25
Validate region string format in AADInstanceDiscovery

### LINKED ISSUE #624: Validate region string in AADInstanceDiscovery
The `AADInstanceDiscovery` function in `authority.go` uses the region string directly in URL construction (e.g., `https://{region}.login.microsoft.com/...`) without validating its format. If an unexpected or malformed region value is provided (e.g., containing dots, slashes, or other special characters), the resulting URL could be incorrect and lead to failed or misdirected requests.

Azure region names follow a consistent pattern of lowercase alphanumeric characters and hyphens (e.g., `eastus`, `westus2`, `east-us-2`). We should validate the region string against this pattern before using it in URL construction.

**Proposed fix:** Add a regex check (`^[a-z][a-z0-9-]*$`) and return a clear error if the region doesn't match.

bgavrilMS:
Fixes #624

## Problem
`AADInstanceDiscovery` uses the region string directly in URL construction (`https://{region}.{host}/...`) without validating its format. Malformed region values containing dots, slashes, or other special characters could produce incorrect URLs.

## Solution
Added a regex validation (`^[a-z][a-z0-9-]*$`) that checks the region string before it is used in URL construction. This matches the Azure region naming convention (e.g., `eastus`, `westus2`, `east-us-2`). If the region is invalid, a clear error is returned.

## Changes
- Added `regexp` import
- Added `validRegion` compiled regex at package level
- Added validation check in `AADInstanceDiscovery` before URL interpolation

### DIFF
--- a/apps/internal/oauth/ops/authority/authority.go
+++ b/apps/internal/oauth/ops/authority/authority.go
@@ -15,6 +15,7 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
@@ -43,6 +44,9 @@ const (
 	loginSTSWindows      = "sts.windows.net"
 	loginMicrosoftOnline = defaultHost
 )
+
+// validRegion matches Azure region names: lowercase alphanumeric and hyphens only.
+var validRegion = regexp.MustCompile(`^[a-z][a-z0-9-]*$`)
 
 // jsonCaller is an interface that allows us to mock the JSONCall method.
 type jsonCaller interface {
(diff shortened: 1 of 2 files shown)


apps/internal/oauth/ops/authority/authority.go
```diff
@@ -592,6 +596,9 @@ func (c Client) AADInstanceDiscovery(ctx context.Context, authorityInfo Info) (I
 		region = detectRegion(ctx)
```
4gust, line 600: We return error even when detected region is not valid ?

user input -> check - > pass or return error  
auto detect -> check -> pass or silent fall ?

4gust (approved):

4gust:
Needs unit test rest is ok

iarekk (approved):

PR #629, opened 2026-06-26
Add WithClaimsFromClient for client-originated claims

Robbie-Microsoft:
## Summary

Adds a per-request `WithClaimsFromClient(claims)` API on the confidential client that forwards **client-originated** JSON claims (e.g. an NSP `xms_az_nwperimid` claim) to the authority as the standard OAuth `claims` parameter. Ports the Confidential Client portion of [.NET MSAL #5999](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5999).

Unlike `WithClaims` (server challenges, which bypass the cache), client claims **participate in the cache**: the entry is keyed on the raw claims value, so different values get separate entries and the same value hits the cache.

## Scope

Confidential client, all six flows (credential, on-behalf-of, FIC, auth code, silent, username/password). Excluded from `AuthCodeURL`. Not added to public client or managed identity.

## Notes

- `CacheExtKeyGenerator` now length-prefixes cache-key components, changing the ext-cache-key hash for requests using `CacheKeyComponents` (`fmi_path`, `client_claims`). One-time re-fetch on upgrade; default cache unaffected.
- Filed a separate pre-existing bug (#630): proactively-refreshed app-only tokens export to `WithCache` under an empty partition key.

`gofmt`, `go vet ./...`, `go build ./...` clean; affected suites pass.


apps/internal/oauth/ops/authority/authority.go
```diff
@@ -365,6 +376,63 @@ func (p AuthParams) MergeCapabilitiesAndClaims() (string, error) {
...
+	m2, err := parseClaimsObject(claims2)
+	if err != nil {
+		return "", err
+	}
+	b, err := json.Marshal(deepMergeClaims(m1, m2))
+	if err != nil {
+		return "", err
+	}
+	return string(b), nil
+}
```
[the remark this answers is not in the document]
Robbie-Microsoft: Good eye, but this is intentional parity with the .NET PR being ported here. In `msal-dotnet`, `ClaimsHelper.MergeClaimsObjects` returns the single non-empty side **verbatim** (no parse), and `GetMergedClaimsAndClientCapabilities` returns `claims` verbatim when there are no client capabilities — `ParseClaimsOrThrow` only runs when capabilities exist or both claim sources are present. This Go code validates in exactly those same cases, so tightening it further would diverge from .NET (and from how server `WithClaims` already behaves on `main`, which also forwards verbatim absent capabilities).

The cache key is unaffected either way: it hashes the raw `CacheKeyComponents["client_claims"]` value, not the merge output. Invalid JSON surfaces as a server-side error, matching the other MSALs.

That said, the doc comment did overstate the invariant for the early-return branches — fixed in 4442d5c so it no longer implies the single-input paths validate.

apps/internal/oauth/ops/authority/authority.go
```diff
@@ -365,6 +376,63 @@ func (p AuthParams) MergeCapabilitiesAndClaims() (string, error) {
...
+// verbatim (the common case, which keeps the value byte-for-byte identical to what the caller passed).
+// When both are present they are parsed, deep-merged with the second object's values winning on
+// conflicting keys, and re-serialized. Each side must be a JSON object; anything else is an error.
+func mergeClaims(claims1, claims2 string) (string, error) {
+	if claims1 == "" {
+		return claims2, nil
+	}
+	if claims2 == "" {
+		return claims1, nil
+	}
```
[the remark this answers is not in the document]
Robbie-Microsoft: Good eye, but this is intentional parity with the .NET PR being ported here. In `msal-dotnet`, `ClaimsHelper.MergeClaimsObjects` returns the single non-empty side **verbatim** (no parse), and `GetMergedClaimsAndClientCapabilities` returns `claims` verbatim when there are no client capabilities — `ParseClaimsOrThrow` only runs when capabilities exist or both claim sources are present. This Go code validates in exactly those same cases, so tightening it further would diverge from .NET (and from how server `WithClaims` already behaves on `main`, which also forwards verbatim absent capabilities).

The cache key is unaffected either way: it hashes the raw `CacheKeyComponents["client_claims"]` value, not the merge output. Invalid JSON surfaces as a server-side error, matching the other MSALs.

That said, the doc comment did overstate the invariant for the early-return branches — fixed in 4442d5c so it no longer implies the single-input paths validate.

apps/internal/base/storage/partitioned_storage.go
```diff
@@ -59,7 +59,7 @@ func (m *PartitionedManager) Read(ctx context.Context, authParameters authority.
 
 	// errors returned by read* methods indicate a cache miss and are therefore non-fatal. We continue populating
 	// TokenResponse fields so that e.g. lack of an ID token doesn't prevent the caller from receiving a refresh token.
-	accessToken, err := m.readAccessToken(aliases, realm, clientID, userAssertionHash, scopes, partitionKeyFromRequest, tokenType, authnSchemeKeyID)
+	accessToken, err := m.readAccessToken(aliases, realm, clientID, userAssertionHash, scopes, partitionKeyFromRequest, tokenType, authnSchemeKeyID, authParameters.CacheExtKeyGenerator())
```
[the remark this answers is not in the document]
Robbie-Microsoft: Nice catch. `CacheExtKeyGenerator` is pre-existing (it already backed `fmi_path` partitioning on the standard manager) and unchanged by this PR — but you're right that it became reachable here. With a single component the separator-less concatenation can't collide, but this PR adds `client_claims` (arbitrary caller JSON) as a second component on the same credential-flow map, so `WithFMIPath(...)` + `WithClaimsFromClient(...)` with a value containing another component's key at a boundary could alias to the same hash and return the wrong cached token.

Fixed in 4442d5c by switching to a length-prefixed encoding (`len:key len:val` per sorted component) so the hash input is injective, plus a `TestCacheKeyComponentHashNoBoundaryCollision` regression test using exactly this ambiguous case. Note this changes the ext-cache-key format, so a persisted `fmi_path`/client-claims entry from a prior build is re-fetched once after upgrade (no data loss).

apps/confidential/confidential_test.go
```diff
@@ -2661,3 +2661,457 @@ func TestWithAttributeAndFMIPath(t *testing.T) {
 		t.Fatalf("Expected access token %q, got %q", accessToken, ar.AccessToken)
 	}
 }
+
+// placeholder-client-claims-tests
+
```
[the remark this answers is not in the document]
Robbie-Microsoft: Removed in 4442d5c — that was leftover scaffolding from when I stubbed the test section. Thanks.

apps/confidential/confidential_test.go
```diff
@@ -2661,3 +2661,325 @@ func TestWithAttributeAndFMIPath(t *testing.T) {
 		t.Fatalf("Expected access token %q, got %q", accessToken, ar.AccessToken)
 	}
 }
+
+// placeholder-client-claims-tests
```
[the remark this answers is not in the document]
Robbie-Microsoft: Removed in 4442d5c — that was leftover scaffolding from when I stubbed the test section. Thanks.

apps/managedidentity/managedidentity.go
```diff
@@ -325,6 +355,17 @@ func (c Client) AcquireToken(ctx context.Context, resource string, options ...Ac
 	}
 	c.authParams.Scopes = []string{resource}
 
+	// Client-originated claims participate in the cache (unlike server-issued WithClaims, which
+	// bypasses it). Partition the cache entry on the raw claims value so different claims values
+	// map to different tokens. The value is forwarded to IMDS later in the request pipeline.
```
4gust: the IMDS and `ClientClaims` check should be done before reading the cache no ?

apps/confidential/confidential.go
```diff
@@ -504,7 +508,77 @@ func WithClaims(claims string) interface {
 	}
 }
 
-// WithAuthenticationScheme is an extensibility mechanism designed to be used only by Azure Arc for proof of possession access tokens.
+// WithClaimsFromClient specifies client-originated claims to include in the token request, for example
+// an NSP (Network Security Perimeter) "xms_az_nwperimid" claim.
```
bgavrilMS: Not a 3p concept, do not use this as example.

apps/confidential/confidential.go
```diff
@@ -504,7 +508,77 @@ func WithClaims(claims string) interface {
 	}
 }
 
-// WithAuthenticationScheme is an extensibility mechanism designed to be used only by Azure Arc for proof of possession access tokens.
+// WithClaimsFromClient specifies client-originated claims to include in the token request, for example
```
bgavrilMS: Make it clear that the client claims MUST be included with each request.

apps/confidential/confidential.go
```diff
@@ -504,7 +508,77 @@ func WithClaims(claims string) interface {
...
+// Unlike [WithClaims] (for server-issued claims challenges, which bypass the token cache), tokens
+// acquired with client claims ARE cached and the cache entry is keyed on the claims value. Different
+// claims values produce separate cache entries, so callers should pass stable, non-dynamic values to
+// avoid unbounded cache growth and should pass the exact same string on each call (the raw string is
+// used verbatim as part of the cache key; MSAL does not normalize it).
+//
+// The claims are sent to the authority as the standard OAuth "claims" body parameter (merged with any
+// server-issued claims and client capabilities); they are not embedded in the client assertion JWT.
+// The argument must be a JSON object. An empty or whitespace-only value is ignored.
+// This option is valid for any confidential client token acquisition method.
```
bgavrilMS: comment not needed, the API is simply not there on PCA

apps/managedidentity/managedidentity.go
```diff
@@ -101,6 +101,13 @@ const (
 	identityServerThumbprintEnvVar      = "IDENTITY_SERVER_THUMBPRINT"
 
 	defaultRetryCount = 3
+
+	// clientClaimsCacheKey is the CacheKeyComponents key used to partition the token cache by
+	// client-originated claims (see WithClaimsFromClient). The component value is the raw claims string.
+	clientClaimsCacheKey = "client_claims"
+
+	// xmsAzNwperimid is the only custom claim MSIv1 (IMDS v1) accepts (see WithClaimsFromClient).
+	xmsAzNwperimid = "xms_az_nwperimid"
```
bgavrilMS: I recommend we do not make this sort of verification in the SDK. This is up to the service (IMDS) to accept / reject. 

I would like to understand the E2E developer experience here: 

1. you pass in client_claims with value other than "xms_az_nsp" to IMDSv1 - what does IMDSv1 do and should MSAL do ?
2. you pass in client_claims with value other than "xms_az_nsp" to ServiceFabric (or other sources) - what should we do? 
3. expected behavior on MSIv2 (not for MSAL Go)

IMO, for 1 we should let IMDS team decide. For 2, we should fail. For 3 - we should have a doc in MSAL.NET

bgavrilMS (approved):

apps/managedidentity/managedidentity.go
Robbie-Microsoft: Good catch. Moved the non-IMDS source guard and the JSON-object validation to the top of `AcquireToken`, before the cache read, so an invalid request (wrong source or malformed claims) fails fast instead of potentially returning a cached token. Fixed in 9e55427.

apps/confidential/confidential.go
Robbie-Microsoft: Removed the NSP / `xms_az_nwperimid` example from the public doc — it now just refers to client-originated claims as a JSON object, with no first-party concept. 9e55427.

apps/confidential/confidential.go
Robbie-Microsoft: Updated the doc to state the exact same string MUST be included on every request, and why: the raw value is part of the cache key, so omitting or changing it silently moves to a different cache partition. 9e55427.

apps/confidential/confidential.go
Robbie-Microsoft: Removed that line. 9e55427.

apps/managedidentity/managedidentity.go
Robbie-Microsoft: Agreed — removed the client-side allowlist (`validateMSIv1Claims`). MSAL now only validates that the value is a JSON object and forwards it as-is; IMDS decides which keys it accepts. This also matches MSAL.NET, which does not enforce an allowlist (it only does JSON-object validation via `ParseClaimsOrThrow`).

On the E2E questions:

1. **Non-`xms_az_nwperimid` claim to IMDSv1** — we now forward it and let IMDS accept/reject. Added `TestWithClaimsFromClientIMDSForwardsNonNwperimidClaims` asserting a non-nwperimid object is sent on the wire unchanged.
2. **Non-IMDS sources (ServiceFabric, App Service, Azure Arc, Azure ML, Cloud Shell)** — we fail at token-acquisition time, before any network call. `TestWithClaimsFromClientNonIMDSSourceErrors` covers all of them.
3. **MSIv2** — N/A for MSAL Go (there's no IMDSv2 path here); doc to live in MSAL.NET.

9e55427.

apps/confidential/confidential.go
```diff
@@ -504,7 +508,77 @@ func WithClaims(claims string) interface {
...
+// Unlike [WithClaims] (for server-issued claims challenges, which bypass the token cache), tokens
+// acquired with client claims ARE cached and the cache entry is keyed on the claims value. Different
+// claims values produce separate cache entries, so callers should pass stable, non-dynamic values to
+// avoid unbounded cache growth. The exact same string MUST be included on every request: the raw
+// string is used verbatim as part of the cache key (MSAL does not normalize it), so omitting it or
+// changing it on a later call silently moves to a different cache partition.
+//
+// The claims are sent to the authority as the standard OAuth "claims" body parameter (merged with any
+// server-issued claims and client capabilities); they are not embedded in the client assertion JWT.
+// The argument must be a JSON object. An empty or whitespace-only value is ignored.
```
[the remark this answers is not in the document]
Robbie-Microsoft: Good catch — the doc overstated local enforcement. Fixed in c219dc4: the docstring now states the single client-claims value is forwarded to the authority verbatim and is validated locally only when merged with server-issued claims or client capabilities; otherwise a malformed/non-object value surfaces as a server-side error. I kept the verbatim forwarding intentionally — the raw string is the cache-key component (re-serializing could reorder keys/whitespace and silently move cache partitions) and it mirrors .NET's MergeClaimsObjects.

4gust (approved):

4gust (approved):

4gust (approved):

PR #631, opened 2026-06-30
Fix app-only token proactive refresh writing to wrong cache partition key

### LINKED ISSUE #630: Proactively-refreshed app-only (client credentials) tokens are written to the wrong external-cache partition key
## Summary

When a confidential client is configured with an **external cache** (`confidential.WithCache(...)`) and acquires an **app-only token** (`AcquireTokenByCredential`, client-credentials grant), a **proactive refresh** (triggered once the token's `refresh_in`/`RefreshOn` has elapsed) writes the refreshed token to the external cache under the partition key `""` instead of the app-token-cache partition key used by reads. The next silent read therefore misses the refreshed token, so MSAL re-acquires from the IdP on every call after `RefreshOn` elapses, defeating the external cache for app tokens.

This is a **correctness/performance** issue (extra IdP round-trips + external-cache writes under an empty key), **not** a security issue. In-memory-only usage (no `WithCache`) is unaffected, because the in-memory manager writes via `authParams` directly rather than via the external-cache partition key.

## Root cause

The read and write paths compute the app-cache partition key differently during a proactive refresh:

- **Read** — `base.Client.AcquireTokenSilent` (`apps/internal/base/base.go`) computes the partition key as `authParams.CacheKey(silent.IsAppCache)`. For an app token (`IsAppCache == true`) this returns the app-token-cache key, e.g. `…_AppTokenCache`.
- **Write** — the proactive refresh calls `base.Client.AuthResultFromToken`, which computes the partition key as `token.CacheKey(authParams)` (`accesstokens.TokenResponse.CacheKey`, `apps/internal/oauth/ops/accesstokens/tokens.go`).

`AcquireTokenSilent` overrides `authParams.AuthorizationType` to `authority.ATRefreshToken` for the non-OBO path. So inside `TokenResponse.CacheKey`, the `ATClientCredentials` branch (which would return `authParams.AppKey()`) is skipped, and execution falls to the `IsConfidentialClient || AuthorizationType == ATRefreshToken` branch, which returns `tr.HomeAccountID()`. For an **app-only** token there is no user/home account, so `HomeAccountID()` is `""`.

Net result: read key = `…_AppTokenCache`, refresh write key = `""` → mismatch.

```
read  (AcquireTokenSilent):   authParams.CacheKey(true)        => "<clientID>_<tenant>_<…>_AppTokenCache"
write (AuthResultFromToken):  token.CacheKey(authParams)       => tr.HomeAccountID() => ""   // app token, ATRefreshToken branch
```

## Reproduction

1. Build a confidential client with an external cache: `confidential.New(authority, clientID, cred, confidential.WithCache(extCache))`.
2. `AcquireTokenByCredential(...)` where the token response carries a `refresh_in` (so `RefreshOn` is set).
3. Advance time past `RefreshOn`, then call `AcquireTokenByCredential(...)` again → proactive refresh.
4. Inspect the external cache: the refreshed token is exported under partition key `""`, separate from the original `…_AppTokenCache` partition. A subsequent acquire reads `…_AppTokenCache`, misses the refreshed token, and goes to the IdP again.

Observed while writing a regression test for client-claims cache partitioning (the empty-key export was visible in the `testCache` map dump: `CACHEKEY[1] len=… key=""`).

## Affected code

- `apps/internal/base/base.go` — `AuthResultFromToken` uses `key := token.CacheKey(authParams)` for the external-cache `Replace`/`Export`.
- `apps/internal/base/base.go` — `AcquireTokenSilent` sets `authParams.AuthorizationType = authority.ATRefreshToken` for the non-OBO path before the refresh write-back.
- `apps/internal/oauth/ops/accesstokens/tokens.go` — `TokenResponse.CacheKey` returns `HomeAccountID()` (empty for app tokens) under the `ATRefreshToken` branch, rather than `AppKey()`.

## Suggested fix direction

Make the refresh write-back use the **same** partition key as the read path for app tokens — e.g. have `AuthResultFromToken` (or `TokenResponse.CacheKey`) return `authParams.AppKey()` when the request is app-only/`IsAppCache`, even when `AcquireTokenSilent` has set `AuthorizationType = ATRefreshToken`. A focused unit test should assert that a proactively-refreshed app token round-trips through an external `WithCache` under the app-token-cache partition key (and is read back without an extra IdP call).

## Notes

- Pre-existing; discovered while porting [dotnet#5999](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5999) (`WithClaimsFromClient`) in PR #629. The client-claims change does **not** touch any of the affected code paths above; this report is split out so it can be fixed independently.

Robbie-Microsoft:
Fixes #630

## Summary

When a confidential client is configured with an external cache (`confidential.WithCache(...)`) and acquires an app-only token (`AcquireTokenByCredential`, client-credentials grant), a **proactive refresh** (triggered once `RefreshOn` has elapsed) wrote the refreshed token to the external cache under the partition key `""` instead of the app-token-cache key used by reads. The next silent read therefore missed the refreshed token, so MSAL re-acquired from the IdP on every call after `RefreshOn` elapsed, defeating the external cache for app tokens.

This is a correctness/performance issue, not a security issue. In-memory-only usage (no `WithCache`) is unaffected.

## Root cause

The read and write paths computed the app-cache partition key differently during a proactive refresh:

- **Read** -- `AcquireTokenSilent` computes `authParams.CacheKey(silent.IsAppCache)` => `..._AppTokenCache`.
- **Write** -- the proactive refresh calls `AuthResultFromToken`, which computes `token.CacheKey(authParams)`.

`AcquireTokenSilent` overrides `authParams.AuthorizationType` to `ATRefreshToken` for the non-OBO path. Inside `TokenResponse.CacheKey`, the `ATClientCredentials` branch is then skipped and execution falls to the `IsConfidentialClient || AuthorizationType == ATRefreshToken` branch, which returns `tr.HomeAccountID()`. For an app-only token there is no home account, so this is `""`.

## Fix

Make the refresh write-back use the same partition key as the read path for app tokens:

- Add an explicit `IsAppCache` flag to `authority.AuthParams`.
- Set it in `base.Client.AcquireTokenSilent` from `silent.IsAppCache` (it survives the `AuthorizationType = ATRefreshToken` override).
- Have `TokenResponse.CacheKey` return `authParams.AppKey()` when `IsAppCache` is set -- mirroring the read path's `authParams.CacheKey(isAppCache)`.

## Test

Added `TestAppTokenProactiveRefreshCachePartitionKey` (in `apps/internal/base/base_test.go`), which seeds a still-valid app token whose `RefreshOn` has elapsed, drives a proactive refresh through an external cache accessor, and asserts the export/replace partition keys equal `..._AppTokenCache` (not `""`). The test fails without the fix (`export key = ""`) and passes with it.

## Validation

- `go build ./...` -- clean.
- `go test ./apps/internal/... ./apps/confidential/...` -- pass.
- The only failing suites (`apps/tests/integration`, `apps/tests/performance`) fail identically on clean `main` (missing `cert.pem` / pre-existing panic) and are unrelated to this change.

## Notes

Pre-existing; discovered while porting dotnet#5999 (`WithClaimsFromClient`) in #629. Split out so it can be fixed independently.

### DIFF
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -334,6 +334,7 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	authParams.AuthorizationType = silent.AuthorizationType
 	authParams.Claims = silent.Claims
 	authParams.UserAssertion = silent.UserAssertion
+	authParams.IsAppCache = silent.IsAppCache
 	if silent.AuthnScheme != nil {
 		authParams.AuthnScheme = silent.AuthnScheme
 	}
--- a/apps/internal/base/base.go
+++ b/apps/internal/base/base.go
@@ -334,7 +334,7 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	authParams.AuthorizationType = silent.AuthorizationType
 	authParams.Claims = silent.Claims
 	authParams.UserAssertion = silent.UserAssertion
-	authParams.IsAppCache = silent.IsAppCache
+	authParams.IsAppTokenCache = silent.IsAppCache
 	if silent.AuthnScheme != nil {
 		authParams.AuthnScheme = silent.AuthnScheme
 	}
--- a/apps/internal/base/base_test.go
+++ b/apps/internal/base/base_test.go
@@ -413,6 +413,86 @@ func TestAccountCachePartitionKey(t *testing.T) {
 	}
 }
 
+// TestAppTokenProactiveRefreshCachePartitionKey verifies that a proactive refresh
+// (RefreshOn elapsed) of an app-only (client credentials) token writes the refreshed
+// token back to the external cache under the app-token-cache partition key, i.e. the
+// same partition key the read path uses, rather than the empty partition key "".
+//
+// This guards against the regression described in
+// https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/630, where
+// AuthResultFromToken computed the write-back key via token.CacheKey(authParams).
+// Because AcquireTokenSilent overrides AuthorizationType to ATRefreshToken and an
+// app-only token has no home account, that returned "" instead of the app-token-cache
+// key, so subsequent silent reads missed the refreshed token and re-hit the IdP.
+func TestAppTokenProactiveRefreshCachePartitionKey(t *testing.T) {
+	ctx := context.Background()
+	pkCache := &partitionKeyCache{}
+	client := fakeClient(t, WithCacheAccessor(pkCache))
+
+	// The proactively-refreshed token is app-only (no ID token), so its HomeAccountID()
+	// is empty — exactly the condition that exposed the bug.
+	client.Token.AccessTokens.(*fake.AccessTokens).AccessToken = accesstokens.TokenResponse{
+		AccessToken:   "refreshed-app-token",
+		ExpiresOn:     time.Now().Add(time.Hour),
+		GrantedScopes: accesstokens.Scopes{Slice: testScopes},
+		TokenType:     "Bearer",
+	}
+
+	// Seed the cache with a still-valid app token whose RefreshOn has already elapsed,
+	// so the next silent acquire serves it from cache and triggers a proactive refresh.
+	_, err := client.manager.Write(
+		authority.AuthParams{
+			AuthorityInfo: authority.Info{
+				AuthorityType: authority.AAD,
+				Host:          fakeAuthority,
+				Tenant:        fakeTenantID,
+			},
+			ClientID:          fakeClientID,
+			Scopes:            testScopes,
+			AuthorizationType: authority.ATClientCredentials,
+			AuthnScheme:       &authority.BearerAuthenticationScheme{},
+		},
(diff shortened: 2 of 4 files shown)


apps/internal/base/base.go
```diff
@@ -334,6 +334,7 @@ func (b Client) AcquireTokenSilent(ctx context.Context, silent AcquireTokenSilen
 	authParams.AuthorizationType = silent.AuthorizationType
 	authParams.Claims = silent.Claims
 	authParams.UserAssertion = silent.UserAssertion
+	authParams.IsAppCache = silent.IsAppCache
```
4gust: the name might be misleading.

4gust (approved):
Approved changes, just need the name change

apps/internal/base/base.go
Robbie-Microsoft: Good call - renamed it to `IsAppTokenCache` in e5d19be, which maps to the `_AppTokenCache` partition suffix the read path uses. Left the pre-existing `AcquireTokenSilentParameters.IsAppCache` as-is since that's existing API.

4gust (approved):

PR #632, opened 2026-07-01
Add SNI certificate over mTLS Proof-of-Possession (PoP) support

Robbie-Microsoft:
## SN/I certificate over mTLS PoP

Confidential client presents its SN/I cert as the **client TLS cert** to ESTS → mTLS-bound `token_type=mtls_pop` token instead of signing `private_key_jwt`. Opt-in, same credential; SNI+Bearer unchanged. Two-leg FIC stacked in **#633**.

- **API** (`apps/confidential`): `WithMtlsProofOfPossession()`, `WithMtlsHTTPClient(factory)`; result surfaces `Metadata.TokenType` / `BindingCertificate` (a `*tls.Certificate` carrying the parsed leaf **and** the private key, ready for `tls.Config.Certificates`) / `BindingCertificateThumbprint()`.
- **Behavior**: `login.*`→`mtlsauth.*` (region optional), AAD-only and tenanted-only, host must be a known Microsoft cloud, all sovereigns supported (legacy usgov+china aliases normalized to their preferred-network host, then rewritten), cert request omits `client_assertion`+`req_cnf`, **fail-closed** on a non-`mtls_pop` response (`MtlsPoPTokenTypeMismatchError`), AT cache keyed by `token_type`+cert.
- **Tests**: unit + ADO-only E2E `Credential_X509_Output_{Pop,Bearer}` — PoP global asserts `mtls_pop` and is proven usable by a strict-200 `mtls_pop`-scheme call to `mtlstb.graph.microsoft.com`; Bearer regional (westus3); no downgrade escape hatch.

---

## `BindingCertificate` carries the private key

`AuthResult.BindingCertificate` is **new in this PR** — it is absent from `main` and from every
released version, so there is no compatibility impact for existing callers.

An earlier revision of *this* PR exposed it as an `*x509.Certificate` — public material only — which
made the documented usage impossible: the README told callers to present it "as the client
certificate on the TLS handshake", but an `*x509.Certificate` has no private key and cannot go into
`tls.Config.Certificates`. MSAL .NET returns an `X509Certificate2` carrying the key handle, so that
was also a parity gap. It was changed to `*tls.Certificate` in `d8e8619`, still within this PR.

It therefore ships as a `*tls.Certificate` with `.Leaf` always populated and `PrivateKey` set to the
key MSAL used.

**Non-exportable keys are out of scope here.** `NewCredFromCert` still requires a concrete
[...]

PR #633, opened 2026-07-06
Add two-leg FIC support over mTLS Proof-of-Possession (PoP)

Robbie-Microsoft:
## Summary

Follow-up to #632. Adds the **developer-orchestrated two-leg Federated Identity Credential (FIC) flow** over mTLS PoP. Leg 1 obtains a certificate-bound federated assertion over mTLS; leg 2 presents that assertion as a **`jwt-pop` client assertion** together with the binding certificate to obtain an mTLS-bound PoP token.

> Stacked on **#632** — review/merge that first. This diff is the FIC delta only.

## Public API (`apps/confidential`)

- `NewCredFromSignedAssertionCallback(cb)` — the leg-2 credential. The callback returns the assertion and its binding certificate together, so the two cannot desync across certificate rotation.
- `SignedAssertion{Assertion, BindingCertificate}` — mirrors MSAL .NET's `ClientSignedAssertion{Assertion, TokenBindingCertificate}`. `BindingCertificate` is a `*tls.Certificate`, so it accepts any `crypto.Signer`, including CNG/HSM-backed keys.

`WithMtlsProofOfPossession()` is unchanged from #632.

## Behavior

- `FromAssertion` sends the `jwt-pop` `client_assertion_type` when mTLS PoP is requested; assertion/callback credentials (FIC leg 2) route to that path.
- An assertion credential requires `NewCredFromSignedAssertionCallback`; a certificate credential (`NewCredFromCert`) infers its binding certificate.
- Under mTLS PoP the signed-assertion callback is invoked **before** the cache read, because the certificate thumbprint partitions the token cache. It therefore runs on every acquisition, including cache hits, unlike a plain assertion callback. MSAL .NET has the same property and documents it on `ClientSignedAssertion`: "It is up to the caller to cache the assertion and certificate if reuse is desired." The assertion is memoized on a per-request copy of the credential; the client's credential is never mutated.

## Considered and dropped

Call-site binding-certificate options (`WithMtlsBindingCertificate`, `WithMtlsBindingTLSCertificate`, and the `MtlsPoPOption` interface that carried them).

MSAL .NET exposes only the paired form and has no call-site binding-certificate supplier anywhere in its public surface; its string-returning `WithClientAssertion` overloads are annotated `// String assertion => cannot return TokenBindingCertificate => use string credential`, i.e. callers are deliberately forced onto the paired overload.

The paired form is also strictly more capable here: `SignedAssertion.BindingCertificate` accepts any `crypto.Signer`, whereas `WithMtlsBindingCertificate` was RSA-only. Keeping both would have reintroduced the assertion/certificate mismatch that `SignedAssertion` exists to make unrepresentable, and forced `WithMtlsProofOfPossession` to become variadic — a signature change visible to every #632 caller.

## Tests

- **Unit**: `TestAcquireTokenByCredentialFICMtlsPoP`, `TestFromAssertionMtlsPoP`, and 21 signed-assertion tests covering callback invocation count, cache partitioning by certificate, deep-copied DER, mismatched keys, ECDSA/opaque signers, and error propagation.
- **E2E** `TestTwoLegFICMtlsPoP_SNI` — **runs unconditionally and passes.** It was previously skip-gated behind `MSAL_RUN_FIC_MTLS_E2E` pending confirmation of the leg-2 `jwt-pop` contract, but the gate was masking a test-configuration bug rather than a protocol gap: the test targeted the FMI lab app, which is not enabled for mTLS client auth, so leg 1 failed with `AADSTS51000 MtlsClientAuth is/are disabled` and leg 2 was never reached. Both legs now target the SN/I allow-listed app in the MSI team tenant — the same app the passing SNI tests in #632 use, and the one MSAL .NET's `Credential_Fic_Output_Pop_TestAsync` uses for this exact flow.

  The test asserts both legs return `token_type=mtls_pop`, that leg 2's token carries leg 1's binding-certificate thumbprint, and that Microsoft Graph accepts the final token over mTLS (HTTP 200). Leg 2 carries leg 1's `BindingCertificate` forward as-is, mirroring .NET's `ClientSignedAssertion.TokenBindingCertificate`. The `acquireMtlsPoPOrSkip` escape hatch, which downgraded an `mtls_pop` -> `Bearer` mismatch to a skip, is removed: both legs now use the same app and endpoint as the SNI tests, which already fail closed.

## Docs

`README.md` two-leg FIC section; `docs/federated_managed_identity.md` subsection; runnable example in `apps/confidential/examples_test.go`.

## Known divergence from MSAL .NET

Go selects the `jwt-pop` `client_assertion_type` based on `IsMtlsPoP`; .NET selects it based on the *presence of a binding certificate*. .NET can therefore express "bound assertion, unbound token" and Go currently cannot. Deliberate for now — noted so it is a conscious choice rather than an oversight.

## Devex

Two clients. The callback **is the credential of the leg-2 client**; its body performs leg 1.

```go
// Client 1 — real certificate. This IS leg 1.
leg1App, _ := confidential.New(authority, clientID, leg1Cred /* NewCredFromCert */)
leg1, _ := leg1App.AcquireTokenByCredential(ctx, []string{exchangeScope},
    confidential.WithMtlsProofOfPossession())
// leg1.AccessToken is a FEDERATED ASSERTION -- not a usable resource token --
// and leg1.BindingCertificate is the certificate it is bound to.

// Client 2 — its credential is the callback.
leg2Cred := confidential.NewCredFromSignedAssertionCallback(
    func(ctx context.Context, o confidential.AssertionRequestOptions) (confidential.SignedAssertion, error) {
        return confidential.SignedAssertion{
            Assertion:          leg1.AccessToken,
            BindingCertificate: leg1.BindingCertificate, // nil -> falls back to the client's own cert
        }, nil
    })
leg2App, _ := confidential.New(ficAuthority, ficClientID, leg2Cred)
final, _ := leg2App.AcquireTokenByCredential(ctx, []string{resourceScope},
    confidential.WithMtlsProofOfPossession())

// NEW STEP -- using the token takes both halves, and the certificate is LEG 1's:
tls.Config{Certificates: []tls.Certificate{*final.BindingCertificate}}
req.Header.Set("Authorization", "mtls_pop "+final.AccessToken)
// final.BindingCertificateThumbprint() == leg1.BindingCertificateThumbprint().
// Client 2 owns no certificate; ESTS bound client 2's token to client 1's cert.
```

`WithMtlsProofOfPossession()` on **both** legs. Leg 2 is the `AcquireTokenByCredential` on client 2 —
the callback is only *how client 2 proves who it is*, and proving that requires leg 1's output.

- **Purely additive.** `NewCredFromAssertionCallback` is unchanged; its callback still returns a bare
  `string`. `SignedAssertion` is that same value plus the certificate, so the only new thing to learn
  is the return type. Existing callers are untouched.
- **One new exported API.** The assertion and its binding certificate are returned **together by
  design**: there is deliberately no `WithBindingCertificate()` call-site option, and that absence
  *is* the enforcement — if you cannot supply them separately, you cannot mismatch them across a
  certificate rotation. Mirrors .NET's `ClientSignedAssertion.TokenBindingCertificate`.
- **Callback timing differs under mTLS PoP.** The binding certificate partitions the token cache and
  selects the TLS connection, so the callback is resolved *before* the cache is consulted — unlike
  bearer/OBO, where it fires only when a request is actually sent. Keep leg 1 on an MSAL client so
  its own cache absorbs the repeat calls.
- **Leg 2 wire: `client_assertion` returns**, with
  `client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-pop`.
- The assertion story across the stack inverts: vanilla cert auth `jwt-bearer` →
  #632 *no assertion at all* → #633 leg 2 `jwt-pop`.
- Using the plain string callback with `WithMtlsProofOfPossession()` fails with a message naming the
  fix; an empty assertion is rejected locally rather than sent.

A runnable demo lives in `apps/tests/devapps/mtlsfic/`.


apps/confidential/confidential.go
```diff
@@ -1001,9 +1046,43 @@ func WithMtlsProofOfPossession() interface {
...
+		return nil, errors.New("mTLS binding certificate key must be an RSA key")
+	}
+	tlsCert := tls.Certificate{PrivateKey: key}
+	for _, cert := range certs {
+		if cert == nil {
+			continue
+		}
+		certKey, ok := cert.PublicKey.(*rsa.PublicKey)
+		if ok && k.E == certKey.E && k.N.Cmp(certKey.N) == 0 {
+			// The signing (leaf) cert matches the private key and must be first.
```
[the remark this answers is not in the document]
Robbie-Microsoft: Done in adacb9a: newTLSBindingCertificate now accepts any crypto.Signer and matches the leaf via signer.Public(), so CNG/HSM-backed RSA keys work while keeping the RSA-only constraint.

apps/confidential/confidential.go
```diff
@@ -967,32 +969,75 @@ func WithAttribute(attrValue string) interface {
...
 				switch t := a.(type) {
 				case *acquireTokenByCredentialOptions:
 					t.isMtlsPoP = true
+					t.mtlsBindingCert = bindingCert
 				case *acquireTokenSilentOptions:
 					t.isMtlsPoP = true
+					t.mtlsBindingCert = bindingCert
 				default:
 					return fmt.Errorf("unexpected options type %T", a)
 				}
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed in 67fe417: WithMtlsProofOfPossession now only sets mtlsBindingCert when the option actually carries one, so composing it more than once can't clear a previously supplied cert.

apps/confidential/confidential.go
```diff
@@ -1019,22 +1104,22 @@ func (cca Client) resolveMtlsBindingCert() (*tls.Certificate, error) {
...
+	return nil, errors.New("mTLS proof-of-possession requires a certificate credential (NewCredFromCert) or WithMtlsBindingCertificate")
 }
 
 // validateMtlsCredential rejects credential kinds that can't perform mTLS proof-of-possession.
-func validateMtlsCredential(cred *accesstokens.Credential) error {
+func validateMtlsCredential(cred *accesstokens.Credential, explicitBindingCert *tls.Certificate) error {
 	if cred == nil {
-		return errors.New("mTLS proof-of-possession requires a certificate credential (NewCredFromCert)")
+		return errors.New("mTLS proof-of-possession requires a certificate or assertion credential")
 	}
```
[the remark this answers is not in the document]
Robbie-Microsoft: Leaving as-is: this branch is the 'no credential supplied' case. The certificate vs. assertion+binding-cert guidance is already surfaced by the adjacent checks - validateMtlsCredential rejects an assertion credential without WithMtlsBindingCertificate, and resolveMtlsBindingCert names both supported combinations.

4gust (approved):

apps/confidential/confidential.go
```diff
@@ -1102,9 +1153,49 @@ func WithMtlsProofOfPossession() interface {
...
-func (cca Client) resolveMtlsBindingCert() (*tls.Certificate, error) {
+// newTLSBindingCertificate assembles a tls.Certificate from a certificate chain and an RSA private
+// key, placing the signing (leaf) certificate first as required for the TLS handshake. The key may
+// be any RSA key implementing crypto.Signer (including CNG/HSM-backed keys), matched to the leaf via
+// its public key.
+func newTLSBindingCertificate(certs []*x509.Certificate, key crypto.PrivateKey) (*tls.Certificate, error) {
+	if len(certs) == 0 || key == nil {
+		return nil, errors.New("WithMtlsBindingCertificate requires a certificate and private key")
+	}
+	signer, ok := key.(crypto.Signer)
```
gladjohn: The follow-up PR #633 actually shows the way forward here. Its WithMtlsBindingCertificate(certs, key) already accepts any RSA crypto.Signer, explicitly including CNG/HSM-backed keys — the binding-cert key never has to be exportable, because Go just calls Sign() on it.

Wire the mTLS client cert's key through crypto.Signer / tls.Config on #632 for the binding cert, so non-exportable keys work out of the box.

apps/confidential/confidential_mtls_test.go
```diff
@@ -140,6 +140,71 @@ func TestAcquireTokenByCredentialMtlsPoP(t *testing.T) {
...
+		mock.WithBody(mtlsPoPTokenBody("final-mtls-token", 3600)),
+		mock.WithCallback(func(r *http.Request) {
+			gotURL = r.URL
+			b, _ := io.ReadAll(r.Body)
+			gotBody, _ = url.ParseQuery(string(b))
+		}),
+	)
+
+	res, err := client.AcquireTokenByCredential(ctx, tokenScope,
+		WithMtlsProofOfPossession(WithMtlsBindingCertificate(certs, key)))
```
gladjohn: This demonstrates a gap from the MSAL.NET FIC handoff. In .NET, CCA receives the assertion and its complete provider-backed binding certificate directly from CCA through ClientSignedAssertion{ Assertion: leg1.AccessToken, TokenBindingCertificate: leg1.BindingCertificate }. Here, leg1.BindingCertificate is public-only, so CCA cannot consume CCA result directly; the application must retain and separately supply the original certs and private key/crypto.Signer. This also separates the assertion from its binding credential and could allow a mismatch during certificate rotation. Should the assertion callback return both the assertion and an opaque signer-backed binding credential, similar to MSAL.NET’s ClientSignedAssertion?

apps/confidential/confidential.go
Robbie-Microsoft: Done in #647: `NewCredFromTLSCertificate` accepts any `crypto.Signer`, so the binding-cert key never has to be exportable. Validated against a real KeyGuard key (`Virtual Iso`=1, export blocked by the OS) - live `mtls_pop` token and Graph HTTP 200.

apps/confidential/confidential_mtls_test.go
Robbie-Microsoft: Both addressed in d891024 on this PR.

`leg1.BindingCertificate` is no longer public-only — `d8e8619` made it a `*tls.Certificate`, so it carries the key, including an opaque `crypto.Signer` (KeyGuard/CNG). Two additions:

```go
// pass leg 1's result straight through, no decomposition
WithMtlsProofOfPossession(WithMtlsBindingTLSCertificate(leg1.BindingCertificate))

// or keep assertion and certificate paired, per ClientSignedAssertion
NewCredFromSignedAssertionCallback(func(ctx context.Context, o AssertionRequestOptions) (SignedAssertion, error) {
    return SignedAssertion{Assertion: leg1.AccessToken, BindingCertificate: leg1.BindingCertificate}, nil
})
```

The callback form closes the rotation-mismatch hole you flagged: the assertion and its certificate are one value.

`NewCredFromAssertionCallback` is byte-for-byte unchanged (shipped v1.4.0-v1.8.0), so this is additive.

Ordering: the certificate is needed before the body is built because x5t#S256 partitions the token cache. Resolved by pulling the callback forward once, only when its cert is actually needed, then memoizing — the callback is invoked exactly once per request and never on a cache hit. Precedence: explicit option > callback cert > credential cert.

16 tests, real TLS handshake with `RequireAnyClientCert`; one uses a `crypto.Signer` that refuses to expose its inner key, so a hidden `*rsa.PrivateKey` assertion anywhere in the path fails. 13 mutations, all caught.

apps/confidential/confidential.go
```diff
@@ -1106,22 +1339,83 @@ func (cca Client) resolveMtlsBindingCert() (*tls.Certificate, error) {
...
+// The callback is not pulled forward when it isn't needed: with an explicit binding certificate, or
+// on any non-mTLS request, the certificate is already known, so the callback stays lazy and runs
+// only if a token request is actually sent (never on a cache hit), exactly as a plain assertion
+// callback does.
+func (cca Client) prepareMtlsPoP(ctx context.Context, authParams authority.AuthParams, explicit *tls.Certificate) (*accesstokens.Credential, *tls.Certificate, error) {
+	cred := cca.cred
+	if err := validateMtlsCredential(cred, explicit); err != nil {
+		return nil, nil, err
+	}
+	if explicit == nil && cred != nil && cred.SignedAssertionCallback != nil {
```
gladjohn: The explicit-certificate precedence can break the atomic assertion/certificate pairing this API is intended to provide. When an explicit certificate is supplied, the signed-assertion callback remains lazy; it can later return an assertion bound to certificate A while the mTLS transport uses explicit certificate B, and the callback certificate is ignored. Please either reject combining these two certificate sources or invoke/validate the paired result and fail when the leaf public key or thumbprint does not match.

apps/confidential/confidential.go
```diff
@@ -1089,9 +1246,85 @@ func WithMtlsProofOfPossession() interface {
...
+	if cert == nil {
+		return nil, errors.New("binding certificate is nil")
+	}
+	if len(cert.Certificate) == 0 || len(cert.Certificate[0]) == 0 {
+		return nil, errors.New("binding certificate carries no certificate chain")
+	}
+	if cert.PrivateKey == nil {
+		return nil, errors.New("binding certificate has no private key, so it cannot be presented on a TLS handshake")
+	}
+	if _, ok := cert.PrivateKey.(crypto.Signer); !ok {
```
gladjohn: This verifies that the private key implements crypto.Signer, but not that signer.Public() matches the leaf certificate public key. A mismatched *tls.Certificate therefore passes local validation and fails later with an opaque TLS handshake error. newTLSBindingCertificate already enforces this invariant; validBindingCertificate should perform the same public-key comparison before retaining or using the certificate.

apps/confidential/confidential.go
```diff
@@ -1089,9 +1246,85 @@ func WithMtlsProofOfPossession() interface {
...
+		return nil, fmt.Errorf("binding certificate private key of type %T does not implement crypto.Signer", cert.PrivateKey)
+	}
+	out := *cert
+	out.Certificate = append([][]byte(nil), cert.Certificate...)
```
gladjohn: This only copies the outer [][]byte slice; each DER byte slice is still shared with the caller. A caller mutating cert.Certificate[0][0] after this call can alter the certificate retained by MSAL and potentially race with an acquisition. Please deep-copy every DER entry, and add a test that mutates bytes within an existing entry rather than replacing the whole entry.

apps/confidential/confidential.go
```diff
@@ -1106,22 +1339,83 @@ func (cca Client) resolveMtlsBindingCert() (*tls.Certificate, error) {
...
+			if err != nil {
+				return nil, nil, err
+			}
+			authParams.Endpoints = endpoints
+		}
+		signed, err := cred.SignedAssertion(ctx, authParams)
+		if err != nil {
+			return nil, nil, err
+		}
+		assertion := signed.Assertion
```
gladjohn, line 1450: Please reject an empty or whitespace-only signed assertion here before memoizing it. Otherwise the request proceeds with an empty client_assertion and fails remotely as a malformed token request. MSAL.NET validates this credential result locally and returns InvalidClientAssertion, which gives callers deterministic behavior and a useful error.

apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -299,7 +322,13 @@ func (c Client) FromAssertion(ctx context.Context, authParameters authority.Auth
 		return TokenResponse{}, err
 	}
 	qv.Set(grantType, grant.ClientCredential)
-	qv.Set("client_assertion_type", grant.ClientAssertion)
+	// For mTLS proof-of-possession the client assertion is certificate-bound: signal it with the
+	// jwt-pop assertion type and present the binding certificate on the TLS handshake (see doTokenResp).
+	assertionType := grant.ClientAssertion
+	if authParameters.IsMtlsPoP {
```
gladjohn: This leaves a functional gap versus MSAL.NET: .NET selects jwt-pop client authentication and the mTLS transport whenever ClientSignedAssertion includes a token-binding certificate, independently of whether the final access token requested is Bearer or MTLS_POP. Here jwt-pop is selected only by IsMtlsPoP, so a certificate-bound FIC assertion cannot be exchanged for an ordinary Bearer access token. Can the binding-certificate/client-auth mode be modeled separately from the requested access-token type?

apps/tests/integration/mtls_pop_integration_test.go
```diff
@@ -205,6 +209,86 @@ func TestCredential_X509_Output_Bearer(t *testing.T) {
...
+		t.Fatal("leg 1 did not return a federated assertion")
+	}
+	if leg1.BindingCertificate == nil {
+		t.Fatal("leg 1 result missing binding certificate")
+	}
+
+	// Leg 2: federated assertion (jwt-pop) + the SAME binding certificate -> final mtls_pop token.
+	// The certificate is carried forward exactly as leg 1 returned it, so a regression that strips
+	// the private key or swaps the certificate fails here rather than silently rebinding.
+	leg2Cred := confidential.NewCredFromAssertionCallback(
```
gladjohn: The live two-leg test still uses the legacy assertion callback plus a separately supplied certificate, so the new preferred paired SignedAssertion callback is covered only by local/mock tests. Please add or convert an E2E leg to NewCredFromSignedAssertionCallback returning both leg1.AccessToken and leg1.BindingCertificate. That would validate the exact handoff API introduced to prevent assertion/certificate rotation mismatches.

gladjohn (changes requested):
Requesting changes for the assertion/certificate binding correctness issues below. I also included the related parity and E2E coverage gaps so the intended flow is validated end to end.

apps/confidential/confidential.go
Robbie-Microsoft: Compares `signer.Public()` against the key parsed from `Certificate[0]`, via the stdlib `Equal(crypto.PublicKey)` interface rather than the RSA-modulus form — this path also carries opaque CNG/KeyGuard signers.

Also rejects a `Leaf` disagreeing with `Certificate[0]` instead of overriding it: keyID (`authority/mtls.go:38`) and cache partition (`base.go:150`) hash `Leaf.Raw`, the mTLS client cache (`comm/mtls.go:113`) hashes `Certificate[0]`. A mismatch splits the two caches.

apps/confidential/confidential.go
Robbie-Microsoft: Per-entry copy; tests mutate `Certificate[0][0]` and `[1][0]`.

Confirming your read: `TestWithMtlsBindingTLSCertificateCopiesInput` and `TestSignedAssertionCallbackCopiesCertificate` both still passed against the shallow copy — they replace entries, which the outer copy already covered.

apps/confidential/confidential.go
Robbie-Microsoft: Agreed — taking the reject option. `validateMtlsCredential` now errors when an explicit binding certificate is combined with `NewCredFromSignedAssertionCallback`. Explicit cert + plain `AssertionCallback` is unchanged.

.NET has no explicit binding-cert option — the cert comes only from `ClientSignedAssertion.TokenBindingCertificate` — so there was no precedence rule to preserve.

Consequence: the callback now resolves on every mTLS PoP acquisition including cache hits, matching .NET. `TestSignedAssertionExplicitCertificateTakesPrecedence` inverted.

apps/confidential/confidential.go
Robbie-Microsoft: Trimmed-empty assertion rejected before memoizing. `InvalidClientAssertion` parity.

apps/internal/oauth/ops/accesstokens/accesstokens.go
Robbie-Microsoft: Agreed — separated. Client auth type now follows the binding certificate (`IsMtlsPoP || MtlsBindingCert != nil`), not the requested token type, mirroring `Mtls || hasCert`.

Scoped to client auth only: transport routing still keys on `IsMtlsPoP`, because #643 models Bearer-over-mTLS with its own flag and folding both into one condition recreates the coupling you're pointing at.

The new arm is inert on this branch — the only site setting `MtlsBindingCert` also sets `IsMtlsPoP` — and goes live with #643. Covered meanwhile by a unit test calling `FromAssertion` with a binding certificate and `IsMtlsPoP` false.

apps/tests/integration/mtls_pop_integration_test.go
Robbie-Microsoft: Converted leg 2 to `NewCredFromSignedAssertionCallback` returning `leg1.AccessToken` + `leg1.BindingCertificate`.

apps/confidential/confidential.go
```diff
@@ -197,10 +229,56 @@ func NewCredFromSecret(secret string) (Credential, error) {
...
+//	    })
+//
+// MSAL invokes the callback at most once per token request, and never when the request is served
+// from the cache without it. Because the binding certificate partitions the token cache and selects
+// the mutual-TLS connection, an mTLS proof-of-possession request resolves the callback before the
+// cache is consulted; requests that don't need the certificate (bearer client credentials or
+// on-behalf-of) invoke it only when a token request is actually sent, exactly like
+// [NewCredFromAssertionCallback].
+//
+// The callback must return a non-empty assertion; an empty or whitespace-only one is rejected
```
gladjohn: Two documentation-contract issues (non-blocking): (1) the statement that an empty or whitespace-only assertion is rejected locally currently applies only to the explicit mTLS-PoP preparation path; a signed-assertion credential used without `WithMtlsProofOfPossession` forwards it like `NewCredFromAssertionCallback`. (2) the documented nil-`BindingCertificate` fallback to the client certificate is not constructible because a `Credential` cannot be both certificate-backed and signed-assertion-callback-backed. Please scope the empty-assertion statement to the PoP path and make the callback certificate explicitly required for signed-assertion mTLS.

apps/tests/integration/mtls_pop_integration_test.go
```diff
@@ -253,6 +257,95 @@ func TestCredential_X509_Output_Bearer(t *testing.T) {
...
+// Leg 2 uses NewCredFromSignedAssertionCallback, the paired handoff API, so this test covers the
+// exact path applications are meant to use to carry leg 1's assertion and certificate forward
+// together rather than through two independent options.
+//
+// Both legs use the SN/I allow-listed app in the MSI team tenant and the global (non-regional)
+// endpoint, which is the only configuration ESTS enables for mTLS PoP. Like the SNI tests, both legs
+// fail closed on an ESTS token_type downgrade (mtls_pop -> Bearer) rather than treating it as
+// inconclusive. Mirrors MSAL .NET's Credential_Fic_Output_Pop_TestAsync /
+// RunTwoLegS2sFicBothLegsPopAsync.
+func TestTwoLegFICMtlsPoP_SNI(t *testing.T) {
```
gladjohn, line 477: **[Acceptance blocker]** The two-leg FIC test is present at this head, but the current-head ADO integration pipeline was skipped by branch filtering, GitHub runs `-short`, and no current-head CodeQL result exists. Please run and attach the required ADO integration/security jobs for `9facda3`, including T1→T2 `cnf.x5t#S256` continuity, the successful resource call, and no-certificate/wrong-scheme/wrong-certificate rejection cells. Source presence alone does not prove this critical flow executed.

apps/confidential/confidential.go
Robbie-Microsoft: Both confirmed, fixed in `e8accf2` + `8007d86`.

(1) Scoped to the PoP path. Docs-only — adding validation off that path would diverge from `NewCredFromAssertionCallback`.

(2) Certificate now required. `prepareMtlsPoP` rejects a nil one naming the callback and field, instead of falling through to a generic error offering `NewCredFromCert` — which this credential can't use. `NewCredFromCert` + PoP unchanged.

The fallback claim was in five places, not two: both `confidential.go` docs, `exported.SignedAssertion`, and the `mtlsfic` README and `main.go`. Deleted, not reworded.

Tests pin both directions — empty assertion and nil cert accepted off the PoP path, rejected on it. Mutation-checked.

apps/tests/integration/mtls_pop_integration_test.go
Robbie-Microsoft: Ran it — [ADO build 1685394](https://dev.azure.com/IdentityDivision/a7934fdd-dcde-4492-a406-7fad6ac00e17/_build/results?buildId=1685394) on the current head `8007d86`, succeeded. `msal-go-ci` on this PR now reads success, not neutral.

Integration Tests: 24 tests, 23 passed, 0 failed. `TestAdfsToken` is a pre-existing skip. `TestTwoLegFICMtlsPoP_SNI` passed, 1460 ms.

Your five cells: continuity and the resource call are in that test — the thumbprint comparison is fatal, then `requireTokenAcceptedByResource`. The three rejections (`MtlsMissingClientCertificate`, `Cnf claim not supported over Bearer protocol.`, `MtlsCnfClaimRequestDataValidationFailed`) are on `TestCredential_X509_Output_Pop`, also green in this build. They sit there deliberately: enforcement is a property of the resource, not of each test.

CodeQL doesn't run on this PR — it runs once this merges down to the feature branch.

PR #638, opened 2026-07-27
deps: replace pkg/browser with cli/browser

### LINKED ISSUE #427: Replace github.com/pkg/browser dependency
Consider replacing the dependency on github.com/pkg/browser which appears to be dead (no commits for 20+ months). Fork github.com/cli/browser is actively developed, and [packaged in more Linux distributions](https://repology.org/project/go:github-cli-browser/versions). This would simplify packaging this library for Linux.

arnaud-dezandee:
Fixes #427.

This re-applies #428, which was merged into the `dev` branch in October 2023 and never landed on `main`. That branch no longer exists upstream, so the change was lost.


arnaud-dezandee:

4gust (approved):

PR #640, opened 2026-07-28
ci: split build_test pipeline into Build / Unit Tests / Integration Tests stages

gladjohn:
## Summary

Refactors the single-job `ado/build_test.yaml` into **three stages** — **Build**, **Unit Tests**, **Integration Tests** — mirroring the MSAL .NET pipeline layout so each section reports independently in Azure DevOps (instead of one long job).

## Before / After

**Before:** one `steps:` job (`GoTool → go get → go build → unit tests → Key Vault + cert → integration tests`).

**After:** `stages:`
- **Build** — `GoTool` + `go build ./apps/...`
- **Unit Tests** (`dependsOn: Build`) — `go test -race -short ./apps/cache/... ./apps/confidential/... ./apps/public/... ./apps/internal/... ./apps/managedidentity/...`
- **Integration Tests** (`dependsOn: Build`) — Key Vault + cert install + `go test -race ./apps/tests/integration/...`

## Notes

- **No change to what is built or tested** — same commands, same coverage, same Key Vault secrets and cert setup.
- Each stage runs on `ubuntu-latest` and re-installs the Go toolchain + deps (Go builds are cheap; keeps stages independent, matching how MSAL .NET's stages build from source).
- Adds **`gotestsum` + `PublishTestResults@2`** so per-stage **pass rates** show up in the ADO test UI (Go's default `go test` doesn't publish results). If you'd rather keep PR 1 minimal, the `gotestsum` bits can be dropped and we revert to the `Go@0` test task.
- Go version is lifted into a `goVersion` variable (`1.22.3`).

## Follow-up

A second PR will add the **Managed Identity v1 (IMDS) E2E** stage on the `MISEManagedIdentity` pool (the same agent MSAL .NET uses), plus a Go E2E test.

---

🤖 Prototyped during an investigation with GitHub Copilot.

### DIFF
--- a/ado/build_test.yaml
+++ b/ado/build_test.yaml
@@ -7,50 +7,129 @@ pr:
     include:
       - main
 
-pool:
-  vmImage: "ubuntu-latest"
+variables:
+  goVersion: "1.22.3"
 
-steps:
-  - task: GoTool@0
-    inputs:
-      version: "1.22.3"
-  - task: Go@0
-    inputs:
-      command: "get"
-      arguments: "-d -v -t -d ./..."
-      workingDirectory: "$(System.DefaultWorkingDirectory)"
-    displayName: "Install dependencies"
-  - task: Go@0
-    inputs:
-      command: "build"
-      arguments: "./apps/..."
-      workingDirectory: "$(System.DefaultWorkingDirectory)"
+stages:
+  # ──────────────────────────────────────────────
+  # Stage 1 — Build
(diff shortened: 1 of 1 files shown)


gladjohn:
Consolidated into #641, which now targets main and contains the full change set (pipeline staging + gotestsum pin + Managed Identity IMDS E2E tests). Closing this stacked PR.

PR #643, opened 2026-07-31
Add Bearer-over-mTLS (WithSendCertificateOverMtls)

Robbie-Microsoft:
## Bearer-over-mTLS (`WithSendCertificateOverMtls`)

Ports MSAL .NET's `CertificateOptions.SendCertificateOverMtls` (.NET #5849 client-credentials + #6009 OBO/refresh/auth-code) to msal-go.

**What it does:** presents the SN/I certificate as the client TLS certificate on the handshake to the token endpoint, routes to the `mtlsauth.*` endpoint, still sends the `private_key_jwt` client assertion in the body (ESTS requires it for this preview), and returns a **plain Bearer** access token. The certificate authenticates the transport; the token is **not** bound to it — `token_type` stays `Bearer`, cached under the standard key (not thumbprint-fenced).

Distinct from:
- **mTLS PoP** (`WithMtlsProofOfPossession`) — binds the token (`token_type=mtls_pop`), thumbprint-fenced cache. A per-request `WithMtlsProofOfPossession` always wins over this app-level flag.
- The existing SNI Bearer cell — signs the same `private_key_jwt`, but to the regular endpoint; the certificate is never on the TLS handshake. The assertion is not what distinguishes the two; the endpoint and the cert-on-handshake are.

### API
`confidential.WithSendCertificateOverMtls()` — app-level `Option` passed to `confidential.New(...)`. Defaults off (zero behavior change when unset). Requires a certificate credential (`NewCredFromCert`); `New()` errors otherwise. Honored by every confidential flow: client credentials, on-behalf-of, authorization code, and silent refresh.

### Implementation
- New `authority.AuthParams.MtlsTransport` flag, independent of `IsMtlsPoP`.
- `accesstokens.doTokenResp` (the single sink for all confidential flows) gates the endpoint rewrite + cert-on-handshake on `IsMtlsPoP || MtlsTransport`. The fail-closed `token_type=mtls_pop` guard stays `IsMtlsPoP`-only, so Bearer-over-mTLS receives its plain Bearer.
- x5c is forced onto the `private_key_jwt` client assertion (mirrors .NET's `Mode=OAuth`); the result carries no binding certificate.

### Tests (RED-first)
- **Unit** (`confidential_bearer_mtls_test.go`, mocked HTTP): builder validation; CC global / regional / mtls_pop-precedence; OBO; auth-code; refresh-via-silent + 2nd-call cache; no-flag negative.
- **Integration** (`bearer_mtls_integration_test.go`): CC live Bearer + cache-hit (allow-listed app); OBO/auth-code request-capture (assert endpoint+body, tolerate service reject); OBO/refresh live-acquire skip-gated (pending app mTLS-enablement, AADSTS700027/392189).

---

Targets `rginsburg/sni-mtls-pop` — sibling to the FIC PR, **not** stacked on it. Retarget to `main` after the SNI PR merges. The FIC test and its skip-hatch are untouched.

## Devex

```go
cred, _ := confidential.NewCredFromCert(certs, key)      // certificate credential REQUIRED
app, _ := confidential.New(authority, clientID, cred,
    confidential.WithSendCertificateOverMtls())          // app-level, not per-call

res, _ := app.AcquireTokenByCredential(ctx, scopes)
// res.Metadata.TokenType == "Bearer"   <- NOT bound to the certificate

// NEW STEP -- there ISN'T one. Using the token takes a single half:
req.Header.Set("Authorization", "Bearer "+res.AccessToken)
// res.BindingCertificate is nil: no tls.Config, no certificate on the resource
// call. Contrast #632, where the handshake is half of the credential.
```

The contrast with #632 is the whole point:

- The certificate authenticates the **transport** only. The request is routed to the
  mTLS endpoint (`mtlsauth.*`), but what comes back is a plain Bearer token — `token_type`
  stays `Bearer`, it is cached under the normal Bearer key, and the resource accepts it
  **without** the certificate being presented.
- Honored by client credentials, on-behalf-of, authorization code, and silent refresh.
  Not applied to `AcquireTokenByUsernamePassword` (sends no client credential) or
  `AcquireTokenByUserFederatedIdentityCredential`; those keep the regular token endpoint.
  Matches .NET's `SendCertificateOverMtls`.
- A per-request `WithMtlsProofOfPossession()` **always takes precedence**.
- `New` errors for any non-certificate credential.

Runnable demos live in `apps/tests/devapps/bearerovermtls/` and
`apps/tests/devapps/mtlscacheisolation/` — the latter shows that PoP and Bearer tokens
for the same client and scope occupy **separate cache entries**.


apps/tests/integration/bearer_mtls_integration_test.go
```diff
@@ -0,0 +1,297 @@
...
+func TestSendCertificateOverMtls_ClientCredential_Live(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping integration test")
+	}
+
+	cert, privateKey, err := getCertDataFromFile(pemFile)
+	if err != nil {
+		t.Fatalf("getCertDataFromFile() failed: %s", errors.Verbose(err))
+	}
+	cred, err := confidential.NewCredFromCert(cert, privateKey)
```
gladjohn, line 106: can this private key reside in keyguard? has it been tested?

Robbie-Microsoft: Yes, tested on a devbox with a real KeyGuard key (`Virtual Iso`=1, export blocked by the OS).

It didn't work initially: this flow still signs a `private_key_jwt` assertion, and golang-jwt's RSA methods require a concrete `*rsa.PrivateKey`, which a non-exportable key can never be. #647 lets a `crypto.Signer` into the credential (mTLS PoP only); #649 adds a `crypto.Signer`-backed `SigningMethod` so it can sign assertions too.

With #649 merged into this branch (clean merge), this flow returns a live token using the KeyGuard key. Classic SN/I works with it as well.

So no change needed here — this is unblocked once #649 lands.

gladjohn: Thanks, that explains the dependency. To confirm, #643 does not support KeyGuard on its own: it requires #647 to carry the crypto.Signer and #649 to sign the private_key_jwt. But the e2e test still uses an exportable key from getCertDataFromFile. how can we test this in the proper way? how will leg 1 in FIC flow handover cert to the second leg?

Robbie-Microsoft: Your summary is exactly right, and #643 alone can't even construct such a credential — `NewCredFromCert` still hard-asserts `*rsa.PrivateKey` on this branch (confidential.go:207). So a KeyGuard cell here would be untestable until #647/#649 land, not merely unexercised.

**Testing it properly.** KeyGuard is Windows + VBS; every pipeline we have is Linux (`go.yml`, `golangci-lint.yml`, and all three `ado/build_test.yaml` jobs are `ubuntu-latest`), so a real KeyGuard key cannot run in CI today. Three layers instead:

1. **Automated, cross-platform** — test the property rather than the hardware: a `crypto.Signer` that refuses to expose its inner key. It fails the moment anything in the path reaches for raw key material, which is precisely the defect that made #647/#649 necessary. These live in #647/#649 (and #633) and run on Linux CI. I mutation-checked them: forcing `*rsa.PrivateKey` back in fails them immediately.
2. **Manual gate, real hardware** — devbox, genuine KeyGuard key (`Virtual Iso`=1, export blocked by the OS): live token + Graph 200 over SN/I. Repeatable via scripted import; happy to attach the script and record it as a release gate.
3. **Automated on real hardware** would need a self-hosted Windows runner with VBS enabled — hosted VM runners generally can't do nested virtualization. That's infra, not code; better as a follow-up issue than a blocker here.

So: no change on #643, and once #647/#649 are in, (1) covers the regression risk continuously.

**FIC leg 1 -> leg 2 handoff:** done in d891024 on #633. `AuthResult.BindingCertificate` is now a `*tls.Certificate` carrying the key (incl. an opaque signer), and leg 2 takes it directly via `WithMtlsBindingTLSCertificate(leg1.BindingCertificate)`, or keeps the pair inseparable via `NewCredFromSignedAssertionCallback` returning `SignedAssertion{Assertion, BindingCertificate}` — the `ClientSignedAssertion` shape. Details in that thread.

apps/tests/integration/bearer_mtls_integration_test.go
```diff
@@ -0,0 +1,297 @@
...
+		t.Fatalf("NewCredFromCert() failed: %s", errors.Verbose(err))
+	}
+
+	rec := &recordingMtlsClient{}
+	// Region westus3 + the Key Vault scope is the allow-listed client-credentials configuration for live
+	// mTLS acquisition; the flag routes it to the regional mtlsauth endpoint.
+	app, err := confidential.New(sniAllowlistedAuthority, sniAllowlistedAppID, cred,
+		confidential.WithSendCertificateOverMtls(),
+		confidential.WithAzureRegion(sniAllowlistedRegion),
+		confidential.WithMtlsHTTPClient(recordingMtlsFactory(rec)),
```
gladjohn: The public API matches the .NET DevEx, but could the client-credentials live test exercise the **default mTLS transport** by removing:

```go
WithMtlsHTTPClient(recordingMtlsFactory(rec))
```

The .NET test only enables `SendCertificateOverMtls`; it doesn’t provide a custom transport.

Since `WithMtlsHTTPClient` is documented as an escape hatch, using it here means the live test doesn’t verify the primary customer path:

```go
app, err := confidential.New(
    authority,
    clientID,
    cred,
    confidential.WithSendCertificateOverMtls(),
    confidential.WithAzureRegion("westus3"),
)
```

The mocked tests already validate the endpoint and request body.

The live test could instead focus on proving that the SDK’s **default transport**:

* Automatically presents the certificate.
* Receives a plain `Bearer` token with no binding certificate.
* Serves the second request from the standard cache.

That would provide closer parity with the .NET integration test and help prevent regressions in the default transport path.

## What .NET Does Today

```csharp
[RunOn(SkipConditions.Linux)] // mTLS is not supported on Linux
public async Task Credential_X509_Output_Bearer_TestAsync()
{
    // X509 (SNI) credential over mTLS transport WITHOUT requesting PoP -> a plain Bearer token.
    // A Bearer token is not certificate-bound, so there is no Graph-over-mTLS call here.
    X509Certificate2 cert =
        CertificateHelper.FindCertificateByName(TestConstants.AutomationTestCertName);

    var certificateOptions = new CertificateOptions
    {
        SendCertificateOverMtls = true
    };

    IConfidentialClientApplication confidentialApp =
        ConfidentialClientApplicationBuilder
            .Create(MsiAllowListedAppIdforSNI)
            .WithAuthority(
                "https://login.microsoftonline.com/bea21ebe-8b64-4d06-9f6d-6a889b120a7c")
            .WithAzureRegion("westus3")
            .WithCertificate(cert, certificateOptions)
            .WithTestLogging()
            .Build();

    AuthenticationResult authResult = await confidentialApp
        .AcquireTokenForClient(new[] { "https://vault.azure.net/.default" })
        .ExecuteAsync()
        .ConfigureAwait(false);

    Assert.IsNotNull(
        authResult,
        "The authentication result should not be null.");

    Assert.AreEqual(
        "Bearer",
        authResult.TokenType,
        "Token type should be Bearer when PoP is not requested.");

    Assert.IsNotNull(
        authResult.AccessToken,
        "Access token should not be null.");
}
```

Robbie-Microsoft: Done in `4dba766`. The live client-credentials cell now configures nothing but the feature itself — `WithSendCertificateOverMtls()` + `WithAzureRegion(...)` — matching `Credential_X509_Output_Bearer_TestAsync`.

Your three points:

- **Presents the certificate** — a successful acquisition against mtlsauth is itself the proof; that endpoint won't issue to a handshake that carried none.
- **Plain Bearer, no binding cert** — asserts `Metadata.TokenType == "Bearer"` and `BindingCertificate == nil`.
- **Second request from cache** — asserts `Metadata.TokenSource == TokenSourceCache` and a token identical to the first.

`WithMtlsHTTPClient` now appears in this file only in comments explaining why it's absent. It remains in the mocked cells, where it injects mocks rather than standing in for the shipped path.

4gust (approved):

PR #644, opened 2026-08-02
Add user-assigned managed identity (UAMI) support for Azure Arc (MSAL.NET #6128 parity)

gladjohn:
## Summary

Port of MSAL.NET [#6128](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6128). Azure Arc is previewing support for user-assigned managed identity (UAMI). This enables it in MSAL Go and, critically, makes the SDK **fail closed** on agents that don't honor the request.

- Remove the client-side block in `New()` that rejected user-assigned identities on Azure Arc.
- Forward the user-assigned selector on the Arc token request (`createAzureArcAuthRequest`): `client_id` / `msi_res_id` / `object_id`, matching the other managed-identity sources.
- After a successful response, validate that the identity Azure Arc actually used matches the one requested (`verifyAzureArcUserAssignedIdentity`). A new agent echoes the used identity in the token response (`client_id` / `object_id`, and `msi_res_id` / `mi_res_id`); MSAL compares it case-insensitively to the requested selector and returns the token only on a match.
- **Fail closed** when the echo is missing or different: return an error, no token. A legacy agent silently returns the system-assigned identity for a UAMI request; without this check MSAL would hand back a token for the wrong identity.
- System-assigned is unaffected; a genuine `404` (identity not assigned to the machine) surfaces as a normal service/call error, not the fail-closed error.

Implementation note: `getTokenForRequest` now captures the raw response fields into `TokenResponse.AdditionalFields` (the typed `UnmarshalJSON` drops unknown fields), so the Arc verification can read the echoed identity.

## Unit tests (run in CI)

- `TestAzureArcUserAssignedAllowedAtConstruction` — construction no longer rejects UAMI on Arc.
- `TestAzureArcUserAssignedHonored` — happy path for client id / resource id / object id (echo matches → token returned; asserts the selector is on the request).
- `TestAzureArcUserAssignedNotHonoredFailsClosed` — no-echo and different-identity both fail closed with no token.
- `TestAzureArcUserAssignedResourceIDAcceptsMsiResIdEcho` — a resource id echoed as `msi_res_id` is accepted.
- `TestAzureArcUserAssignedNotFoundSurfacesServiceError` — a `404` (`identity_not_found`) surfaces as a service error, not "not supported".

## E2E (MI E2E – Azure Arc pipeline)

- `TestAzureArcUserAssignedNonExistentE2E` — a REAL acquisition for a non-existent UAMI on an actual Azure Arc machine; asserts no token is returned. Self-gates via `GetSource()` (skips off-Arc). Mirrors .NET's `AcquireToken_ForNonExistentUami_OnAzureArc_Fails`.

## Validation

The fail-closed path was validated end-to-end for MSAL.NET on a real Azure Arc machine running a legacy Connected Machine agent (1.54): the agent returns the system-assigned identity for a UAMI request, and MSAL rejects it instead of leaking it. MSAL Go uses the same HIMDS protocol, and the unit tests above reproduce both the honored and fail-closed behaviors.

## Notes

- **Do not merge yet**: the positive path (agent echoes the used identity → token returned) is not testable against any released Connected Machine agent (latest is 1.66; none implement the echo). This mirrors the gating on MSAL.NET #6128 and should merge once a preview agent with the echo contract is available.
- Field naming matches the agreed contract: the Arc **request** uses `msi_res_id` (a live Connected Machine agent silently ignores `mi_res_id` and returns the system-assigned identity); the response echo is read as both `msi_res_id` and `mi_res_id`.


apps/managedidentity/managedidentity.go
```diff
@@ -456,9 +451,57 @@ func (c Client) acquireTokenForAzureArc(ctx context.Context, resource string) (A
...
+func verifyAzureArcUserAssignedIdentity(id ID, token accesstokens.TokenResponse) error {
+	var requested, echoed string
+	switch t := id.(type) {
+	case UserAssignedClientID:
+		requested = string(t)
+		echoed = additionalStringField(token.AdditionalFields, miQueryParameterClientId)
+	case UserAssignedResourceID:
+		requested = string(t)
+		// The request selector is mi_res_id; accept either spelling on the echo.
+		echoed = additionalStringField(token.AdditionalFields, miQueryParameterResourceId, miQueryParameterResourceIdIMDS)
```
Avery-Dunn: Since it's now covering Azure Arc scenarios, we might want to give the "miQueryParameterResourceIdIMDS" constant a more generic name.

gladjohn: Good call - renamed to `miQueryParameterMsiResourceId` in 01e8401. It names the value (the `msi_res_id` spelling) rather than a source, which reads well next to `miQueryParameterResourceId = \"mi_res_id\"` and no longer implies IMDS-only now that `verifyAzureArcUserAssignedIdentity` accepts it as one of the two resource-id spellings Azure Arc may echo. Renamed all usages across managedidentity.go and the tests; build/vet/unit tests are green.

apps/tests/e2e/managedidentity_arc_e2e_test.go
```diff
@@ -29,3 +34,29 @@ import (
...
+	defer cancel()
+
+	result, err := client.AcquireToken(ctx, armResource)
+	if err == nil {
+		t.Fatalf("expected a 404 for a non-existent user-assigned identity, got token source %v", result.Metadata.TokenSource)
+	}
+	if result.AccessToken != "" {
+		t.Fatal("no token should be returned for a non-existent user-assigned identity")
+	}
+	if !strings.Contains(err.Error(), "404") && !strings.Contains(strings.ToLower(err.Error()), "identity_not_found") {
```
4gust: This test only passes if the error says "404" or "identity_not_found". But on an Arc agent that doesn't support user-assigned identities yet, the request returns the system-assigned identity with no echo, so the code fails closed with "did not confirm the requested user-assigned managed identity" instead. That message has neither string, so this test would fail on the current agent even though the code did the right thing. Both outcomes mean no token was returned, so can we also accept the fail-closed message here, and fix the comment that says it works on every agent?

apps/managedidentity/managedidentity.go
```diff
@@ -456,9 +451,57 @@ func (c Client) acquireTokenForAzureArc(ctx context.Context, resource string) (A
...
+func verifyAzureArcUserAssignedIdentity(id ID, token accesstokens.TokenResponse) error {
+	var requested, echoed string
+	switch t := id.(type) {
+	case UserAssignedClientID:
+		requested = string(t)
+		echoed = additionalStringField(token.AdditionalFields, miQueryParameterClientId)
+	case UserAssignedResourceID:
+		requested = string(t)
+		// The request selector is mi_res_id; accept either spelling on the echo.
+		echoed = additionalStringField(token.AdditionalFields, miQueryParameterResourceId, miQueryParameterMsiResourceId)
```
4gust: This assumes the agent echoes the identity back in the same form it was requested (a resource id request expects a resource id echo, an object id request expects an object id echo). If the real agent always echoes client_id no matter how you asked, then resource id and object id requests would fail closed even when they were honored. Can we confirm the echo contract before this merges so we don't reject valid requests later?

apps/tests/e2e/managedidentity_arc_e2e_test.go
gladjohn: Fixed in bfbe842. Verified on a real Azure Connected Machine agent (1.67, UAMI preview): a non-existent user-assigned identity returns no token but the surfaced error differs by agent — a UAMI-aware agent returns 404 (identity_not_found), a legacy agent ignores the selector and returns the system-assigned identity which MSAL rejects (fail closed). The test now accepts either error and the comment no longer claims every agent returns 404.

apps/managedidentity/managedidentity.go
gladjohn: Confirmed the echo contract on a real Arc agent (1.67, UAMI preview): the agent echoes back the SAME selector type that was requested — a client_id request echoes client_id, an object_id request echoes object_id, and a resource-id request echoes the resource id. It does NOT always echo client_id, so resource-id/object-id requests are not falsely rejected. One real bug surfaced and is now fixed (46d372c): the resource-id request must be sent as msi_res_id, not mi_res_id — the agent silently ignores mi_res_id and returns the system-assigned identity. The verify side still accepts either spelling on the echo as a safety net.

apps/managedidentity/managedidentity.go
```diff
@@ -456,9 +451,77 @@ func (c Client) acquireTokenForAzureArc(ctx context.Context, resource string) (A
...
 
+// setUserAssignedQueryParam adds the user-assigned identity selector to params. Azure Arc and IMDS
+// both use the IMDS "msi_res_id" spelling for the resource id; a system-assigned identity adds
+// nothing. An unsupported ID type is rejected (New already validates the caller's input).
+func setUserAssignedQueryParam(params url.Values, id ID) error {
+	switch t := id.(type) {
+	case UserAssignedClientID:
+		params.Set(miQueryParameterClientId, string(t))
+	case UserAssignedResourceID:
+		params.Set(miQueryParameterMsiResourceId, string(t))
```
Avery-Dunn, line 468: Probably just a minor issue in the PR description, but just to be clear: we do need to send `miQueryParameterMsiResourceId`/`msi_res_id` in the request, right? Not `miQueryParameterResourceId`/`mi_res_id`?

Because the PR description says:
> Forward the user-assigned selector on the Arc token request (createAzureArcAuthRequest): client_id / **mi_res_id** / object_id, matching the other managed-identity sources.

Avery-Dunn (approved):

apps/managedidentity/managedidentity.go
gladjohn: You're correct — the **request** sends `miQueryParameterMsiResourceId` / `msi_res_id` (as in the diff here, `params.Set(miQueryParameterMsiResourceId, string(t))`), not `mi_res_id`. A live Connected Machine agent silently ignores `mi_res_id` for the resource-id selector and returns the system-assigned identity, so `msi_res_id` is required.

The PR description was stale (it predated that fix) — I've corrected it: the request bullet now reads `client_id` / `msi_res_id` / `object_id`, and the Notes clarify the request uses `msi_res_id` while the response echo is read as both spellings (`msi_res_id` / `mi_res_id`) as a safety net. Thanks for the catch!

PR #646, opened 2026-08-06
Harden PEM and region validation

4gust:
## Summary
- retain legacy RFC 1423 PEM support with a clear runtime migration warning; encrypted PKCS#8 remains unsupported
- validate explicit regions as lowercase ASCII DNS labels and ignore invalid auto-detected environment/IMDS regions with global fallback
- add regression coverage and update stale design documentation

## Validation
- go test ./apps/confidential -count=1
- go test ./apps/internal/oauth/ops/authority -count=1


apps/internal/oauth/ops/authority/authority.go
```diff
@@ -717,8 +717,10 @@ func (c Client) AADInstanceDiscovery(ctx context.Context, authorityInfo Info) (I
 func detectRegion(ctx context.Context) string {
 	region := os.Getenv(regionName)
 	if region != "" {
-		region = strings.ReplaceAll(region, " ", "")
-		return strings.ToLower(region)
+		if validRegion.MatchString(region) {
```
Avery-Dunn, line 720: The previous code had some normalization (`strings.ReplaceAll(region, " ", "")` to remove spaces, `strings.ToLower(region)` to make it lowercase), is there a reason we don't want to do that anymore?

apps/confidential/confidential.go
```diff
@@ -77,8 +77,10 @@ const (
 )
 
 // CertFromPEM converts a PEM file (.pem or .key) for use with [NewCredFromCert]. The file
-// must contain the public certificate and the private key. If a PEM block is encrypted and
-// password is not an empty string, it attempts to decrypt the PEM blocks using the password.
+// must contain the public certificate and an unencrypted PKCS#1 or PKCS#8 private key.
+// For security, encrypted PEM blocks are not supported; password is retained for API compatibility
```
Avery-Dunn: Since the previous behavior allowed an encrypted key+password, wouldn't this be a breaking change to whoever is still passing in encrypted keys?

If so, new runtime errors caused solely by updating their MSAL version isn't a great developer experience for our customers. I don't really know the conventions for Go when it comes to breaking changes/deprecation like this, but do either of these seem like a good option?
1) If the password option is truly a dead end now, we could remove it from the signature: a very clear compile-time break might be preferable over breaking only during testing or after deployment
2) We could add a new non-password version of the API with loud and clear deprecation warnings on the existing API, and then actually remove the old API in a version or two

4gust: Updated

Avery-Dunn: And how do these changes interact with the changes on this other recent PR that also adjusted the `CertFromPEM` function? https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/645

Just want to make sure nothing got reverted unexpectedly in the merge and we have some other breaking change.

4gust: I created a base function that is then wrapped so that we can mark the function as deprecated. Create a new wrapper that is not default.
If other PR modifies anything will be easy to do that in base function.

apps/internal/oauth/ops/authority/authority.go
Avery-Dunn: Resolved in a separate discussion, normalization should not be needed.

apps/confidential/confidential_legacy_pem_test.go
```diff
@@ -1,155 +0,0 @@
-// Copyright (c) Microsoft Corporation.
```
Avery-Dunn, line 1: Why delete this test class that was recently added by https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/645?

4gust:
Superseded by the focused replacements: #650 (regional authority validation) and #651 (PEM API migration).

PR #647, opened 2026-08-06
Support non-exportable keys (KeyGuard/CNG/HSM) for mTLS PoP

Robbie-Microsoft:
Stacked on #632 — base is `rginsburg/sni-mtls-pop`. One feature commit, plus merge commits bringing in base (which now carries `AuthResult.BindingCertificate` as a `*tls.Certificate`, and main through #645).

## What

`confidential.NewCredFromTLSCertificate(tls.Certificate) (Credential, error)` — `PrivateKey` may be any `crypto.Signer`, so non-exportable keys (KeyGuard/VBS, CNG, HSM) work with mTLS PoP. Validates `signer.Public()` against the leaf and preserves the caller's leaf-first chain in x5c.

## Why it's needed

`NewCredFromCert` asserts `key.(*rsa.PrivateKey)`. A non-exportable key can only ever surface as a `crypto.Signer`, so it's rejected at construction. **Pre-existing in `main`, not introduced by #632.** `WithMtlsHTTPClient` doesn't rescue it — `resolveMtlsBindingCert()` still requires `cred.Cert`/`cred.Key`.

mTLS PoP is the only flow that can support these keys: `FromClientCertificate` sends no `client_assertion` and no `req_cnf`, so the key is used only in the TLS handshake, and `crypto/tls` needs only a `crypto.Signer` (PKCS#1 v1.5 on 1.2, RSA-PSS on 1.3). `resolveMtlsBindingCert` → `MtlsBindingCert` → `BuildMtlsClient` already treat `Key` as opaque, so no transport change was required.

## Guardrail

Signer-only credentials are flagged `signerOnly` → `accesstokens.Credential.SignerOnly`, gated at the top of `Credential.JWT()`, which every assertion-signing flow funnels through (client credentials, OBO, refresh token, auth code, user FIC, ADFS/dSTS RS256):

> this credential's private key is not exportable and can only be used with WithMtlsProofOfPossession()

## Please weigh this part separately

I also relaxed **`NewCredFromCert`** to accept a `crypto.Signer` with an RSA public key. It changes an existing exported constructor, so it's a distinct decision and can be dropped independently — `NewCredFromTLSCertificate` is the priority. Non-regressing: `nil`, non-signers, `*ecdsa.PrivateKey` and `ed25519.PrivateKey` all still fail with the identical `"key must be an RSA key"`, and `*rsa.PrivateKey` still matches its concrete case first.

## Out of scope: private_key_jwt

golang-jwt v5's `SigningMethodRSAPSS`/`SigningMethodRSA` also require a concrete `*rsa.PrivateKey`, so classic SN/I additionally needs a `crypto.Signer`-backed `jwt.SigningMethod`. Not attempted here; the gate above turns that limitation into a clear error instead of a confusing one.

## Validation

**Live token from ESTS using a real Windows KeyGuard key** (NCrypt signer, `Virtual Iso`=1, export blocked at the OS): `token_type: mtls_pop`, Graph returned 200 on the mTLS endpoint, cache hit on repeat. Reproduced.

Committed tests are cross-platform and need no Windows or KeyGuard — a fake signer wraps an ephemeral RSA key without ever exposing it. `TestBuildMtlsClientSignerHandshake` performs a real TLS 1.2 and 1.3 client-certificate handshake against an `httptest` server through the built-in transport, reproducing the devbox result in CI. Chain-order and x5c assertions are mutation-checked (dropping intermediates, reversing order, and a bogus signature each fail the suite).

`go build ./...`, `go vet ./...`, `go test ./apps/confidential/... ./apps/internal/...` all pass.

## Notes for reviewers

- **+100/-15 non-test.** The rest of the diff is tests (+464) and docs/examples (+81).
- No new module dependencies, no platform-specific code. `go.mod`/`go.sum` untouched; every new API is ≤ Go 1.16. The NCrypt signer lives in caller code, not in MSAL.
- `WithMtlsHTTPClient` and `comm.MtlsClientFactory` documented themselves as the escape hatch "for keys the built-in transport cannot use (for example CNG/HSM-backed keys)". That premise is now false, so they're reworded to "for callers who must own the TLS handshake themselves".
- `confidential_signer_test.go` compares `res.BindingCertificate.Leaf.Equal(certs[0])` and asserts `PrivateKey != nil`: base now returns a `*tls.Certificate` (which has no `Equal` method), and the private key surviving to the caller is exactly what this PR is about.
- README gains a "Non-exportable keys (KeyGuard, CNG, HSM)" section, plus `ExampleNewCredFromTLSCertificate`.

## Devex

Steps 1–2 of the usual cert flow **disappear**: there is no `os.ReadFile`, no
`CertFromPEM`, because there is no private key to load.

```go
// ncryptsigner is a SAMPLE under apps/tests/devapps/, not public API.
signer, _ := ncryptsigner.Open("CurrentUser", "My", thumbprint)
defer signer.Close()

if iso, _ := signer.IsVirtualIsolated(); !iso { /* fail: not a VBS-isolated key */ }

    Certificate: signer.Chain(),   // leaf-first DER; self-signed root dropped (x5c convention)
    PrivateKey:  signer,           // any crypto.Signer
})

res, _ := app.AcquireTokenByCredential(ctx, scopes,
    confidential.WithMtlsProofOfPossession())   // MANDATORY, not an optimization

// NEW STEP -- using the token takes both halves, exactly as in #632:
tls.Config{Certificates: []tls.Certificate{*res.BindingCertificate}}
req.Header.Set("Authorization", "mtls_pop "+res.AccessToken)
// res.BindingCertificate.PrivateKey is STILL the CNG signer, so the resource
// handshake is signed inside the VBS boundary too -- the key never leaves it.
```

- The **thumbprint is a public lookup key** (SHA-1 of the cert DER), not the key handle.
  `Open` resolves it via `CertFindCertificateInStore(CERT_FIND_HASH)` →
  `CryptAcquireCertificatePrivateKey(ONLY_NCRYPT)`. `ONLY_NCRYPT` is required: a
  VBS-isolated key has no legacy CAPI representation.
- Omit `WithMtlsProofOfPossession()` and you hit the signer-only gate at the top of
  `Credential.JWT()` — *"this credential's private key is not exportable and can only be
  used with WithMtlsProofOfPossession()"*. Every JWT signing path needs an
  `*rsa.PrivateKey`, which a KeyGuard key can never be.
- **`signerOnly` is set from the Go type of the key alone.** MSAL cannot distinguish a
  VBS-isolated key from a software key behind a `crypto.Signer`, and makes **no isolation
  claim** either way. Applications must call `IsVirtualIsolated()` themselves.
- Nothing in MSAL is Windows-specific and no API was added for KeyGuard; `crypto.Signer`
  is the entire integration surface.

A runnable sample and E2E live in `apps/tests/devapps/keyguard/` and
`apps/tests/e2e/keyguard_mtls_e2e_test.go`, both against a real VBS-isolated key.


apps/confidential/confidential.go
```diff
@@ -1073,9 +1142,12 @@ func WithAttribute(attrValue string) interface {
 // certificate. The authority must be tenanted (not /common, /organizations, or /consumers) and in a
 // supported cloud.
 //
-// The binding certificate is inferred from a [NewCredFromCert] credential. The result exposes the
-// public binding certificate via [AuthResult.BindingCertificate] and its thumbprint via
-// [AuthResult.BindingCertificateThumbprint]; the private key is never surfaced.
+// The binding certificate is inferred from a [NewCredFromCert] or [NewCredFromTLSCertificate]
+// credential. The result exposes the public binding certificate via [AuthResult.BindingCertificate]
+// and its thumbprint via [AuthResult.BindingCertificateThumbprint]; the private key is never surfaced.
```
gladjohn: It is only partial parity with MSAL.NET. The PR achieves parity for acquiring an mtls_pop token using a non-exportable signing key, but not for the complete developer experience of taking the returned binding certificate and calling AKV over mTLS.

Robbie-Microsoft: Closed that gap. `AuthResult.BindingCertificate` is now a `*tls.Certificate` carrying the parsed leaf and the (possibly non-exportable) private key, so it drops straight into `tls.Config.Certificates`. Landed on #632 as d8e8619 and merged through the stack; the E2E now presents it verbatim to Graph over mTLS.

gladjohn (approved):

gladjohn:
can you please explain this comment in the PR desc 

```
I also relaxed NewCredFromCert to accept a crypto.Signer with an RSA public key. It changes an existing exported constructor, so it's a distinct decision and can be dropped independently — NewCredFromTLSCertificate is the priority. Non-regressing: nil, non-signers, *ecdsa.PrivateKey and ed25519.PrivateKey all still fail with the identical "key must be an RSA key", and *rsa.PrivateKey still matches its concrete case first.
```

apps/confidential/confidential.go
gladjohn: Please do not resolve until we can validate the full e2e

apps/confidential/confidential.go
```diff
@@ -230,6 +252,52 @@ func NewCredFromCert(certs []*x509.Certificate, key crypto.PrivateKey) (Credenti
...
+// is used when it matches that leaf; otherwise the leaf is parsed from cert.Certificate[0].
+func NewCredFromTLSCertificate(cert tls.Certificate) (Credential, error) {
+	if len(cert.Certificate) == 0 || len(cert.Certificate[0]) == 0 {
+		return Credential{}, errors.New("tls.Certificate must contain at least one certificate")
+	}
+	signer, ok := cert.PrivateKey.(crypto.Signer)
+	if !ok {
+		return Credential{}, errors.New("tls.Certificate.PrivateKey must implement crypto.Signer")
+	}
+	leaf := cert.Leaf
```
gladjohn: We trust and retain caller-owned `cert.Leaf` when only its `Raw` matches the DER, allowing a self-inconsistent Leaf to bypass cert/key validation and remain mutable; please parse and retain an MSAL-owned leaf from `Certificate[0]`.

apps/confidential/confidential.go
```diff
@@ -230,6 +252,52 @@ func NewCredFromCert(certs []*x509.Certificate, key crypto.PrivateKey) (Credenti
...
+	if leaf == nil || !bytes.Equal(leaf.Raw, cert.Certificate[0]) {
+		var err error
+		leaf, err = x509.ParseCertificate(cert.Certificate[0])
+		if err != nil {
+			return Credential{}, fmt.Errorf("could not parse the leaf certificate: %v", err)
+		}
+	}
+	// every public key type x509 can parse implements Equal, added in Go 1.15
+	pub, ok := leaf.PublicKey.(interface{ Equal(crypto.PublicKey) bool })
```
gladjohn, line 349: A typed-nil `crypto.Signer` passes the interface assertion and panics at `signer.Public()`; please reject typed-nil signers and add a regression test so malformed credentials return an error.

apps/confidential/confidential.go
```diff
@@ -230,6 +252,52 @@ func NewCredFromCert(certs []*x509.Certificate, key crypto.PrivateKey) (Credenti
...
+	if !ok || !pub.Equal(signer.Public()) {
+		return Credential{}, errors.New("key doesn't match the leaf certificate")
+	}
+	cred := Credential{cert: leaf, key: cert.PrivateKey}
+	// tls.Certificate stores the chain leaf first, which is the order x5c requires
+	for _, der := range cert.Certificate {
+		cred.x5c = append(cred.x5c, base64.StdEncoding.EncodeToString(der))
+	}
+	if _, ok := cert.PrivateKey.(*rsa.PrivateKey); !ok {
+		cred.signerOnly = true
```
gladjohn, line 359: `signerOnly` is set for every non-`*rsa.PrivateKey` key but described as “non-exportable,” which is false for exportable ECDSA keys or RSA signer wrappers; please rename/reword it to describe JWT compatibility rather than exportability.

apps/confidential/confidential.go
```diff
@@ -230,6 +252,52 @@ func NewCredFromCert(certs []*x509.Certificate, key crypto.PrivateKey) (Credenti
...
+		if err != nil {
+			return Credential{}, fmt.Errorf("could not parse the leaf certificate: %v", err)
+		}
+	}
+	// every public key type x509 can parse implements Equal, added in Go 1.15
+	pub, ok := leaf.PublicKey.(interface{ Equal(crypto.PublicKey) bool })
+	if !ok || !pub.Equal(signer.Public()) {
+		return Credential{}, errors.New("key doesn't match the leaf certificate")
+	}
+	cred := Credential{cert: leaf, key: cert.PrivateKey}
```
gladjohn, line 353: Please document that MSAL retains the supplied signer, may invoke `Sign` concurrently through its cached mTLS client, and requires the signer/native handle to remain valid for the Client and any returned binding certificate.

apps/confidential/confidential_signer_test.go
```diff
@@ -0,0 +1,345 @@
...
+				}
+			}
+		})
+	}
+}
+
+// TestNewCredFromTLSCertificateChainOrder verifies a chain that isn't leaf first is rejected rather
+// than silently producing a credential whose x5c and binding certificate disagree with the leaf
+// actually presented on the handshake.
+func TestNewCredFromTLSCertificateChainOrder(t *testing.T) {
```
gladjohn: `TestNewCredFromTLSCertificateChainOrder` validates only that the leaf is first, not intermediate ordering or chain validity; please rename it accordingly or add the missing coverage.

apps/confidential/confidential.go
```diff
@@ -230,6 +252,52 @@ func NewCredFromCert(certs []*x509.Certificate, key crypto.PrivateKey) (Credenti
...
+	signer, ok := cert.PrivateKey.(crypto.Signer)
+	if !ok {
+		return Credential{}, errors.New("tls.Certificate.PrivateKey must implement crypto.Signer")
+	}
+	leaf := cert.Leaf
+	if leaf == nil || !bytes.Equal(leaf.Raw, cert.Certificate[0]) {
+		var err error
+		leaf, err = x509.ParseCertificate(cert.Certificate[0])
+		if err != nil {
+			return Credential{}, fmt.Errorf("could not parse the leaf certificate: %v", err)
```
gladjohn: This parse error uses `%v` and loses the underlying error chain; please use `%w` for consistency and `errors.Is`/`errors.As` support.

gladjohn (changes requested):
Requesting changes because `NewCredFromTLSCertificate` retains caller-owned mutable `Leaf` state instead of treating the DER certificate as authoritative, so the validated cert/key invariant can be invalidated after construction and the same parsed object is shared across credential/result boundaries. The new constructors can also panic on typed-nil `crypto.Signer` values. Please make parsed certificate state MSAL-owned, reject typed-nil signers, add a real TLS test across the constructor-to-transport pipeline, and clarify the signer lifetime/concurrency and `SignerOnly` contracts.

apps/confidential/confidential.go
Robbie-Microsoft: Agreed, fixed in `6913279`. `cert.Leaf` is now ignored; the leaf is always parsed from `cert.Certificate[0]`.

`bytes.Equal` on `Raw` proved the DER matched but not that the parsed fields did, so a hand-built `x509.Certificate` with a correct `Raw` and a bogus `PublicKey` would have been used for the key-match check below and then retained as the credential's certificate for the thumbprint and binding cert. One parse per credential at construction is the right trade for MSAL owning the leaf.

apps/confidential/confidential.go
Robbie-Microsoft: Done in `6913279`.

apps/confidential/confidential.go
Robbie-Microsoft: Good catch, and it was worse than flagged — `NewCredFromCert` had the identical exposure at its `case crypto.Signer:` branch, which this PR introduced when it relaxed that constructor. Both are guarded in `6913279` by a shared `isNilSigner`, which mirrors the existing `isNilClient` in `comm/mtls.go`, so it's not a new pattern here.

`TestNewCredFromTypedNilSigner` covers both constructors. Mutation-checked: with the guard removed both subtests fail with `runtime error: invalid memory address or nil pointer dereference`; restored, both pass.

apps/confidential/confidential.go
Robbie-Microsoft: Documented in `6913279`. Three things a CNG/HSM caller needs that weren't stated:

- the `Credential` retains the signer for its lifetime and never copies it or extracts private material;
- `Sign` can be called concurrently, since MSAL caches one mTLS client per confidential client and handshakes can overlap, so the signer must be safe for concurrent use;
- the signer must stay usable at least as long as the `Client`, and in practice longer, because `AuthResult.BindingCertificate` hands the same key back for the caller's own mTLS call to the resource — closing it once a token is acquired breaks that call.

Cross-referenced from `NewCredFromCert` rather than duplicated.

apps/confidential/confidential.go
Robbie-Microsoft: You're right about the doc. The flag keys off "not an `*rsa.PrivateKey`", which also catches an exportable ECDSA key or a signer wrapping ordinary software RSA — exportability was never the trigger. Reworded in `6913279` to state the actual capability: MSAL can't sign a client assertion with the key, so the credential is limited to mTLS PoP, whatever backs it.

I left the field name alone deliberately. It's plumbed through as `accesstokens.Credential.SignerOnly` and referenced throughout the PR stacked on top of this one, so renaming here means doing it twice for no behavior change. Happy to rename it there if you'd still like it.

apps/confidential/confidential_signer_test.go
Robbie-Microsoft: Fair — renamed to `TestNewCredFromTLSCertificateRejectsNonLeafFirstChain` in `6913279`, which is all it ever asserted.

Broader chain-order coverage does already exist, in `TestNewCredFromTLSCertificate` (`confidential_signer_test.go:109-114`): for every `i`, `cred.x5c[i] == base64(certs[i].Raw)`, so the whole chain is pinned per-index in the caller's leaf-first order, under both the `leaf` set and unset subtests. Chain *validity* is deliberately out of scope — x5c is passed through and the service validates it.

apps/confidential/confidential.go
Robbie-Microsoft: Only .NET's IMDSv2 E2E tests auth to AKV, and they're Managed Identity on the MSALMSIV2 pool against `tokenbinding.vault.azure.net`, which isn't resolvable outside that pool. That test also treats a non-200 from AKV as `Assert.Inconclusive`, so it isn't gating there either.

Go's SNI-cert-over-mTLS E2E auths to Graph, same as .NET's SNI mTLS tests. We've validated the full E2E over Graph — KeyGuard key confirmed VBS-isolated, mtls_pop token, `cnf` verified against the leaf, resource 200, 401 without the cert. Just not AKV.

Robbie-Microsoft:
> can you please explain this comment in the PR desc
> 
> ```
> I also relaxed NewCredFromCert to accept a crypto.Signer with an RSA public key. It changes an existing exported constructor, so it's a distinct decision and can be dropped independently — NewCredFromTLSCertificate is the priority. Non-regressing: nil, non-signers, *ecdsa.PrivateKey and ed25519.PrivateKey all still fail with the identical "key must be an RSA key", and *rsa.PrivateKey still matches its concrete case first.
> ```

NewCredFromCert required exactly a *rsa.PrivateKey. KeyGuard/CNG/HSM keys can't produce one — the private material never leaves the isolated environment, so you only get a crypto.Signer. The relaxation adds a crypto.Signer case whose Public() is an *rsa.PublicKey.

Flagged separately because it widens a shipped exported constructor; NewCredFromTLSCertificate is new API with no compat surface. Drop that case and nothing else in the PR moves.

Non-regressing. *rsa.PrivateKey also satisfies crypto.Signer, but its concrete case is listed first, so existing callers keep the old path with signerOnly=false instead of being marked signer-only. nil, non-signers, *ecdsa.PrivateKey, ed25519.PrivateKey, and signers with EC public keys all still fail with "key must be an RSA key".

Found and fixed while checking: a typed-nil *rsa.PrivateKey panicked. It hits the concrete case before the crypto.Signer case isNilSigner guards, so it needed its own nil check. Predates this PR — main asserts key.(*rsa.PrivateKey), which a typed nil also satisfies, then derefs. Guarded in 9352d5d, with a test covering both constructors.

4gust (approved):

PR #649, opened 2026-08-10
Sign client assertions with a non-exportable key (crypto.Signer)

Robbie-Microsoft:
Adds an internal `jwt.SigningMethod` that signs through a `crypto.Signer`, so a non-exportable key (Windows KeyGuard/VBS-isolated, CNG, HSM) can sign the `private_key_jwt` client assertion.

Depends on #647, which is the base branch. #647 lets a `crypto.Signer` into a `Credential` but gated it out of every assertion-signing flow; this replaces that gate.

**What changes**

- `accesstokens/signingmethod.go`: `signerMethod` wraps the built-in method `Credential.JWT` already selected. Only `Sign` is new — hash the signing string, delegate to `signer.Sign`. `Alg()` and `Verify()` delegate unchanged, so the `alg` header, `thumbprint(cert, alg)` and verification are untouched. PS256 passes `&rsa.PSSOptions{SaltLength: rsa.PSSSaltLengthEqualsHash, Hash: crypto.SHA256}` to match `jwt.SigningMethodPS256` exactly (not `PSSSaltLengthAuto`, which `SigningMethodPS256.VerifyOptions` would happily accept); RS256 passes the bare `crypto.SHA256` for PKCS #1 v1.5. Constructed per call — no `jwt.RegisterSigningMethod`, so golang-jwt's package-level method map is untouched for other consumers in the process.
- `Credential.JWT`: the unconditional early return becomes a method swap when `SignerOnly` is set.
- The gate survives for a signer whose `Public()` isn't an `*rsa.PublicKey`. `NewCredFromTLSCertificate` validates the signer against the leaf, not against RSA, so an ECDSA/Ed25519 leaf plus matching signer constructs fine; the service takes only RSA assertions, so that case still errors, now with `"this credential's private key must be RSA to sign a client assertion"`.

**Impact**

Fixes `private_key_jwt` for KeyGuard/CNG/HSM keys across every assertion flow: client credentials, OBO, authorization code, refresh token, user FIC, and ADFS/dSTS. Unblocks #643 (Bearer-over-mTLS) — that flow needs an assertion signed by the same non-exportable key it presents on the handshake.

Wire format is unchanged. No new dependencies; `go.mod`/`go.sum` untouched. No public API change, no platform-specific code, no build tags.

**Tests**

Cross-platform, no Windows or KeyGuard dependency — a fake signer wraps an ephemeral `*rsa.PrivateKey` and is not type-assertable to one. Assertions are verified against the certificate's public key rather than just checking `err == nil`, and PSS is verified with `SaltLength: PSSSaltLengthEqualsHash` so a salt-length regression can't slip through `PSSSaltLengthAuto` verification. Headers from a signer-only credential are compared byte-for-byte against an `*rsa.PrivateKey` credential's for the same input, across AAD/ADFS/dSTS and both `SendX5C` values.

Mutation-tested: 11 deliberate breakages (PSS salt length -> `Auto`, unset `PSSOptions.Hash`, RS256 opts -> PSS opts, corrupted signature, wrong digest input, hardcoded `Alg()`, `Verify` short-circuited to nil, RSA gate removed, `SignerOnly` gate bypassed/never taken, wrong method wrapped) were each caught.

`go build ./...`, `go vet ./...`, `go test ./apps/confidential/... ./apps/internal/...` and `golangci-lint run ./...` (v2.12, matching CI) all pass.

**Validation**

Validated end-to-end against a real Windows KeyGuard key (NCrypt `crypto.Signer`, `Virtual Iso`=1, export blocked by the OS) on a devbox, using the lab SN/I app:

| flow | before | with this PR |
|---|---|---|
| classic SN/I (`private_key_jwt`, x5c) | rejected at signing | live token |
| Bearer-over-mTLS (#643) | rejected at signing | live token |

The second row is the case Gladwin raised on #643 — that branch merges cleanly with this one and then issues a token with the non-exportable key.

## Devex

**No new exported API.** Same calling code as #647 — the restriction is what changes.

```go
    Certificate: chain, PrivateKey: signer})

res, _ := app.AcquireTokenByCredential(ctx, scopes)   // no mTLS PoP option needed
```

- #647's hard gate becomes a branch: a `crypto.Signer` can now sign client assertions,
  so a non-exportable key works with assertion-based flows and is no longer confined to
  mTLS PoP.
- `signerMethod` wraps the selected JWT method and delegates `Alg()` to it, so on the
  normal path **the assertion is byte-for-byte what an exportable key would produce**.
- **One case does change the wire format, deliberately:** if the signer cannot produce a
  PS256 signature — some CNG configurations, HSMs and smart cards cannot do RSA-PSS at
  all — MSAL retries once with PKCS#1 v1.5. That rebuilds the whole assertion rather than
  re-signing it, because the algorithm determines both the `alg` header *and* which
  thumbprint header is used: `PS256`/`x5t#S256` becomes `RS256`/`x5t`. Matches MSAL .NET.
  The fallback applies only to signer-backed keys, never to ADFS/dSTS (already RS256) or
  to exportable keys (Go's software RSA always supports PSS).
- **Zero exported functions added, removed or renamed** relative to #647, and no
  `// Deprecated` markers — the change is entirely internal.


apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -149,6 +145,26 @@ func (c *Credential) JWT(ctx context.Context, authParams authority.AuthParams) (
 		thumbprintKey = "x5t"
 	}
 
+	// A non-exportable key (Windows KeyGuard/CNG/HSM) can only ever be a crypto.Signer, which jwt's
+	// built-in RSA methods reject because they need an *rsa.PrivateKey. Wrap the method selected above
+	// in one that delegates to the signer. Alg() is unchanged, so the assertion's wire format is too.
+	if c.SignerOnly {
```
gladjohn: **Missing PS256→RS256 fallback for non-exportable signers.** For AAD this path hard-selects `PS256`, wraps it in `signerMethod`, and signs via `crypto.Signer`. If the signer can't perform RSA-PSS — common for the very CNG/KeyGuard/HSM/smartcard providers this PR targets — `signer.Sign(…, *rsa.PSSOptions)` fails and the whole acquisition errors out, with no fallback.

MSAL .NET handles exactly this in `JsonWebToken.Sign`: it signs PS256, and on `catch (CryptographicException) when (useSha2AndPss)` rebuilds the header as RS256 (`x5t`) and re-signs with PKCS#1 (_"Some private key providers do not support PSS"_). An exportable `*rsa.PrivateKey` never hits this (Go software RSA always does PSS), so the gap is specific to the signer path introduced here.

Suggestion: in the `SignerOnly` branch, on a PS256 sign failure rebuild the assertion with the RS256 header/thumbprint and re-sign via the RS256 signer method. A test with a signer that rejects `*rsa.PSSOptions` (asserting the assertion comes back `alg=RS256`) would lock it in.

Robbie-Microsoft: Done in `3061226`. Fallback rebuilds the header rather than re-signing — `alg` and `x5t#S256`→`x5t`. Gated to `SignerOnly && !isADFSorDSTS`: an exportable key is an `*rsa.PrivateKey` and Go's software RSA always does PSS, and ADFS/dSTS already start at RS256. If the retry also fails, both errors surface.

Test is `TestJWTSignerOnlyPS256Fallback` — signer rejects `*rsa.PSSOptions`, asserts `alg=RS256`, `x5t` == SHA-1 thumbprint, no `x5t#S256`, and exactly 1 PSS + 1 PKCS#1 attempt.

One divergence worth naming: `crypto.Signer` returns an opaque `error`, so there's no analog to your `when (useSha2AndPss)` filter — any PS256 failure on that path triggers the retry.

gladjohn:
This PR is stacked on open PR #647, which is stacked on open PR #632. The complete `main → #649` change spans 41 files and roughly 6,800 added lines, so #649 cannot be evaluated as an independent feature boundary. Can we collapse or rebase the stack before approval so the final security behavior is reviewable against `main`?

apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -149,23 +145,67 @@ func (c *Credential) JWT(ctx context.Context, authParams authority.AuthParams) (
...
+		return token.SignedString(c.Key)
 	}
 
-	assertion, err := token.SignedString(c.Key)
-	if err != nil {
+	assertion, err := build(signingMethod, thumbprintKey)
+	if err == nil {
+		return assertion, nil
+	}
+	if !c.SignerOnly || isADFSorDSTS {
```
gladjohn: We currently interpret every `crypto.Signer` failure as “PSS unsupported” and retry with RS256. A signer may instead fail because of permission denial, cancellation, HSM availability, key deletion, or another provider error. Please distinguish unsupported PSS where possible, or explicitly document and security-review the decision to downgrade after any PS256 error.

apps/confidential/confidential.go
```diff
@@ -157,8 +157,8 @@ type Credential struct {
 	key  crypto.PrivateKey
 	x5c  []string
 
-	// signerOnly is set when key can't be exported as an *rsa.PrivateKey, which restricts the
-	// credential to mTLS proof-of-possession. See NewCredFromTLSCertificate.
+	// signerOnly is set when key can't be exported as an *rsa.PrivateKey, which means MSAL must sign
+	// through its crypto.Signer interface. See NewCredFromTLSCertificate.
 	signerOnly bool
```
gladjohn: Now that signer-backed credentials can sign normal assertions, is `signerOnly` still necessary? The required path is derivable from `key`: use the existing implementation for `*rsa.PrivateKey`; otherwise require `crypto.Signer`. Keeping a separate boolean duplicates state across both credential layers and permits inconsistent combinations of `Key` and `SignerOnly`.

apps/confidential/confidential.go
```diff
@@ -255,14 +253,12 @@ func NewCredFromCert(certs []*x509.Certificate, key crypto.PrivateKey) (Credenti
...
-// CNG/HSM-backed keys: the signer stays in its protected store and is invoked only to sign the TLS
-// handshake.
+// CNG/HSM-backed keys: the signer stays in its protected store and is invoked only to sign.
 //
-// Such a credential is limited to mTLS proof-of-possession (see [WithMtlsProofOfPossession]), because
-// that's the only flow in which the key isn't used to sign a client assertion: the TLS client
-// certificate authenticates the client and binds the token, so no client_assertion is sent. Any other
-// flow returns an error explaining the restriction. When cert.PrivateKey is an *rsa.PrivateKey the
-// credential has no such restriction and behaves like one from [NewCredFromCert].
+// The credential works in every flow. mTLS proof-of-possession (see [WithMtlsProofOfPossession])
```
gladjohn: “Works in every flow” is broader than the new feature-level coverage. The signer-backed request test exercises client credentials, while OBO, authorization code, refresh token, and user FIC rely on shared `JWT()` wiring. Please add a small table-driven request-construction test proving the signer credential produces `client_assertion` through each claimed public flow.

README.md
```diff
@@ -235,24 +235,31 @@ if err != nil {
...
+- **Every flow is supported**: MSAL signs the `client_assertion` through the signer, so
+  `private_key_jwt`/SN/I works for client credentials, on-behalf-of, authorization code, refresh
+  token, and ADFS/dSTS. `WithMtlsProofOfPossession()` is the one flow that sends no
+  `client_assertion` at all — the key is used solely for the TLS handshake and the token is bound to
+  the certificate. The signer's public key must be an `*rsa.PublicKey` to sign an assertion, because
+  the service accepts only RSA assertions; a non-RSA signer is limited to mTLS PoP.
 - **No extra transport work is needed**: `crypto/tls` signs the handshake through the signer on both
   TLS 1.2 (PKCS#1 v1.5) and TLS 1.3 (RSA-PSS), so the built-in mTLS transport handles these keys.
   `WithMtlsHTTPClient` remains available if you need to own the handshake for other reasons.
 - **The signer lives in your code**: MSAL adds no platform-specific dependencies. Implement
```
gladjohn: The MSAL API is clear once the application already has a `crypto.Signer`, but obtaining that signer is the difficult part for KeyGuard/CNG and isn’t demonstrated here. Can we link a supported NCrypt signer example, or scope this wording to “supports caller-provided `crypto.Signer` implementations” rather than implying an out-of-box KeyGuard developer experience?

apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -149,23 +145,67 @@ func (c *Credential) JWT(ctx context.Context, authParams authority.AuthParams) (
...
+		}
 
-	if authParams.SendX5C {
-		token.Header["x5c"] = c.X5c
+		return token.SignedString(c.Key)
 	}
 
-	assertion, err := token.SignedString(c.Key)
-	if err != nil {
+	assertion, err := build(signingMethod, thumbprintKey)
```
gladjohn, line 193: Signer-backed operations may involve hardware or remote providers. Can we check `ctx.Err()` before the initial signing attempt and again before the RS256 fallback? We cannot cancel `crypto.Signer.Sign` in flight, but we should avoid starting a signing operation—or a second one—after acquisition has already been canceled.

apps/internal/oauth/ops/accesstokens/signingmethod.go
```diff
@@ -0,0 +1,83 @@
...
+	"fmt"
+
+	"github.com/golang-jwt/jwt/v5"
+)
+
+// signerMethod adapts a jwt.SigningMethod so signing goes through a crypto.Signer instead of
+// requiring an *rsa.PrivateKey. That's the only way to sign with a key whose private material can't
+// be exported, such as a Windows KeyGuard (VBS-isolated) or other CNG/HSM-backed key.
+//
+// Everything except signing delegates to the wrapped method, so the assertion is byte-for-byte what
```
gladjohn: “Byte-for-byte” is inaccurate for the complete assertion because PS256 signatures are randomized, and separate assertion builds also contain different timestamps and `jti` values. Please describe this as preserving the algorithm, header, thumbprint, and verification semantics instead.

4gust (approved):

Robbie-Microsoft:
> This PR is stacked on open PR #647, which is stacked on open PR #632. The complete `main → #649` change spans 41 files and roughly 6,800 added lines, so #649 cannot be evaluated as an independent feature boundary. Can we collapse or rebase the stack before approval so the final security behavior is reviewable against `main`?

yes, of course

apps/internal/oauth/ops/accesstokens/accesstokens.go
Robbie-Microsoft: Took option (b). We can't do (a): `crypto.Signer` has no "unsupported algorithm" sentinel, and CNG `NTE_NOT_SUPPORTED`, PKCS#11 `CKR_MECHANISM_INVALID` and vendor HSM statuses have nothing portable in common. A permission denial still triggers the retry — that behavior is unchanged.

So the deliverable is the rationale replacing "nothing else can realistically have failed". Bounds: only signer-backed credentials reach it; RS256 is what we already send ADFS/dSTS; the retry hits a key store the caller already controls; if RS256 also fails, both errors surface.

I also wrapped `Sign` failures in a typed `signerError` and gated the retry on it — but that gate can't fire today. Inside `JWT()` the header and claims always marshal and the signer was already type-asserted, so every error reaching it is a signing failure. It's a guard for future fallible steps in `build()`, not a fix; removing it leaves the suite green.

apps/confidential/confidential.go
Robbie-Microsoft: Agreed, removed from both layers. It was a memo of a type assertion — `NewCredFromTLSCertificate` literally did `if _, ok := cert.PrivateKey.(*rsa.PrivateKey); !ok`. `JWT()` derives it once. Safe because `accesstokens` is under `apps/internal/`.

Your inconsistent-combinations point had teeth: two tests existed only to pin `Key` and `SignerOnly` disagreeing. Deleted — unrepresentable now.

apps/confidential/confidential.go
Robbie-Microsoft: Added `TestSignerCredentialAllFlows`: client credentials, OBO, auth code, refresh token, user FIC — all through the public API. Asserts `client_assertion`, jwt-bearer type, grant type, PS256 verification against the signer's public key, `x5t#S256`, and signer call count. Corrupting the signature fails all five.

It also caught a doc error: a confidential client's username/password flow sends no client credential, so it's a second assertion-less flow alongside mTLS PoP. Docs and README now say "every flow that authenticates with a client assertion" and name both. ADFS/dSTS covered one layer down.

README.md
Robbie-Microsoft: The example existed but wasn't linked. #647 added `apps/tests/devapps/keyguard` — Windows NCrypt signer over a KeyGuard key, plus provisioning. Now linked as a worked example, explicitly not a supported API. Also took your scoping suggestion: MSAL ships no signer implementation, and obtaining one is the platform-specific part. Kept the general framing since this isn't Windows-only.

apps/internal/oauth/ops/accesstokens/accesstokens.go
Robbie-Microsoft: Both checks added. `ctx` previously only reached `AssertionCallback`. Removing either fails its test.

apps/internal/oauth/ops/accesstokens/signingmethod.go
Robbie-Microsoft: Fixed. Now claims only indistinguishability to the service — same `alg`, thumbprint, signature format, verification semantics — and says the bytes can't match, since PS256 salts are random and each assertion has a fresh `jti`/`nbf`/`exp`.

PR #659, opened 2026-08-21
Demo apps for the msal-go mTLS PR stack

Robbie-Microsoft:
## What this is

Runnable demo/sample apps showcasing the msal-go **mTLS PR stack**, for a live presentation.

`apps/tests/devapps/mtls/` is a standalone `package main` with CLI subcommands, one per capability:

| Subcommand         | PR | Demonstrates |
|--------------------|----|--------------|
| `mtls-pop`         | **#632** | An `mtls_pop` token: `Metadata.TokenType == "mtls_pop"`, `BindingCertificateThumbprint()`, and the decoded `cnf["x5t#S256"]` claim all match — the token is bound to the certificate. |
| `bearer-over-mtls` | **#643** | Same certificate, client built with `WithSendCertificateOverMtls()`: `TokenType` is `Bearer` and `BindingCertificate` is nil — an unbound token that still travelled over the mTLS transport. |
| `cache-isolation`  | **#632 + #643** | PoP and Bearer tokens for the same client/scope occupy separate cache entries and never collide. |
| `signer`           | **#647** `NewCredFromTLSCertificate` / **#649** signer-backed JWT assertions | A `crypto.Signer` whose private key is never exported (the exact API shape a KeyGuard/TPM/HSM key uses), driving a real client-certificate TLS handshake against a local `httptest` server. Runs fully **offline**. |
| `fic`              | **#633** FIC leg 2 (`NewCredFromSignedAssertionCallback`) | The developer-orchestrated two-leg FIC flow over mTLS PoP. Always explains the flow; runs live only when a second federated app is configured. Never fabricates a token. |

## Why this targets an integration branch, not `main`

The samples use APIs from **five unmerged PRs** (#632, #633, #643, #647, #649) and **would not compile against `main`**. The base branch `robbie/mtls-demo-base` is those five PRs merged together (three ordinary merge commits on top of #632). This PR's diff is therefore limited to the new sample files.

**This PR is for demonstration and should be retargeted to `main` once the stack merges.**

## Notes

- No certificate, key, or secret is embedded or committed. The SN/I certificate is supplied at runtime via `-cert` / `MTLS_CERT_PATH`; missing config yields a clear, actionable error, never a panic.
- Cross-platform: no Windows-only code, no build tags. KeyGuard itself is shown from a recording; the `signer` demo uses a software RSA key wrapped in a signer-only type to demonstrate the identical API shape on any platform. CI's `go build ./apps/...` passes on Linux.
- `signer` and `fic`'s explanatory path run with no network. Defaults are the known-good lab values from #632's passing integration tests.

See `apps/tests/devapps/mtls/README.md` for full run instructions.

### DIFF
--- a/apps/tests/devapps/mtls/README.md
+++ b/apps/tests/devapps/mtls/README.md
@@ -1,0 +1,79 @@
+# msal-go mTLS demo app
+
+`package main` demos for the msal-go mutual-TLS (mTLS) PR stack. Each subcommand isolates one
+capability and prints clearly labelled output suitable for a live presentation.
+
+This app targets an **integration branch** that combines five as-yet-unmerged PRs. It will not compile
+against `main` until that stack merges, because the samples use APIs those PRs introduce.
+
+## Subcommands and which PR each maps to
+
+| Subcommand         | PR                                              | Demonstrates |
+|--------------------|-------------------------------------------------|--------------|
+| `mtls-pop`         | **#632** mTLS PoP                               | An `mtls_pop` token: `Metadata.TokenType == "mtls_pop"`, `BindingCertificateThumbprint()`, and the token's decoded `cnf["x5t#S256"]` claim all agree — the token is bound to the certificate. |
+| `bearer-over-mtls` | **#643** bearer-over-mTLS                       | The same certificate, client built with `WithSendCertificateOverMtls()`: `TokenType` is `Bearer` and `BindingCertificate` is nil — an unbound token that still travelled over the mTLS transport. |
+| `cache-isolation`  | **#632 + #643**                                 | PoP and Bearer tokens for the same client/scope occupy separate cache entries and never collide (each type is re-served from cache on a second call). |
+| `signer`           | **#647** `NewCredFromTLSCertificate` / **#649** signer-backed JWT assertions | A `crypto.Signer` whose private key is never exported (the exact API shape a KeyGuard/TPM/HSM key uses), driving a real client-certificate TLS handshake against a local `httptest` server. Runs fully offline. |
+| `fic`              | **#633** FIC leg 2 (`NewCredFromSignedAssertionCallback`) | The developer-orchestrated two-leg federated identity credential (FIC) flow over mTLS PoP. Explains the flow always; runs live only when a second federated app is configured. Never fabricates a token. |
+
+> KeyGuard itself is Windows/VBS-specific and is shown from a pre-existing recording. The `signer`
+> subcommand here uses a **software** RSA key wrapped in a signer-only type so it demonstrates the
+> identical API shape while running on any platform (CI builds this with `go build ./apps/...` on
+> Linux).
+
+## Building
+
+```sh
+go build ./apps/tests/devapps/mtls/...
+# or run directly:
+go run ./apps/tests/devapps/mtls <subcommand> [flags]
+```
+
+## Configuration
+
+Defaults are the known-good lab values proven by #632's passing integration tests
+(`apps/tests/integration/mtls_pop_integration_test.go`). They are public identifiers, not secrets, and
+every one is overridable by flag or environment variable.
+
+| Flag              | Env var              | Default |
+|-------------------|----------------------|---------|
+| `-client-id`      | `MTLS_CLIENT_ID`     | `163ffef9-a313-45b4-ab2f-c7e2f5e0e23e` |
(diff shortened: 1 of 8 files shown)


gladjohn (changes requested):
Thanks for putting this demo together. It helps explain parts of the proposed APIs, but I still do not think #659 is the right delivery approach. I recommend that we close this PR and move the relevant examples and E2E tests into the feature PRs that own them.

The concern raised during review was not that we needed another demo application. We need each security feature to show a clear, independently reviewable developer experience and prove that experience end to end.

Today, #659 depends on several unmerged PRs and a custom integration branch. It does not compile against `main`. A reviewer should not need this chain:

```text
#632 → #633 → #643 → #647 → #649 → integration branch → #659
```

to understand how a customer uses one feature.

Each feature PR should show its complete path:

```text
public API
  → developer code
  → real credential
  → token acquisition
  → returned binding certificate
  → resource call
  → successful E2E
```

## What is still missing

The `mtls-pop` sample acquires an `mtls_pop` token and validates its `cnf` thumbprint. That is useful, but it stops before the developer uses the returned certificate to call the protected resource:

```text
result.AccessToken + result.BindingCertificate
  → configure the resource TLS client
  → Authorization: mtls_pop <token>
  → call Graph or AKV over mTLS
```

The `signer` demo uses a software-generated RSA key and a separately constructed `http.Client`. It does not prove the real scenario with a non-exportable KeyGuard, CNG, or HSM-backed key through the complete MSAL flow.

The cache demo creates separate clients instead of proving the behavior with shared cache state. The FIC command can also succeed after explaining the flow without executing the live exchange.

These examples are helpful, but they do not replace feature-level E2E acceptance evidence.

## Suggested model

.NET PR #6167 is closer to the structure I expect:

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6167

It is based on `main`, has one purpose, and proves one concrete scenario:

```text
real non-exportable VBS/KeyGuard certificate
  → public MSAL API
  → real mtls_pop token
  → validate cnf.x5t#S256
  → Graph mTLS endpoint
  → 200 OK
```

It also proves the negative case:

```text
same bound token + no client certificate
  → Graph
  → 401 / MtlsMissingClientCertificate
```

This clearly shows the credential, public API, returned token, certificate binding, and resource enforcement. One improvement I would make to #6167 is to use `result.BindingCertificate` for the Graph call, proving that the certificate returned by MSAL works for the resource-side mTLS handshake.

## Recommendation

Please move each example and its E2E proof into the PR that delivers that feature:

- **#632 mTLS PoP:** token acquisition, returned `BindingCertificate`, resource call, and positive/negative E2E tests.
- **Non-exportable key support:** real platform-backed key, public signer API, MSAL token request, returned certificate, and resource call.
- **#633 FIC:** complete two-leg public API flow, binding validation, and final resource call.

This would reduce the PR stack and let reviewers understand and accept each feature independently.

**Recommendation: close #659 and make each feature PR independently demonstrate and prove its complete public developer experience.**

bgavrilMS:
@gladjohn - I agree with you that we need to nail down the developer experience first. We should not expose new APIs only to deprecate them later. I am particularly concerned about how we represent certificates across all MSALs. 

Today, we use certificates only for client_assertion computation - i.e. we need cert thumbprint and Sign() 
With these changes, the certificate must also work for mTLS. 

I think it should be sufficient to have a slice of:

- #632 
- #647 
- E2E test showing how developer can call Graph or AKV

as long as we actually expose the BindignCertificate on the AuthResult.

Robbie-Microsoft:
Closing — each demo now lives in the PR it demonstrates, so the feature and its usage review together.

| in this PR | destination |
| --- | --- |
| `pop.go` | #632 → `apps/tests/devapps/mtlspop/` |
| `fic.go` | #633 → `apps/tests/devapps/mtlsfic/` |
| `bearer.go` | #643 → `apps/tests/devapps/bearerovermtls/` |
| `cache.go` | #643 → `apps/tests/devapps/mtlscacheisolation/` |
| `common.go` | inlined into each demo |
| `main.go` | dropped (subcommand dispatcher) |
| `README.md` | one per demo |
| `signer.go` | dropped — it simulated a non-exportable key with `rsa.GenerateKey`; #647 ships the real one (`apps/tests/devapps/keyguard/`, NCrypt/CNG over a VBS-isolated key) |

`robbie/mtls-demo-base` can be deleted with it.

Issue #654, opened 2026-08-18
Regional instance discovery regionalizes un-normalized sovereign aliases (diverges from MSAL .NET)


Robbie-Microsoft:
## Summary

When a region is configured, `AADInstanceDiscovery` regionalizes `AuthorityInfo.Host` verbatim. Sovereign legacy aliases are never normalized to their preferred network host first, so Go produces `usgovvirginia.login.usgovcloudapi.net` where MSAL .NET produces `usgovvirginia.login.microsoftonline.us`.

Found while working on #632. Pre-existing, sits at the discovery layer, and affects all regional sovereign callers rather than just mTLS — the mTLS path does **not** share it (see below), so filing separately rather than fixing there.

## The mechanism

`apps/internal/oauth/ops/authority/authority.go`, in the `region != ""` branch:

```go
environment := authorityInfo.Host
switch environment {
case loginMicrosoft, loginWindows, loginSTSWindows, defaultHost:
    environment = loginMicrosoft
}

resp.TenantDiscoveryEndpoint = fmt.Sprintf(tenantDiscoveryEndpointWithRegion, region, environment, authorityInfo.Tenant)
metadata := InstanceDiscoveryMetadata{
    PreferredNetwork: fmt.Sprintf("%v.%v", region, authorityInfo.Host),
    PreferredCache:   authorityInfo.Host,
    Aliases:          []string{fmt.Sprintf("%v.%v", region, authorityInfo.Host), authorityInfo.Host},
}
```

Two separate problems here:

1. **The normalization switch covers only the four public hosts.** Sovereign aliases (`login.usgovcloudapi.net` → `login.microsoftonline.us`, `login.chinacloudapi.cn` → `login.partner.microsoftonline.cn`) are not in it, so they pass through un-normalized. `GetKnownMetadata` in `known_metadata.go` already knows these mappings and is not consulted here.

2. **`PreferredNetwork` and `Aliases` use `authorityInfo.Host`, not the normalized `environment`.** So even for a public host the two disagree: `login.windows.net` + a region yields `TenantDiscoveryEndpoint` on `{region}.login.microsoftonline.com` but `PreferredNetwork` on `{region}.login.windows.net`. That inconsistency is internal to this one function.

## Comparison with MSAL .NET

`RegionAndMtlsDiscoveryProvider` resolves the preferred network host through `TryGetKnownEnviromentPreferredNetwork` before regionalizing, so the alias is collapsed first:

| | host produced |
[...]

PR #661, opened 2026-08-29
mTLS Proof-of-Possession (SNI): certificate-bound tokens, two-leg FIC, Bearer-over-mTLS, and non-exportable key support

Robbie-Microsoft:
## Summary

This branch adds **RFC 8705 mutual-TLS client authentication and certificate-bound access tokens** to MSAL Go, bringing it to parity with MSAL .NET's `WithMtlsProofOfPossession`. It consolidates five reviewed PRs (#632, #633, #643, #647, #649) plus CI coverage.

**88 files, +20,858 / −147.** The majority of the diff is focused regression coverage and runnable sample apps.

No public symbol is removed, renamed, or re-signatured. The new mTLS APIs are additive. Token responses are now validated case-insensitively against the active authentication scheme before caching, so missing or mismatched `token_type` values fail closed.

> **Azure SDK: no action item.** `azidentity` builds a closed MSAL option list and does not forward caller options, request PoP, or preserve the binding certificate on `azcore.AccessToken`, so this feature is reachable only from a directly constructed `confidential.Client` today. Nothing you do now breaks.

---

## 1. How mTLS works here

The single most common misconception: **there are two independent mTLS connections**, and this feature only owns the first one.

```
                 (1) client auth                    (2) token presentation
  +--------+   mTLS handshake    +-------------+   mTLS handshake   +----------+
  |  Your  | ---- cert --------> | mtlsauth.*  |   ---- cert -----> | Resource |
  |  app   | <-- mtls_pop token -|  (Entra)    |   Authorization:   |          |
  +--------+                     +-------------+   mtls_pop <tok>   +----------+
       \____ MSAL does this ____/  \________ your code does this ________/
```

**Leg (1) — acquiring the token (MSAL's job).**

1. The authority is validated *before any network work*: AAD (`MSSTS`) must be tenanted (not `/common`, `/organizations`, `/consumers`) and on a `login.*` host belonging to a known Microsoft cloud. ADFS is rejected by authority *type*, so `login.microsoftonline.com/adfs` is refused even though its host looks valid. Tenanted dSTS is supported and keeps its own token endpoint — the tenancy and host guards constrain the *derived* `mtlsauth.*` host, and nothing is derived for dSTS. This matches .NET, where `ValidateAadAuthorityForPop` returns early for non-AAD authorities and the rewrite is conditional on a `login.`-prefixed host.
2. The token endpoint host is rewritten `login.*` → `mtlsauth.*`, preserving any configured region (`{region}.mtlsauth.microsoft.com`). Region is optional — the global endpoint is production-ready.
3. MSAL builds (and caches, keyed by certificate thumbprint) an `*http.Client` whose transport carries the certificate in `tls.Config.Certificates`.
4. **For a direct certificate credential, the TLS handshake is the client authentication.** That form sends *no* `client_assertion`, *no* `client_secret`, and *no* `req_cnf`. A signed-assertion callback instead sends its certificate-bound assertion as `jwt-pop` and presents the same certificate on the handshake.
5. Entra returns `token_type=mtls_pop` with a `cnf.x5t#S256` confirmation claim: the SHA-256 of the presented certificate.

**Leg (2) — calling the resource (your job).** Send `Authorization: mtls_pop <token>` *and* present the same certificate on the TLS handshake to the resource. The resource compares the certificate on its connection against `cnf/x5t#S256` in the token. A stolen token is useless without the private key — that is the entire point of the feature.

MSAL hands you everything leg (2) needs as `result.BindingCertificate` (see §3).

---

## 2. Per-flow behavior: acquiring, and calling the resource afterwards

Three distinct modes ship here. They differ in **what the certificate proves** and therefore in **what you do next**.

| | **mTLS PoP** | **Bearer-over-mTLS** | **Two-leg FIC over mTLS PoP** |
|---|---|---|---|
| **Opt in with** | `WithMtlsProofOfPossession()` (per request) | `WithSendCertificateOverMtls()` (at `New()`) | `WithMtlsProofOfPossession()` on both legs |
| **Supported flows** | `AcquireTokenByCredential` **only** | `AcquireTokenByCredential`, `AcquireTokenSilent`, `AcquireTokenByAuthCode`, `AcquireTokenOnBehalfOf` | Leg 1 `ByCredential`; leg 2 `ByCredential` w/ callback credential |
| **Endpoint** | `mtlsauth.*` | `mtlsauth.*` | `mtlsauth.*` on both legs |
| **Client auth to Entra** | Direct certificate: handshake only. Callback: `jwt-pop` + handshake | Certificate credential: `jwt-bearer` + handshake. Signed callback: `jwt-pop` + handshake | Leg 1: handshake. Leg 2: `jwt-pop` client assertion **+** handshake |
| **Returned `token_type`** | `mtls_pop` | `Bearer` | `mtls_pop` |
| **Token bound to cert?** | **Yes** (`cnf.x5t#S256`) | **No** — ordinary bearer token | Yes |
| **`result.BindingCertificate`** | populated | **`nil`** (deliberate) | populated |
| **Calling the resource** | `Authorization: mtls_pop <tok>` **+ same cert on the TLS handshake** | `Authorization: Bearer <tok>` over ordinary TLS. **No client cert.** | `Authorization: mtls_pop <tok>` + cert, as mTLS PoP |
| **Resource must support** | mTLS PoP validation | nothing new | mTLS PoP validation |
| **Cache partition** | by token type + `x5t#S256` | shared with plain Bearer | by token type + `x5t#S256` |

**Why Bearer-over-mTLS exists, and why sharing a cache partition with plain Bearer is correct.** It upgrades *client authentication* from a shared secret to a certificate, without asking the resource to learn a new token type. The token it returns genuinely **is** a plain bearer token — unbound, indistinguishable on the wire — so serving it to a plain-bearer caller is correct, not a leak. `BindingCertificate` is `nil` there precisely so callers can't mistake it for a bound token. `mtls_pop` and `Bearer` tokens can never cross-serve: they are separated by both the cache **read** filter (token-type match) and the **write** key (token type + `AuthnSchemeKeyID`).

**Two-leg FIC, concretely.** Leg 1 uses the SN/I certificate over mTLS PoP to obtain a certificate-bound *federated assertion*. Leg 2 presents that assertion as a `jwt-pop` client assertion **together with the same certificate** on the handshake, yielding the final resource token. The assertion and the TLS identity must agree; the callback credential (§4) is what supplies both.

All three modes are documented in the README, each with a worked example.

---

## 3. New certificate structure

```go
type AuthResult struct {
    // ... existing fields unchanged ...
    BindingCertificate *tls.Certificate `json:"-"`
}

func (ar AuthResult) BindingCertificateThumbprint() string  // base64url SHA-256 (x5t#S256)
```

Design decisions worth knowing:

- **It is a `*tls.Certificate`, not a `*x509.Certificate`.** `.Leaf` is always populated with an independently parsed leaf, and the value can be returned from `tls.Config.GetClientCertificate` for leg (2). `GetClientCertificate` avoids CA-list filtering and supports resources that request the certificate after the initial handshake.
- **`PrivateKey` is only guaranteed to be a `crypto.Signer`.** Do not type-assert to `*rsa.PrivateKey` and do not assume the key material can be exported — it may be VBS-isolated or in an HSM.
- **It is excluded from JSON (`json:"-"`).** `encoding/json` walks into an `*rsa.PrivateKey`'s exported fields, so marshalling an `AuthResult` would otherwise emit the private exponent and primes into whatever consumed the output — a log, a trace, a cache file.
- **`nil` on any unbound path**, including Bearer-over-mTLS. Nil-check before use.

---

## 4. New certificate functionality

**Any `crypto.Signer` can now back a credential.** Previously the credential constructors required an exportable `*rsa.PrivateKey`. A KeyGuard (VBS-isolated) key, a CNG key handle, a TPM key, or an HSM key reached through `PKCS#11` can only ever surface in Go as a `crypto.Signer` — never as an `*rsa.PrivateKey` — so they were structurally impossible to use. Now:

- `NewCredFromTLSCertificate(tls.Certificate)` accepts any `crypto.Signer` as `PrivateKey`. It always parses the leaf from `Certificate[0]` and ignores `cert.Leaf` entirely, so every field of the credential derives from the DER actually presented on the handshake and a stale `Leaf` cannot disagree with it. It validates the signer's public key against that leaf, copies the DER so later mutation can't change the credential, and preserves your leaf-first chain in `x5c`.
- `NewCredFromCert` was relaxed the same way, and still rejects non-RSA keys with an identical error.
- `NewCredFromSignedAssertionCallback` returns both an assertion **and** the certificate it is bound to, which is what makes leg 2 of FIC expressible.

**Every flow works with a non-exportable key**, not just mTLS PoP. MSAL signs `client_assertion` through the signer, so `private_key_jwt`/SN/I works for client credentials, on-behalf-of, auth code, refresh token, user FIC, and ADFS/dSTS. Two paths send no assertion at all: direct certificate-credential mTLS PoP (the handshake is the proof) and confidential username/password (sends no client credential).

**No extra transport work.** `crypto/tls` signs the handshake through the signer on both TLS 1.2 (`PKCS#1` v1.5) and TLS 1.3 (RSA-PSS), so the built-in transport handles these keys. This is covered by a **real TLS 1.2 and 1.3 client-certificate handshake test** against an `httptest` server driven by a signer-only key — reproducing the hardware result in CI, on Linux, with no hardware.

**PS256 → RS256 fallback.** Assertions are signed PS256 (RSA-PSS). Some providers cannot do RSA-PSS at all, so if the signer fails, MSAL rebuilds the assertion once with RS256 (`PKCS#1` v1.5) and an `x5t` thumbprint, matching MSAL .NET. It follows *any* signing failure rather than only an unsupported-algorithm one, because `crypto.Signer` offers no portable way to distinguish them; a failure RS256 can't fix still fails. An exportable `*rsa.PrivateKey` never falls back. **MSAL Go has no logging hook, so a successful retry is silent — the only signal is the `alg` on the wire.**

---

## 5. New public APIs

| Symbol | Package | Purpose |
|---|---|---|
| `WithMtlsProofOfPossession()` | `confidential` | Per-request opt-in to certificate-bound tokens (`AcquireTokenByCredential`) |
| `WithSendCertificateOverMtls()` | `confidential` | App-level opt-in to Bearer-over-mTLS |
| `NewCredFromTLSCertificate(tls.Certificate) (Credential, error)` | `confidential` | Credential from a cert whose key is any `crypto.Signer` |
| `NewCredFromSignedAssertionCallback(func(...) (SignedAssertion, error)) Credential` | `confidential` | Assertion + the certificate it is bound to (FIC leg 2) |
| `SignedAssertion` (type alias) | `confidential` | `{ Assertion string; BindingCertificate *tls.Certificate }` |
| `AssertionRequestOptions` (type alias) | `confidential` | Client ID, final token endpoint, runtime tenant, canonical authority, claims, capabilities, FMI path, and correlation ID for assertion callbacks |
| `WithMtlsHTTPClient(func(tls.Certificate) *http.Client) Option` | `confidential` | Own the mTLS handshake yourself |
| `AuthResult.BindingCertificate` / `.BindingCertificateThumbprint()` | `confidential` | The bound certificate and its `x5t#S256` |
| `MtlsPoPTokenTypeMismatchError` | `errors` | mTLS PoP was requested but the provider returned another `token_type`; match with `errors.As` on a **value** target |

---

## 6. Deprecated or removed APIs

**None.** No symbol is deprecated, removed, renamed, or re-signatured; no `Deprecated:` marker is added anywhere in the diff. Existing callers compile and behave identically without changes.

---

## 7. Behavior changes & consumer impact

Items 1-5 and 7 are scoped to the new mTLS paths. Item 6 applies to all network token responses.

1. **`WithHTTPClient` interacts with mTLS differently.** `WithHTTPClient` accepts an interface (`Do` + `CloseIdleConnections`), but a TLS client certificate can only be installed through an `*http.Transport`. So on mTLS paths MSAL requires the configured client to be an `*http.Client` whose `Transport` is an `*http.Transport` — it clones that transport, preserving your proxy, dialer, and root CAs. If it isn't (an opaque `RoundTripper`, or a transport setting `DialTLS`/`DialTLSContext`), the request **fails with an actionable error naming `WithMtlsHTTPClient`** rather than silently falling back to `http.DefaultTransport` and routing a credential-bearing request outside your proxy, pinning, auditing, and egress controls.
   **This applies to any wrapper type.** A library that constructs the `confidential.Client` itself — `azidentity` among them — cannot forward `WithMtlsHTTPClient` from its own caller, so mTLS PoP is reachable today only from a directly constructed client.
2. **Redirects are refused on the mTLS leg** unless your client sets `CheckRedirect`. A nil policy is *not* "no policy" — it is Go's default, which follows up to 10 redirects, and token requests carry `GetBody`, so a 307/308 would replay the credential-bearing body **and** offer the binding certificate to the redirect target.
3. **TLS 1.2 floor on the mTLS leg**, raised only if you set something lower. A pinned TLS 1.3 is preserved.
4. **A caller-supplied `GetClientCertificate` is cleared on MSAL's clone**, because it takes precedence over `Certificates` during the handshake and would silently suppress the binding certificate. Your config object is never mutated.
5. **Each mTLS client gets a private TLS session cache.** `tls.Config.Clone` shallow-copies `ClientSessionCache` (it's an interface), so clients built for different certificates would otherwise share tickets and a client holding cert B could resume a session the server authenticated under cert A.
6. **`token_type` validation fails closed**: after response validation, MSAL compares `token_type` case-insensitively with the active authentication scheme and rejects missing or mismatched values before caching. mTLS PoP mismatches retain `errors.MtlsPoPTokenTypeMismatchError`.
7. **Cache keys now include token type and `AuthnSchemeKeyID`.** Additive partitioning; plain-bearer behavior is unchanged.

---

## 8. Linux and cross-platform

**The feature is fully cross-platform. Nothing about mTLS PoP is Windows-only.**

The confusion is worth pre-empting: KeyGuard is a Windows technology, but KeyGuard is only *one possible source* of a key. MSAL's contract is `crypto.Signer` — a stdlib interface with no platform dependency.

- **Library code is 100% portable.** No build tags, no `golang.org/x/sys` in library code, no cgo. The `x/sys` dependency is used **only** by the `ncryptsigner` sample (§9).
- **On Linux**, use `PKCS#11` (SoftHSM, CloudHSM, Luna), a KMS-backed signer (Azure Key Vault, AWS KMS), a TPM 2.0 key, or an ordinary PEM file. Anything implementing `crypto.Signer` works identically.
- **MSAL ships no signer implementation on any platform** and adds no platform-specific dependency. Obtaining the signer is the platform-specific part, and it is deliberately yours.
- **The Linux CI job proves it**: the signer-backed TLS 1.2/1.3 handshake tests and the full mTLS PoP suite run on `ubuntu-latest` with no hardware.

---

## 9. `ncryptsigner`: a sample, not API — and how CI covers it

`apps/tests/devapps/keyguard/ncryptsigner` is a **worked example, not a supported API.** It lives under `apps/tests/devapps/`, is not imported by any library code, and is not part of the module's public surface. It shows a Windows NCrypt `crypto.Signer` over a KeyGuard (VBS-isolated) key, plus provisioning steps — the platform-specific work MSAL cannot do for you.

Two CI jobs cover it, one running its tests and one linting it. They catch different failure modes:

| Job | Runner | What it does | Why |
|---|---|---|---|
| **`Build (windows)`** | `windows-latest`, Go 1.22 | Adds `./apps/tests/devapps/...` to the unit-test run | *Executes* the KeyGuard tests on real Windows. `-race` is omitted (needs cgo/gcc). |
| **`lint (windows)`** | `ubuntu-latest`, `GOOS=windows` | Runs golangci-lint over the cross-compiled tree | golangci-lint analyses **one GOOS at a time**, so the existing `lint` job never sees a `//go:build windows` file. Setting `GOOS` lints the CNG code **without needing a Windows runner.** |

Both are new jobs with new names, so the existing `Build` and `lint` check names are unchanged and no branch-protection rule needs updating.

---

## 10. Developer experience

MSAL gets you leg ① — the token from Entra. **You own leg ② — the call to the resource.** For a bound token that second leg is not an ordinary HTTP call, so each PR below is shown end to end: certificate on disk through to a request the resource accepts.

### What actually goes on the wire

| | **#632** mTLS PoP | **#643** Bearer over mTLS | **#633** Two-leg FIC | **#647 / #649** Non-exportable key |
|---|---|---|---|---|
| **Opt in with** | `WithMtlsProofOfPossession()` on `AcquireTokenByCredential` | `WithSendCertificateOverMtls()` on `New()` | `WithMtlsProofOfPossession()` on both legs' acquire calls | *nothing — a different credential constructor* |
| Client assertion in POST body | Direct certificate: **none**; callback: `jwt-pop` | Certificate credential: `jwt-bearer`; signed callback: `jwt-pop` | `jwt-pop` (leg 2) | *unchanged* |
| Client cert on TLS handshake | yes | yes | yes | *unchanged* |
| Token endpoint | `mtlsauth.*` | `mtlsauth.*` | `mtlsauth.*` | *unchanged* |
| `Metadata.TokenType` | `mtls_pop` | `Bearer` | `mtls_pop` | *unchanged* |
| Bound to the cert? | **yes** (`cnf.x5t#S256`) | no | **yes** | *unchanged* |
| `BindingCertificate` | populated | `nil` | populated | *unchanged* |
| Resource call | cert on connection **+** `Authorization: mtls_pop <tok>` | `Authorization: Bearer <tok>`, no client cert | cert on connection **+** `Authorization: mtls_pop <tok>` | *unchanged* |

**Direct certificate-credential #632 is the only form that sends no client assertion** — the certificate on the handshake both authenticates the client and binds the token. Callback-bound forms send `jwt-pop`. #643 sends `jwt-bearer` for a certificate credential and `jwt-pop` for a signed callback. #647 and #649 change only where the private key lives.

#632, #643, and #633 require a **tenanted** authority; `/common`, `/organizations`, and `/consumers` are rejected.

Every identifier below is defined. Errors are `_` for length only — all of these return errors you must check.

```go
ctx       := context.Background()
authority := "https://login.microsoftonline.com/<tenant-id>" // must be tenanted
clientID  := "<client-id>"
scope     := "api://<resource-app-id>/.default"
resource  := "https://<resource-host>/<path>"
```

### #632 — SN/I certificate over mTLS PoP

One certificate does two jobs: authenticates you to Entra, and binds the token to itself.

```go
// ① the token
pemData, _    := os.ReadFile("cert.pem")           // leaf + intermediates + private key
certs, key, _ := confidential.CertFromPEM(pemData, "")
cred, _       := confidential.NewCredFromCert(certs, key)
app, _        := confidential.New(authority, clientID, cred)

res, _ := app.AcquireTokenByCredential(ctx, []string{scope},
    confidential.WithMtlsProofOfPossession())

// res.Metadata.TokenType             == "mtls_pop"
// res.BindingCertificate             *tls.Certificate — chain + live private key
// res.BindingCertificateThumbprint() == the token's cnf["x5t#S256"] claim

// ② the resource: the key goes on the connection, the token goes in the header
client := &http.Client{Transport: &http.Transport{
    TLSClientConfig: &tls.Config{
        GetClientCertificate: func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
            return res.BindingCertificate, nil
        },
        MinVersion:    tls.VersionTLS12,
        MaxVersion:    tls.VersionTLS12,
        Renegotiation: tls.RenegotiateOnceAsClient,
    },
}}
req, _ := http.NewRequestWithContext(ctx, http.MethodGet, resource, nil)
req.Header.Set("Authorization", "mtls_pop "+res.AccessToken) // not "Bearer"
rsp, _ := client.Do(req)
```

`res.BindingCertificate` drops straight into `tls.Config` — you never rebuild it from the PEM. A resource enforcing PoP compares the token's `cnf["x5t#S256"]` against the certificate it saw on the connection; if either half is missing it rejects the call.

You pass `WithMtlsProofOfPossession()` to the **acquire call**, not to `New()` — and `AcquireTokenByCredential` is the only method that accepts it. mTLS PoP is available only for client-credentials (daemon) authentication, never for a flow with a signed-in user. Contrast #643, whose option goes on `New()` and applies to every flow the client runs.

If Entra returns an unbound token when you asked for a bound one, MSAL fails the acquisition rather than handing back a token that only looks bound.

### #643 — Bearer over mTLS

Upgrades *client authentication* from a shared secret to a certificate without asking the resource to learn a new token type. The signed assertion is still sent; the certificate is added to the handshake. Nothing is bound.

```go
// ① the token
pemData, _    := os.ReadFile("cert.pem")
certs, key, _ := confidential.CertFromPEM(pemData, "")
cred, _       := confidential.NewCredFromCert(certs, key)

// an app-level option: it goes on New(), not on the acquire call
app, _ := confidential.New(authority, clientID, cred,
    confidential.WithSendCertificateOverMtls())

res, _ := app.AcquireTokenByCredential(ctx, []string{scope}) // no per-call option

// res.Metadata.TokenType == "Bearer"
// res.BindingCertificate == nil — deliberately, so you cannot mistake it for a bound token

// ② the resource: an ordinary Bearer call, no client certificate
client := &http.Client{}
req, _ := http.NewRequestWithContext(ctx, http.MethodGet, resource, nil)
req.Header.Set("Authorization", "Bearer "+res.AccessToken)
rsp, _ := client.Do(req)
```

Being app-level, it covers the four flows that send a client credential: client credentials, on-behalf-of, authorization code, and silent refresh. `AcquireTokenByUsernamePassword` sends no client credential and is excluded, as is `AcquireTokenByUserFederatedIdentityCredential`; both keep using the regular endpoint. This matches MSAL .NET's `SendCertificateOverMtls`. A per-call `WithMtlsProofOfPossession()` always wins and returns a bound token instead.

It requires a credential that can present a certificate, and `New` says so immediately rather than failing at token time:

```
WithSendCertificateOverMtls requires a certificate credential, such as one from NewCredFromCert,
or a signed-assertion callback (NewCredFromSignedAssertionCallback) that returns a binding certificate
```

The callback credential from #633 (`NewCredFromSignedAssertionCallback`) is accepted. Its certificate is produced at request time, so the callback is pulled forward once per request — the same step mTLS PoP uses — and its certificate selects the mutual-TLS transport while the token stays Bearer. This is the case MSAL .NET calls *implicit Bearer-over-mTLS*. A callback that returns no certificate fails the request rather than silently downgrading to the plain endpoint.

Go requires the explicit `WithSendCertificateOverMtls` opt-in here; .NET's equivalent is ungated, because its builder validates that the option is paired only with a certificate credential and so cannot express the gated form.

### #633 — Two-leg federated identity credential

Your certificate authenticates app 1; app 1's cert-bound assertion authenticates app 2. Both legs are mTLS PoP.

```go
exchangeScope := "api://AzureADTokenExchange/.default"
ficClientID   := "<second-app-client-id>"
ficAuthority  := "https://login.microsoftonline.com/<second-tenant-id>"

// ① leg 1 — an ordinary #632 client; only the scope differs
pemData, _    := os.ReadFile("cert.pem")
certs, key, _ := confidential.CertFromPEM(pemData, "")
leg1Cred, _   := confidential.NewCredFromCert(certs, key)
leg1App, _    := confidential.New(authority, clientID, leg1Cred)

// ② leg 2 — the assertion AND the certificate it is bound to travel together
leg2Cred := confidential.NewCredFromSignedAssertionCallback(
    func(ctx context.Context, _ confidential.AssertionRequestOptions) (confidential.SignedAssertion, error) {
        // Runs on every invocation, so a rotated certificate is picked up automatically.
        // leg1App has its own cache, so the repeat call is cheap.
        leg1, err := leg1App.AcquireTokenByCredential(ctx, []string{exchangeScope},
            confidential.WithMtlsProofOfPossession())
        if err != nil {
            return confidential.SignedAssertion{}, err
        }
        return confidential.SignedAssertion{
            Assertion:          leg1.AccessToken,        // the federated assertion
            BindingCertificate: leg1.BindingCertificate, // required; nil fails the request
        }, nil
    })

leg2App, _ := confidential.New(ficAuthority, ficClientID, leg2Cred)
res, _     := leg2App.AcquireTokenByCredential(ctx, []string{scope},
    confidential.WithMtlsProofOfPossession())

// res.Metadata.TokenType             == "mtls_pop"
// res.BindingCertificate             *tls.Certificate — chain + live private key
// res.BindingCertificateThumbprint() == the token's cnf["x5t#S256"] claim, and equals leg 1's

// ③ the resource
client := &http.Client{Transport: &http.Transport{
    TLSClientConfig: &tls.Config{
        GetClientCertificate: func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
            return res.BindingCertificate, nil
        },
        MinVersion:    tls.VersionTLS12,
        MaxVersion:    tls.VersionTLS12,
        Renegotiation: tls.RenegotiateOnceAsClient,
    },
}}
req, _ := http.NewRequestWithContext(ctx, http.MethodGet, resource, nil)
req.Header.Set("Authorization", "mtls_pop "+res.AccessToken)
rsp, _ := client.Do(req)
```

The binding certificate rides on the credential, not on a call-site option — there is deliberately no `WithBindingCertificate()`. MSAL .NET takes the same position with `ClientSignedAssertion.TokenBindingCertificate`. Use the callback's own `ctx`, as above, so leg 2's cancellation and deadlines propagate into leg 1.

### #647 — Non-exportable key for mTLS PoP

Not a new flow, and no new option. **One line changes** — how the credential is built — and #632, #643, and #633 work exactly as written above:

```go
// before: an exportable key from a PEM
cred, _ := confidential.NewCredFromCert(certs, key)

// now: a key that never leaves the provider
cred, _ := confidential.NewCredFromTLSCertificate(tls.Certificate{...})
```

In full, with the key sourced from Windows CNG:

```go
thumbprint := "<SHA-256 hex thumbprint>"

// ① the key — a crypto.Signer; the private key is never materialized in this process
signer, _ := ncryptsigner.Open("CurrentUser", "My", thumbprint)
defer signer.Close()

// MSAL picks its signing path from the Go TYPE of the key, so it cannot tell a VBS-isolated
// key from a software key behind a crypto.Signer, and claims neither. Assert it yourself —
// a provisioning step that silently fell back to a software key is otherwise invisible.
isolated, _ := signer.IsVirtualIsolated()
if !isolated {
    return errors.New("key is not VBS-isolated")
}

// ② the credential — the only line that differs from #632
    Certificate: signer.Chain(), // DER chain, leaf first
    PrivateKey:  signer,         // any crypto.Signer
})

// ③ the token — the client and the acquire call are identical to #632.
// Nothing here is signer-aware; MSAL treats the key as opaque.
app, _ := confidential.New(authority, clientID, cred)
res, _ := app.AcquireTokenByCredential(ctx, []string{scope},
    confidential.WithMtlsProofOfPossession())

// res.Metadata.TokenType             == "mtls_pop"
// res.BindingCertificate             *tls.Certificate — chain + live private key
// res.BindingCertificateThumbprint() == the token's cnf["x5t#S256"] claim

// ④ the resource — res.BindingCertificate.PrivateKey IS your signer.
// The TLS handshake calls Sign() on it; the key still never leaves the provider.
client := &http.Client{Transport: &http.Transport{
    TLSClientConfig: &tls.Config{
        GetClientCertificate: func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
            return res.BindingCertificate, nil
        },
        MinVersion:    tls.VersionTLS12,
        MaxVersion:    tls.VersionTLS12,
        Renegotiation: tls.RenegotiateOnceAsClient,
    },
}}
req, _ := http.NewRequestWithContext(ctx, http.MethodGet, resource, nil)
req.Header.Set("Authorization", "mtls_pop "+res.AccessToken)
rsp, _ := client.Do(req)
```

`ncryptsigner` is a sample under `apps/tests/devapps/keyguard/`, not part of the library — any `crypto.Signer` works, so Linux HSM and TPM providers need no MSAL change. `signer.Chain()` returns the DER chain leaf-first, which is what both `tls.Certificate` and the JWT `x5c` header want; a chain of one means intermediates are missing, which some relying parties reject.

### #649 — The same key signs client assertions

`private_key_jwt` is not new. What is new is that its key no longer has to be an exportable `*rsa.PrivateKey`, so a non-exportable key is no longer shut out of every flow that sends an assertion.

Take the credential built in #647 and simply don't ask for a bound token:

```go
// ① the key and credential — exactly as in #647
signer, _ := ncryptsigner.Open("CurrentUser", "My", thumbprint)
defer signer.Close()
    Certificate: signer.Chain(),
    PrivateKey:  signer,
})

// ② the token — same credential as #647; the difference is what you DON'T pass.
// No WithMtlsProofOfPossession, so MSAL signs a client assertion with the signer
// instead of presenting the certificate on a handshake.
//
// WithX5C puts the full public chain in the assertion's x5c header. Required for subject
// name / issuer (SN/I) registration, which trusts any certificate with a given subject
// from a given issuer, so certificates rotate without re-registration. Not needed when the
// registration pins one specific certificate — the x5t#S256 header already identifies it.
app, _ := confidential.New(authority, clientID, cred, confidential.WithX5C())

res, _ := app.AcquireTokenByCredential(ctx, []string{scope})

// res.Metadata.TokenType == "Bearer"
// res.BindingCertificate == nil — nothing ties this token to the key that signed for it

// ③ the resource: an ordinary Bearer call, no client certificate
client := &http.Client{}
req, _ := http.NewRequestWithContext(ctx, http.MethodGet, resource, nil)
req.Header.Set("Authorization", "Bearer "+res.AccessToken)
rsp, _ := client.Do(req)
```

This works in every flow that authenticates with a client assertion — client credentials, on-behalf-of, authorization code, refresh token, user federated identity credential, and the ADFS/dSTS RS256 path. The authority need not be tenanted here.

**What changed inside:**

| | Before | Now |
|---|---|---|
| Key type | concrete `*rsa.PrivateKey` | any `crypto.Signer` whose public key is RSA |
| Algorithm | PS256 (`x5t#S256`); RS256 (`x5t`) for ADFS/dSTS | same, plus a one-shot RS256 retry when a signer cannot do RSA-PSS |
| Cancellation | not checked | `ctx` checked before each signing attempt |

The wire format is unchanged — `alg`, headers, and thumbprints are identical for an exportable key. The RS256 retry is reachable **only** by signer-backed keys: an `*rsa.PrivateKey` never falls back, because Go's software RSA always supports PSS, so no caller silently loses PSS on a key that has it. Some providers (CNG, KeyGuard, HSMs, smart cards) cannot do RSA-PSS at all, which is what the retry exists for; MSAL .NET does the same. The outcome is deliberately not memoized, so one transient provider failure cannot pin the process to RS256 for its lifetime. If RS256 fails too, both errors are reported.

Signing may be slow and `crypto.Signer.Sign` cannot be cancelled once running, so MSAL's only control is not starting one for an acquisition you have already cancelled — hence the `ctx` checks.

### Try it before you have a tenant

| Rung | Needs | Command |
|---|---|---|
| Read the flow | nothing | `go run ./apps/tests/devapps/mtlspop` — prints every step, sends nothing |
| Unit tests | nothing | `go test ./apps/confidential/...` |
| Live token | a tenant + SN/I cert | add `-cert`, `-client-id`, `-authority`, `-resource` |

Five runnable demos, each with its own README: `mtlspop`, `bearerovermtls`, `mtlsfic`, `mtlscacheisolation`, `keyguard`.

---

## 11. Tests and validation

- **502 test functions** across the branch, up from 286 on `main`.
- `go build ./...`, `go vet ./...`, `gofmt`, and `golangci-lint` (both GOOS) clean.
- `go test -race` clean on `./apps/confidential/...`.
- Coverage includes: generated assertion audiences and callback options matching the final global, regional, sovereign, private-cloud, and dSTS request URI; callback-bound `jwt-pop` across client credentials, auth code, OBO, and silent refresh; token-type mismatch rejection before caching; complete mutable certificate isolation; nil/typed-nil custom factories and 307/308 refusal; timeout precedence; context-aware region single-flight and dSTS bypass; real TLS 1.2/1.3 signer-only handshakes; and self-issued rollover chain handling.
- Every headline claim was **mutation-tested** — the guard was deliberately broken to confirm a named test fails.
- Five runnable sample apps: `mtlspop`, `mtlsfic`, `bearerovermtls`, `mtlscacheisolation`, and `keyguard`.

---

## 12. Known limitations and deliberate exclusions

- **`WithSendCertificateOverMtls` requires an explicit opt-in for signed-assertion callback credentials**, where MSAL .NET enables the equivalent implicitly whenever the callback returns a certificate. Keeping the option required makes the transport switch explicit and turns a callback that returns no certificate into an error rather than a silent downgrade. Open to matching .NET if ungated parity is preferred.
- **mTLS PoP is `AcquireTokenByCredential`-only**, matching MSAL .NET (`AcquireTokenForClient`). The binding certificate authenticates the application, not a user.
- **`WithInstanceDiscovery(false)` is the private-cloud opt-in**, which necessarily disables the known-cloud allowlist for *every* authority in that app, not only the private one. Accepted tradeoff — it reuses the existing explicit trust signal rather than inventing a second one.
- **Known issue #654:** combining `WithAzureRegion` with a *legacy* sovereign alias (`login.usgovcloudapi.net`, `login.chinacloudapi.cn`) produces a regional **instance-discovery** host built from the un-normalized alias. The mTLS token endpoint normalizes correctly; instance discovery does not. Prefer the modern hostname when configuring a region.
- **On any mTLS request the signed-assertion callback runs on every call, including cache hits**, because the certificate partitions the cache and must be resolved before the lookup. Inherent to the design and matches MSAL .NET, which additionally invokes its delegate a second time to build the request body; Go invokes it once.

---

## 13. Release and versioning

- **Minor bump (1.10.0):** eight new exported symbols — five functions, two types, one method — plus the new `AuthResult.BindingCertificate` field. Nothing removed, renamed, or re-signatured.
- **`go.mod`:** `golang.org/x/sys` moves from indirect to direct (the `ncryptsigner` sample imports it), exactly what `go mod tidy` emits. No new third-party dependency; every new API is Go 1.18-compatible.


apps/confidential/confidential.go
```diff
@@ -228,6 +358,89 @@ func NewCredFromCert(certs []*x509.Certificate, key crypto.PrivateKey) (Credenti
...
+	leaf, err := x509.ParseCertificate(der)
+	if err != nil {
+		return Credential{}, fmt.Errorf("could not parse the leaf certificate: %w", err)
+	}
+	// every public key type x509 can parse implements Equal, added in Go 1.15
+	pub, ok := leaf.PublicKey.(interface{ Equal(crypto.PublicKey) bool })
+	if !ok || !pub.Equal(signerPub) {
+		return Credential{}, errors.New("key doesn't match the leaf certificate")
+	}
+	cred := Credential{cert: leaf, key: cert.PrivateKey}
```
gladjohn, line 464: Reconstructing the TLS certificate drops `SupportedSignatureAlgorithms` and other certificate state such as OCSP staples and signed certificate timestamps. Preserve a deep snapshot of the complete caller-supplied certificate so TLS behavior does not change when the certificate is wrapped with a `crypto.Signer`.

apps/internal/oauth/ops/internal/comm/mtls_test.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	return tls.Certificate{Certificate: [][]byte{der}, PrivateKey: signerKey{key: key}}
+}
+
+// TestBuildMtlsClientSignerHandshake proves crypto/tls can complete a client-certificate handshake
+// when the certificate's key is only a crypto.Signer, which is all a non-exportable key
+// (KeyGuard/CNG/HSM) can ever be. TLS 1.3 signs with RSA-PSS and TLS 1.2 with PKCS#1 v1.5, so both
+// are covered -- which is why each subtest asserts the version it actually negotiated. Without that
+// assertion, losing MaxVersion would silently collapse both subtests onto one padding scheme.
+func TestBuildMtlsClientSignerHandshake(t *testing.T) {
```
gladjohn, line 76: This test accepts either RSA-PSS or PKCS#1, so it does not cover a hardware signer that rejects PSS. Add a signer that supports only PKCS#1 and verify that the copied `SupportedSignatureAlgorithms` causes TLS to select a compatible algorithm.

apps/internal/oauth/ops/authority/mtls.go
```diff
@@ -0,0 +1,252 @@
...
+	// matter what its tenant or host look like. Checking the host first would be both wrong and
+	// misleading. Wrong, because AuthorityType is decided by the authority's first path segment and
+	// never by its host (see NewInfoFromAuthorityURI), so a non-AAD authority can sit on a login.*
+	// host and satisfy every other check here - https://login.microsoftonline.com/adfs and
+	// https://login.microsoftonline.com/dstsv2/<tenant> both do. Misleading, because it would tell
+	// an ADFS or dSTS caller to bring a login.* host, which would not help.
+	//
+	// AAD's value is "MSSTS", so the zero value of Info is not an AAD authority and this guard fails
+	// closed for any Info that was assembled by hand rather than parsed from an authority URL.
+	if i.AuthorityType != AAD {
```
gladjohn: This rejects all non-AAD authorities, including tenanted dSTS authorities supported by MSAL .NET. Tenanted dSTS should retain its existing endpoint rather than being rejected or rewritten to `mtlsauth.*`.

apps/internal/oauth/ops/authority/mtls_test.go
```diff
@@ -0,0 +1,660 @@
...
+// with no type check at all. AuthorityType is decided by the authority's first path segment and never
+// by its host (see NewInfoFromAuthorityURI), so a non-AAD authority can be login.*-hosted:
+// https://login.microsoftonline.com/adfs and https://login.microsoftonline.com/dstsv2/<tenant> both
+// derived an mtlsauth.* endpoint before this guard existed.
+//
+// Each case holds the host and the tenant fixed and varies exactly one thing, the authority type, by
+// running the same parsed Info twice with only AuthorityType changed. The control must be accepted,
+// which proves every other guard passes for these params, so the rejection can only be the type
+// check; the message assertions prove which guard fired.
+func TestMtlsTokenEndpointRequiresAnAADAuthority(t *testing.T) {
```
gladjohn: Add tenanted dSTS coverage for mTLS, regional configuration, and cache hits. The dSTS endpoint should remain unchanged, region should not alter it, and cached tokens should remain usable.

apps/confidential/confidential.go
```diff
@@ -284,20 +499,120 @@ func WithClientCapabilities(capabilities []string) Option {
...
+// NOT bound to it — token_type stays Bearer and the token is cached under the normal Bearer cache key.
+//
+// This is an app-level option honored by the client credentials, on-behalf-of, authorization code,
+// and silent refresh flows. It is not applied to [Client.AcquireTokenByUsernamePassword] (which
+// sends no client credential at all) or [Client.AcquireTokenByUserFederatedIdentityCredential];
+// those two flows continue to use the regular token endpoint. This matches MSAL .NET, whose
+// SendCertificateOverMtls covers the same four flows. A per-request [WithMtlsProofOfPossession]
+// always takes precedence over it. It requires a certificate credential, such as one from
+// [NewCredFromCert]; New returns an error for any other credential kind. That includes
+// [NewCredFromSignedAssertionCallback], even though its callback returns a binding certificate: that
```
gladjohn: The binding certificate returned with a signed assertion is discarded unless explicit PoP was requested. MSAL .NET also uses that certificate to enable mTLS transport when the requested access token remains Bearer. Preserve the assertion and certificate as one atomic callback result and use the certificate for this flow.

apps/confidential/confidential_signed_assertion_test.go
```diff
@@ -0,0 +1,1128 @@
...
+	if res.BindingCertificate.PrivateKey != crypto.PrivateKey(signer) {
+		t.Error("AuthResult.BindingCertificate.PrivateKey is not the opaque signer the callback supplied")
+	}
+}
+
+// TestSignedAssertionCredentialWithoutMtls pins the compatibility requirement: used without mTLS
+// proof-of-possession, a signed-assertion credential behaves exactly like one from
+// NewCredFromAssertionCallback — a bearer client_assertion, the callback invoked lazily and only
+// once, its certificate ignored, and the same AssertionRequestOptions delivered.
+func TestSignedAssertionCredentialWithoutMtls(t *testing.T) {
```
gladjohn, line 896: This test currently requires the callback certificate to be ignored. Replace that expectation with a loopback TLS test proving that the callback is invoked once, its certificate is used on the token request, and the returned access token remains Bearer.

apps/internal/oauth/ops/internal/comm/mtls.go
```diff
@@ -0,0 +1,380 @@
...
+	// tls.Config.Clone shallow-copies ClientSessionCache, because it is an interface: every clone of
+	// one base config points at the same cache object. Two clients built for different binding
+	// certificates would then share tickets, and a client holding certificate B could resume a
+	// session the server authenticated under certificate A. The TLS peer identity would silently
+	// disagree with the certificate this client is bound to, with the assertion that certificate
+	// signed, and with the KeyID the resulting token is cached under. Give every client a private
+	// cache instead. mtlsClient keeps one client per certificate thumbprint, so the cache stays
+	// scoped to a single identity while resumption still works for repeated calls on that identity.
+	// A capacity of 0 selects Go's default.
+	transport.TLSClientConfig.ClientSessionCache = tls.NewLRUClientSessionCache(0)
```
gladjohn: Installing a TLS session cache when the source configuration has none changes authentication behavior. It can also reduce hardware-signer operations through session resumption. Preserve a nil cache and respect `SessionTicketsDisabled`.

apps/internal/oauth/ops/internal/comm/mtls_network_test.go
```diff
@@ -0,0 +1,494 @@
...
+	}
+}
+
+// TestBuildMtlsClientIsolatesSessionCachePerCertificate is the cross-certificate resumption test.
+// tls.Config.Clone shallow-copies ClientSessionCache because it is an interface, so every client
+// cloned from one base used to point at the same cache object. A client bound to certificate B could
+// then resume a session the server had authenticated under certificate A: the TLS peer identity, the
+// certificate that signed the assertion and the cache KeyID would all disagree. The run is pinned to
+// TLS 1.2 so the session ticket is delivered inside the handshake and resumption is deterministic.
+func TestBuildMtlsClientIsolatesSessionCachePerCertificate(t *testing.T) {
```
gladjohn, line 424: Add coverage for TLS resumption behavior. Check `DidResume` and signer invocation counts for nil caches, caller-provided caches, disabled session tickets, and different binding-certificate identities.

apps/internal/oauth/ops/internal/comm/mtls.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+	}
+	if t, ok := http.DefaultTransport.(*http.Transport); ok && t != nil {
+		if err := rejectTLSDialHooks(t, "http.DefaultTransport", defaultHookRemedy); err != nil {
+			return nil, err
+		}
+		return t.Clone(), nil
+	}
+	// Freshly constructed and never handed out, so it cannot carry a hook.
+	return &http.Transport{}, nil
```
gladjohn: Falling back to a bare `http.Transport` when `http.DefaultTransport` is a custom `RoundTripper` silently loses proxy, pinning, auditing, signing, tracing, and egress behavior. Fail closed and require an explicit mTLS client factory when the transport cannot be cloned safely.

apps/internal/oauth/ops/internal/comm/mtls_network_test.go
```diff
@@ -0,0 +1,494 @@
...
+// TestBuildMtlsClientRejectsTLSDialHooksOnDefaultTransport closes the path that never touches the
+// caller's transport at all. http.DefaultTransport is an exported package-level variable, so
+// tracing, proxy-injection and test libraries can install a TLS dial hook on it process-wide. Both
+// entry points that fall through to it - no configured client, and an *http.Client whose Transport
+// is nil - must be checked too, or MSAL would clone a transport that silently drops the binding
+// certificate through a route the caller-transport check cannot see.
+//
+// This test mutates process-global state. It must never call t.Parallel or run alongside anything
+// that touches the default transport, and it restores the original hooks through t.Cleanup.
+func TestBuildMtlsClientRejectsTLSDialHooksOnDefaultTransport(t *testing.T) {
```
gladjohn, line 292: Add a custom non-`*http.Transport` default transport and verify that mTLS client creation fails before any request is sent. The test should also confirm that no bare fallback transport is used.

apps/internal/oauth/ops/accesstokens/accesstokens.go
```diff
@@ -444,15 +608,42 @@ func (c Client) doTokenResp(ctx context.Context, authParams authority.AuthParams
...
 	}
-	return resp, resp.Validate()
+	if err := resp.Validate(); err != nil {
+		return resp, err
+	}
+	// mTLS PoP is a security primitive: when the caller requests a certificate-bound token the
+	// identity provider must honor it. Fail closed on a downgrade (e.g. token_type=Bearer) rather
+	// than returning a token that only looks bound. Mirrors MSAL .NET's TokenClient token_type
+	// check (error code "token_type_mismatch").
+	if authParams.IsMtlsPoP && !strings.EqualFold(resp.TokenType, authority.AccessTokenTypeMtlsPoP) {
```
gladjohn, line 663: Token-type validation is currently one-sided. Explicit PoP must require `mtls_pop`, while Bearer-over-mTLS must require Bearer. Reject mismatches before constructing or caching the authentication result.

apps/internal/oauth/ops/accesstokens/accesstokens_tokentype_mtls_test.go
```diff
@@ -0,0 +1,82 @@
...
+		t.Fatalf("expected MtlsPoPTokenTypeMismatchError, got %T: %v", err, err)
+	}
+	if mism.Expected != authority.AccessTokenTypeMtlsPoP || mism.Actual != authority.AccessTokenTypeBearer {
+		t.Fatalf("mismatch = {Expected:%q Actual:%q}, want {mtls_pop, Bearer}", mism.Expected, mism.Actual)
+	}
+}
+
+// TestFromClientCertificateMtlsPoPTokenTypeHonored verifies the guard passes when the identity
+// provider returns token_type=mtls_pop as requested.
+func TestFromClientCertificateMtlsPoPTokenTypeHonored(t *testing.T) {
```
gladjohn, line 73: Add the inverse mismatch case where a Bearer-over-mTLS request receives `mtls_pop`. Verify that acquisition fails and that the response is not written to the token cache.

apps/confidential/confidential.go
```diff
@@ -1012,6 +1452,288 @@ func WithAttribute(attrValue string) interface {
...
+		return nil, fmt.Errorf("binding certificate leaf could not be parsed: %w", err)
+	}
+	if cert.Leaf != nil && !cert.Leaf.Equal(leaf) {
+		return nil, errors.New("binding certificate Leaf is not the certificate in Certificate[0]: the handshake would present one certificate while the token was bound and cached under the other")
+	}
+	certKey, ok := leaf.PublicKey.(publicKeyEqual)
+	if !ok {
+		return nil, fmt.Errorf("binding certificate public key of type %T cannot be compared with the private key", leaf.PublicKey)
+	}
+	if !certKey.Equal(signer.Public()) {
```
gladjohn: `validBindingCertificate` calls `signer.Public()` without rejecting a typed-nil signer. A callback can therefore pass validation and panic here. Apply the same typed-nil protection used at the other credential boundaries.

apps/confidential/confidential_signed_assertion_test.go
```diff
@@ -0,0 +1,1128 @@
...
+	}
+	return client
+}
+
+// TestSignedAssertionCallbackRejectsUnusableCerts covers the guardrails on the certificate the
+// callback returns: one that can't complete a handshake must fail with a clear error before any
+// network call, not panic and not produce a broken handshake. A nil BindingCertificate is not in
+// this table because it is rejected before validBindingCertificate runs, with a different message;
+// see TestSignedAssertionNilBindingCertificateErrors.
+func TestSignedAssertionCallbackRejectsUnusableCerts(t *testing.T) {
```
gladjohn, line 215: Add callback tests for a typed-nil signer and for a signer returning a typed-nil public key. Both cases should return a validation error without performing network activity or changing the cache.

apps/internal/oauth/ops/internal/comm/mtls.go
```diff
@@ -0,0 +1,380 @@
...
+		factory, base, generation := c.mtlsFactory, c.client, c.mtlsGeneration
+		c.mtlsMu.Unlock()
+
+		// Build outside the lock. mtlsMu is a plain sync.Mutex, so a factory that calls back into
+		// this Client would deadlock permanently, and a merely slow factory would serialize creation
+		// for every other certificate. MSAL .NET builds outside its lock too: SimpleHttpClientFactory
+		// evaluates CreateMtlsHttpClient(cert) before GetOrAdd is entered.
+		entry := mtlsCacheEntry{owned: factory == nil}
+		if factory != nil {
+			entry.client = factory(*cert)
```
gladjohn: `factory(*cert)` copies only the outer `tls.Certificate` structure. Its DER chains and other slices remain shared, allowing the factory to mutate certificate state after the cache identity has been calculated. Pass a deep copy to the factory.

apps/internal/oauth/ops/internal/comm/mtls_test.go
```diff
@@ -0,0 +1,254 @@
...
+func TestMtlsClientRejectsNilFactoryResult(t *testing.T) {
+	cert := &tls.Certificate{Certificate: [][]byte{{0x22}}, PrivateKey: testKey}
+	c := &Client{}
+	c.SetMtlsClientFactory(func(tls.Certificate) HTTPClient { return nil })
+	if _, err := c.mtlsClient(cert); err == nil {
+		t.Error("mtlsClient with nil-returning factory = nil error, want error")
+	}
+}
+
+func TestMtlsClientUsesFactoryOverride(t *testing.T) {
```
gladjohn, line 241: Add a factory that mutates the certificate it receives and prove that the retained credential, cache key, and returned binding certificate remain unchanged. Also verify that a single static client cannot be reused for different certificate identities.

apps/confidential/confidential.go
```diff
@@ -195,16 +230,111 @@ func NewCredFromSecret(secret string) (Credential, error) {
...
+		// a typed-nil *rsa.PublicKey satisfies the assertion, so reject it here rather than
+		// dereference it below
+		if !ok || pub == nil {
+			return cred, errors.New("key must be an RSA key")
+		}
+		k = pub
+	default:
 		return cred, errors.New("key must be an RSA key")
 	}
 	for _, cert := range certs {
```
gladjohn, line 346: `NewCredFromCert` retains the caller-owned parsed certificate while separately snapshotting `cert.Raw` for `x5c`. Later mutation can make the assertion, TLS identity, cache key, and returned binding certificate describe different states. Derive every representation from one immutable DER snapshot.

apps/confidential/confidential_signer_test.go
```diff
@@ -0,0 +1,895 @@
...
+			}()
+			if _, err := test.newCred(); err == nil {
+				t.Fatal("expected an error because the key is a nil pointer")
+			}
+		})
+	}
+}
+
+// TestNewCredFromCertWithSigner covers the same non-exportable key arriving through NewCredFromCert.
+func TestNewCredFromCertWithSigner(t *testing.T) {
```
gladjohn, line 314: Add mutation and same-key renewal tests for `NewCredFromCert`. Changes to the caller's parsed certificate after construction must not affect assertions, TLS presentation, cache identity, or the returned certificate.

apps/internal/base/base.go
```diff
@@ -551,6 +674,9 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 		return AuthResult{}, err
 	}
 
+	if authParams.IsMtlsPoP {
+		ar.BindingCertificate = bindingCertWithLeaf(authParams.MtlsBindingCert)
```
gladjohn: `bindingCertWithLeaf` silently returns nil when certificate DER is missing or malformed. A successful `mtls_pop` result must always contain a usable binding certificate. Return an error instead of producing a successful result with a nil certificate.

apps/internal/base/binding_cert_test.go
```diff
@@ -0,0 +1,367 @@
...
+		<-done
+	}
+	if shared.Leaf != nil {
+		t.Error("concurrent resolution mutated the shared certificate")
+	}
+}
+
+// TestBindingCertWithLeafNilHandling pins the pre-existing nil behavior: no certificate, no DER, or
+// unparsable DER all yield nil rather than a half-built certificate.
+func TestBindingCertWithLeafNilHandling(t *testing.T) {
```
gladjohn, line 300: Add malformed and missing-DER cases and verify that acquisition fails before the token is cached. A subsequent request must perform a new network acquisition rather than returning the invalid result from cache.

README.md
```diff
@@ -145,6 +145,259 @@ use with `NewCredFromCert`.
...
+  `http.RoundTripper` (a tracing, retry, pinning or request-signing wrapper), or an `*http.Transport`
+  that sets `DialTLS`/`DialTLSContext`, cannot carry the binding certificate into the handshake;
+  rather than silently rerouting a credential-bearing request onto `http.DefaultTransport`, mTLS
+  token requests fail with an error pointing at `WithMtlsHTTPClient`. Redirects are refused on that
+  leg unless the `WithHTTPClient` client sets `CheckRedirect`, because a 307 or 308 would replay the
+  request body and present the binding certificate to the redirect target.
+- **`WithMtlsHTTPClient` is required when the value you pass to `WithHTTPClient` is not an
+  `*http.Client`.** `WithHTTPClient` takes an interface (`Do` + `CloseIdleConnections`), but a TLS
+  client certificate can only be installed through `*http.Transport`, so any other implementation is
+  rejected before the request is sent. Azure's `azidentity` is the concrete case: it passes its own
```
gladjohn: The current Azure SDK for Go integration cannot follow this guidance. `azidentity` does not expose arbitrary MSAL options, request this PoP flow, or preserve the token type and binding certificate when converting the result to `azcore.AccessToken`. Remove this claim until the SDK has a supported integration.

apps/confidential/confidential_mtls_httpclient_ext_test.go
```diff
@@ -0,0 +1,73 @@
...
+// could construct a value of that function type — the one supported override did not compile for
+// consumers.
+//
+// Passing a plain func(tls.Certificate) *http.Client literal below is the whole point: if the
+// signature ever regresses to an internal type, this file stops compiling.
+//
+// MSAL .NET's equivalent, IMsalMtlsHttpClientFactory with
+// GetHttpClient(X509Certificate2) -> HttpClient, is public on every target framework, so this is also
+// a parity fix.
+func TestWithMtlsHTTPClientIsCallableExternally(t *testing.T) {
```
gladjohn, line 32: This extension test proves direct MSAL usage but not the documented `azidentity` scenario. Add coordinated Azure SDK acceptance coverage that acquires and consumes the certificate-bound result, or limit the documentation to direct MSAL clients.

apps/internal/base/base.go
```diff
@@ -111,12 +131,90 @@ type AuthResult struct {
...
+	// only has to implement crypto.Signer, so besides an exportable *rsa.PrivateKey it may be a
+	// non-exportable key backed by KeyGuard, CNG or an HSM, and callers must not assume the key
+	// material can be exported. The value drops directly into tls.Config.Certificates, so present it
+	// as the client certificate when calling the resource and the connection will match the token
+	// binding.
+	//
+	// It is excluded from JSON: encoding/json walks into an *rsa.PrivateKey's exported fields, so
+	// marshalling an AuthResult would otherwise emit the private exponent and primes into whatever
+	// consumes the output (a log, a trace, a cache file).
+	BindingCertificate *tls.Certificate `json:"-"`
```
gladjohn, line 151: `json:"-"` protects the certificate only when the complete `AuthResult` is serialized. `BindingCertificate` can still be serialized or logged independently and may contain an exportable RSA private key. Consider exposing a safer binding-certificate representation or clearly document the field as secret-bearing.

apps/internal/base/binding_cert_json_test.go
```diff
@@ -0,0 +1,191 @@
...
+	if !strings.Contains(got, `"AccessToken":"at"`) {
+		t.Errorf("AuthResult no longer serializes its other fields: %s", got)
+	}
+}
+
+// TestAuthResultJSONIndicatorsAreReal proves the assertions above can actually fail. It marshals the
+// same certificate through a struct WITHOUT the json:"-" tag and requires every indicator to show up.
+// If encoding/json ever stopped walking into *rsa.PrivateKey, the test above would pass vacuously and
+// this one would catch that.
+func TestAuthResultJSONIndicatorsAreReal(t *testing.T) {
```
gladjohn, line 111: Extend this test to cover direct serialization and common formatting of `BindingCertificate`, not only serialization of its containing `AuthResult`. The expected public diagnostics contract should ensure private key values cannot be disclosed accidentally.

gladjohn:
Thanks for putting this together @Robbie-Microsoft. I reviewed the history from **#632, #633, #643, #647, and #649** against this consolidated PR. A substantial set of the security and correctness findings raised during those reviews has now been addressed in the implementation and carried forward with dedicated regression, integration, and E2E coverage.

**The resolved security work includes:**

- fail-closed handling when an mTLS PoP request receives a weaker Bearer token, including preventing that response from entering the cache;
- certificate-specific token-cache partitioning using the binding certificate identity, including same-key certificate-renewal and A→B→A cache-isolation coverage;
- prevention of HTTP 307/308 credential replay on the mTLS token leg;
- fail-closed handling when an application transport cannot safely preserve proxy, pinning, auditing, request-signing, or egress controls;
- rejection of TLS dial hooks that can bypass `TLSClientConfig` and present a different certificate than the one MSAL associates with the request;
- isolation of TLS session state between different binding-certificate identities;
- trusted-authority validation before certificates or FIC assertions are transmitted;
- validation that the signer public key matches the certificate used for binding;
- preservation of assertion/certificate atomicity so certificate rotation cannot pair an assertion for certificate A with TLS identity B;
- isolation of returned certificate DER and parsed certificate state from MSAL’s internally retained credential and transport state;
- support for non-exportable `crypto.Signer` implementations without exporting private-key material;
- PS256 → RS256 compatibility for providers that cannot perform RSA-PSS;
- cancellation checks before hardware/remote signing and before a fallback signing operation;
- bounded per-certificate mTLS transport caching with idle-connection cleanup;
- regional mTLS PoP coverage; and
- end-to-end proof of the resource binding contract, including negative cases for a missing certificate, wrong authorization scheme, and different certificate.

Several of these review findings resulted in changes to the implementation or security model and were subsequently protected by targeted regression tests. It is good to see that work preserved as the previously stacked changes were consolidated into a single reviewable PR.

apps/confidential/confidential.go
Robbie-Microsoft: Fixed in f3b32df. The three fields are carried on the credential as `accesstokens.TLSCertFields` and restored in `resolveMtlsBindingCert`, deep-copied and nil-preserving — an empty `SupportedSignatureAlgorithms` isn't the same as an unset one.

apps/internal/oauth/ops/internal/comm/mtls_test.go
Robbie-Microsoft: Added `TestPSSRefusingSignerReachesTheHandshake` and `TestNewCredFromTLSCertificatePreservesHandshakeFields` in `confidential_mtls_certfields_test.go`. The signer returns an error for `*rsa.PSSOptions`, so it models a key that genuinely cannot do PSS rather than one that merely prefers not to. Verified both fail when the field is dropped.

apps/internal/oauth/ops/internal/comm/mtls.go
Robbie-Microsoft: Fixed — a nil cache is now left alone. Nil means the caller declined resumption, so installing one there quietly re-enabled a TLS feature they had turned off. A non-nil cache is still replaced per client so two certificates can't share tickets.

apps/internal/oauth/ops/internal/comm/mtls_network_test.go
Robbie-Microsoft: Added `TestBuildMtlsClientLeavesNilSessionCacheNil` in `comm/mtls_isolation_test.go` with both halves — nil stays nil, and a shared cache is replaced with a private one. Verified in both directions.

apps/internal/oauth/ops/internal/comm/mtls.go
Robbie-Microsoft: Fixed, it fails closed now. One thing worth adding: the bare `&http.Transport{}` also had no `Proxy` function, so unlike `http.DefaultTransport` it ignored `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY`. In a proxy-required environment that turned a security fallback into an unexplained connectivity failure. The error names that.

apps/internal/oauth/ops/internal/comm/mtls_network_test.go
Robbie-Microsoft: Added `TestCloneBaseTransportFailsClosedOnReplacedDefaultTransport`, which swaps in a non-`*http.Transport` and asserts creation fails before any request is made, and `TestCloneBaseTransportStillWorksWithRealDefaultTransport` so the new branch can't start firing on the ordinary path unnoticed.

apps/internal/oauth/ops/accesstokens/accesstokens.go
Robbie-Microsoft: Fixed. One note: I used a plain error rather than `MtlsPoPTokenTypeMismatchError`, because that type documents `Expected` as always being `mtls_pop` — reusing it for the inverse case would break its own contract. Happy to add a separate typed error if you'd rather have one callers can match on.

apps/internal/oauth/ops/accesstokens/accesstokens_tokentype_mtls_test.go
Robbie-Microsoft: Added `TestUnrequestedMtlsPoPTokenTypeRejected`, plus `TestUnrequestedBearerTokenTypeAccepted` so the check can't be tightened into a blanket rejection of every non-`mtls_pop` response without a test failing.

apps/confidential/confidential.go
Robbie-Microsoft: Fixed — both a typed-nil signer and a signer returning a typed-nil public key are rejected before anything is called on them. The constructors already guarded these two shapes; the callback path didn't.

apps/confidential/confidential_signed_assertion_test.go
Robbie-Microsoft: Added `TestValidBindingCertificateRejectsTypedNilSigner` and `TestValidBindingCertificateRejectsTypedNilPublicKey`. Both recover from a panic and fail explicitly if one happens, so they distinguish "returned an error" from "didn't crash by luck". There's a positive case alongside them.

apps/internal/oauth/ops/internal/comm/mtls.go
Robbie-Microsoft: Fixed. The chain is deep-copied before the cache key is computed, so the key and the bytes presented on the wire derive from the same snapshot. `PrivateKey` is deliberately still shared — it's the live signer, and a non-exportable key can't be copied at all.

apps/internal/oauth/ops/internal/comm/mtls_test.go
Robbie-Microsoft: Added `TestMtlsClientPinsCertificateDER`, where the factory retains what it was handed and the test then mutates the caller's DER, and `TestMtlsClientPinnedCertReachesBuiltClient` for the default path where `BuildMtlsClient` constructs the client.

apps/confidential/confidential.go
Robbie-Microsoft: Fixed. The matched leaf is re-parsed from a private copy of the DER, and x5c is encoded from that same copy, so the leaf and the x5c entry can't describe different bytes. This is what `NewCredFromTLSCertificate` already did.

apps/confidential/confidential_signer_test.go
Robbie-Microsoft: Added `TestNewCredFromCertIsolatesCertificateDER`, which mutates the caller's `Raw` after construction and asserts the credential's leaf, its x5c entry, and the agreement between them all hold.

apps/internal/base/base.go
Robbie-Microsoft: Fixed, and moved ahead of `m.Write`. Resolving it afterwards was the worse half: the unusable token was already in the cache, so every later acquisition served it from there and never re-ran this path. Same guard on the cache-read path in `AcquireTokenSilent`.

apps/internal/base/binding_cert_test.go
Robbie-Microsoft: Added `TestAuthResultFromTokenRequiresBindingCertificate` covering nil, no DER chain, empty leaf and unparsable leaf, each asserting nothing was written to the cache, and `TestAcquireTokenSilentRequiresBindingCertificate` for the cache-hit path. There's a success case asserting the token *is* cached, so the "nothing cached" assertions aren't passing vacuously.

apps/internal/base/base.go
Robbie-Microsoft: Doc fixed, but your comment made me check the claim rather than assume it, and I had it partly wrong. Marshalling the `*tls.Certificate` on its own does reach the key. `fmt` does not — `%v`, `%+v` and `%#v` print a nested pointer as an address. That's incidental rather than a protection, though: a logger that serializes structs instead of formatting them gets the marshalling behavior. The comment now says exactly that.

apps/internal/base/binding_cert_json_test.go
Robbie-Microsoft: Added `TestBindingCertificateDirectSerializationLeaks`, which asserts direct marshalling leaks, the `AuthResult` path does not, and all three `fmt` verbs do not. The first assertion is inverted on purpose — if `encoding/json` ever stopped walking into the key, the tag would still be correct but the warning would be stale, and this catches that.

apps/confidential/examples_test.go
```diff
@@ -84,3 +86,133 @@ func ExampleClient_AcquireTokenByCredential_withFMIPath() {
...
+
+	// Leg 2: federated assertion (jwt-pop) + binding cert -> final mtls_pop token. One callback
+	// returns both, so the assertion and the certificate it is bound to can never drift apart.
+	assertionCred := confidential.NewCredFromSignedAssertionCallback(
+		func(context.Context, confidential.AssertionRequestOptions) (confidential.SignedAssertion, error) {
+			return confidential.SignedAssertion{
+				Assertion:          leg1.AccessToken,
+				BindingCertificate: leg1.BindingCertificate,
+			}, nil
+		})
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed — leg 1 is computed inside the callback now, and the callback uses its context instead of discarding it. This is the exact hazard the `mtlsfic` sample warns about in its own comment, so the example was contradicting the sample.

apps/confidential/confidential.go
```diff
@@ -1012,6 +1452,288 @@ func WithAttribute(attrValue string) interface {
...
+	if len(cert.Certificate) == 0 || len(cert.Certificate[0]) == 0 {
+		return nil, errors.New("binding certificate carries no certificate chain")
+	}
+	if cert.PrivateKey == nil {
+		return nil, errors.New("binding certificate has no private key, so it cannot be presented on a TLS handshake")
+	}
+	signer, ok := cert.PrivateKey.(crypto.Signer)
+	if !ok {
+		return nil, fmt.Errorf("binding certificate private key of type %T does not implement crypto.Signer", cert.PrivateKey)
+	}
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed — rejected with an error before `Public()` is called.

apps/confidential/confidential.go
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed — the typed-nil public key is rejected before it reaches `Equal`, which asserts its argument's type and then dereferences it.

apps/tests/devapps/keyguard/ncryptsigner/signer.go
```diff
@@ -0,0 +1,655 @@
...
+	}
+
+	if pss, ok := opts.(*rsa.PSSOptions); ok {
+		saltLen := pss.SaltLength
+		switch saltLen {
+		case rsa.PSSSaltLengthAuto, rsa.PSSSaltLengthEqualsHash:
+			// Auto means "as large as possible" when signing, but CNG wants an explicit length and
+			// every TLS PSS scheme uses a salt the size of the hash.
+			saltLen = hash.Size()
+		}
```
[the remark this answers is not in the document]
Robbie-Microsoft: Leaving this one as-is. The fixed salt length is deliberate and documented at `signer.go:229`; it's correct for both TLS (RFC 8446) and JWA (RFC 7518), and this is sample code rather than library code. Happy to switch it to `PSSSaltLengthAuto` if you'd rather it match Go's default semantics exactly.

apps/internal/oauth/ops/accesstokens/signingmethod.go
```diff
@@ -0,0 +1,111 @@
...
+	return s.method.Alg()
+}
+
+// Verify delegates to the wrapped method because verification uses the public key, which is always
+// available.
+func (s *signerMethod) Verify(signingString string, sig []byte, key interface{}) error {
+	return s.method.Verify(signingString, sig, key)
+}
+
+// Sign hashes signingString and has key, which must be a crypto.Signer, sign the digest.
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed.

docs/federated_managed_identity.md
```diff
@@ -119,6 +119,112 @@ func main() {
...
+  the only way to supply it: MSAL exposes no call-site option that would let the assertion and the
+  certificate be sourced separately.
+- Results expose the binding certificate as `BindingCertificate` (a `*tls.Certificate` carrying the
+  parsed leaf and the private key, ready for `tls.Config.Certificates`) and its thumbprint as
+  `BindingCertificateThumbprint()`.
+- The private key may be non-exportable (Windows KeyGuard, CNG, an HSM); MSAL only requires that it
+  implement `crypto.Signer`.
+- Sovereign clouds are supported via `login.microsoftonline.us` (US Gov) and
+  `login.partner.microsoftonline.cn` (China); the legacy hosts `login.usgovcloudapi.net` and
+  `login.chinacloudapi.cn` are rejected with guidance toward the supported host.
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed — the doc now says the legacy hosts are treated as aliases and resolve to their preferred-network host, which matches `authority.go` and what the README already said.

apps/confidential/confidential.go
```diff
@@ -284,20 +499,120 @@ func WithClientCapabilities(capabilities []string) Option {
...
+// It isn't needed for non-exportable keys: the binding certificate arrives here as a
+// [tls.Certificate] whose PrivateKey only has to implement [crypto.Signer], and a signer supplied
+// through [NewCredFromTLSCertificate] is passed straight to the built-in transport, which crypto/tls
+// signs with on both TLS 1.2 and 1.3. A KeyGuard, CNG or HSM-backed key is presented on the
+// handshake like any other.
+//
+// When unset, MSAL auto-builds and caches an mTLS client per certificate thumbprint.
+//
+// A caller-supplied client belongs to the caller: MSAL never calls CloseIdleConnections on it during
+// its own cache housekeeping, so a factory is free to memoize and return one shared client.
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed — it now recommends memoizing one client per certificate thumbprint rather than sharing a single client across certificates.

.github/workflows/golangci-lint.yml
```diff
@@ -24,5 +24,23 @@
...
+    runs-on: ubuntu-latest
+    env:
+      GOOS: windows
+    steps:
+      - uses: actions/setup-go@v6
+        with:
+          go-version: "1.22"
+      - uses: actions/checkout@v5
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v9
```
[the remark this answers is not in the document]
Robbie-Microsoft: Already addressed. The only third-party action is SHA-pinned (`go.yml:31` and `:50`); the first-party `actions/*` are intentionally left on tags, with a comment stating that policy.

.github/workflows/go.yml
```diff
@@ -50,3 +50,35 @@
...
+        run: go get -v -t -d ./...
+
+      - name: Build
+        run: go build ./apps/...
+
+      # ./apps/tests/devapps/... is what reaches the KeyGuard packages; the other dev apps report
+      # "no test files" and cost nothing. -race is omitted deliberately: it requires cgo and a C
+      # toolchain, and these are single-goroutine parsing and validation tests.
+      - name: Unit Tests
+        run: go test -short ./apps/cache/... ./apps/confidential/... ./apps/public/... ./apps/internal/... ./apps/managedidentity/... ./apps/tests/devapps/...
```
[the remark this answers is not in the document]
Robbie-Microsoft: Already addressed — `permissions: contents: read` is set at `go.yml:12` and `golangci-lint.yml:15`.

apps/internal/oauth/ops/authority/mtls.go
Robbie-Microsoft: Fixed. .NET's `ValidateAadAuthorityForPop` returns early for non-AAD rather than rejecting, and the `login.` -> `mtlsauth` rewrite is conditional on host shape, not authority type. `MsalErrorMessage`'s "only supported for AAD authority type" is defined but never thrown - likely the origin.

dSTS now keeps its endpoint. Also scoped the tenancy and trusted-host guards to AAD: they constrain a *derived* host, and nothing is derived for dSTS. Deleting the type check alone wasn't enough - dSTS then fell through to the `login.*` check.

apps/internal/oauth/ops/authority/mtls_test.go
Robbie-Microsoft: Added. Unit: own host, `login.*` host, untrusted host, scheme/userinfo, region, unresolved endpoint fails closed - all assert no `mtlsauth`. End-to-end: `TestMtlsPoPDSTSKeepsItsEndpointAndCaches` plus a regional case. The mock panics on an empty queue, so the cache hit is proven by consuming no response.

apps/confidential/confidential.go
Robbie-Microsoft: Fixed. The atomic result already shipped (`SignedAssertion{Assertion, BindingCertificate}` is the same shape as `ClientSignedAssertion`), but the cert was only consumed on the PoP path. All four bearer paths now resolve it through the same `resolveSignedAssertion`; `New` no longer refuses the combination.

**One divergence to confirm:** Go requires `WithSendCertificateOverMtls`; .NET's Case 2 is ungated because `Validate` pins that option to cert credentials, so .NET can't express the gate. I kept it - routing a client credential to another host shouldn't be implicit, and a callback returning no cert becomes an error instead of a silent downgrade. Want ungated parity instead?

Also: callback runs once per request, not twice.

apps/confidential/confidential_signed_assertion_test.go
Robbie-Microsoft: Added `TestSendCertificateOverMtls_SignedAssertion_RealHandshake` - real loopback TLS, client cert required. Pins callback-runs-once + cert-on-the-wire + token-stays-Bearer together.

Kept `TestSignedAssertionCredentialWithoutMtls` requiring the cert to be ignored: that client asks for neither mTLS option, so returning a cert must not by itself change routing. Clarified its doc and pointed it at the new test.

README.md
Robbie-Microsoft: Removed. `newMSALClient` builds a closed option list and never forwards caller options; zero references to `WithMtlsProofOfPossession` or `BindingCertificate` in the package. The guidance described something callers can't do.

Was in five places, not just this line.

apps/confidential/confidential_mtls_httpclient_ext_test.go
Robbie-Microsoft: Took "limit the documentation to direct MSAL clients" rather than adding Azure SDK acceptance coverage - that needs three azidentity changes first (forward options, request PoP, preserve token type + binding cert in `azcore.AccessToken`).

PR #666, opened 2026-09-10
Add WithConfigurableHTTPClient

### LINKED ISSUE #665: [Bug] ServiceFabric requires *http.Client breaks MIC in azidentity
`v1.9.0` introduced the change in https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/656.

[This](https://github.com/AzureAD/microsoft-authentication-library-for-go/blob/c80dc3709f5bdd3e9875947b3798b648e7f8d9c1/apps/managedidentity/servicefabric.go#L38-L42) requirement is a runtime breaking change and has broken MIC in `azidentity` as we pass a custom client to `WithHTTPClient` that contains our HTTP middleware (retry logic, logging, tracing, etc).

jhendrixMSFT:
This provides a mechanism for callers to supply a wrapped *http.Client when MSAL requires configuring the client or its transport. Updated MI for ServiceFabric to consume a configurable client. Removed nil check for TLSNextProto as it's incorrect and will always fail on a used transport (TLSNextProto is lazily populated by the standard library).

Fixes https://github.com/AzureAD/microsoft-authentication-library-for-go/issues/665


Robbie-Microsoft (commented):
Two general follow-ups for the cross-repository integration:

Could you link the paired azidentity implementation so we can validate that it augments the actual Azure Core transport without bypassing the pipeline and preserves the expected default and custom transport behavior?

As non-blocking regression coverage for removing the old `TLSNextProto != nil` rejection, please consider a test using a previously exercised HTTP/2 transport whose `TLSNextProto` has been lazily populated, confirming that it remains accepted and that Service Fabric certificate pinning is still enforced.

apps/managedidentity/managedidentity.go
```diff
@@ -273,11 +290,28 @@ func New(id ID, options ...ClientOption) (Client, error) {
...
-		serviceFabricClient, serviceFabricURL, err := serviceFabricCertificateVerifiedHTTPClient(client.httpClient)
+		serviceFabricURL, err := serviceFabricEndpoint()
 		if err != nil {
 			return Client{}, err
 		}
-		client.httpClient = serviceFabricClient
+
+		switch tt := client.httpClient.(type) {
+		case ClientConfigurer:
+			err = tt.ConfigureClient(func(c *http.Client) (*http.Client, error) {
```
Robbie-Microsoft, line 319: Please track calls to `augment` and fail unless `ConfigureClient` invokes it synchronously exactly once with a non-nil client and augmentation returns a non-nil client; otherwise a no-op configurer can return success without even creating the client containing the mandatory pin and redirect policy.

apps/managedidentity/managedidentity.go
```diff
@@ -200,14 +200,31 @@ func WithClaims(claims string) AcquireTokenOption {
...
+// This is required when the implementation of HTTPClient isn't a *http.Client and the credential
+// flow requires customizing the client and its underlying transport.
+func WithConfigurableHTTPClient(client ClientConfigurer) ClientOption {
+	return func(c *Client) {
+		c.httpClient = client
+	}
+}
+
+// ClientConfigurer is an ops.HTTPClient that abstracts a configurable HTTP client.
+type ClientConfigurer interface {
```
Robbie-Microsoft, line 238: Please document that `ConfigureClient` must propagate configuration errors, complete configuration before returning, route every subsequent `Do` call through the client returned by `augment` without bypassing its transport or redirect policy, and forward idle-connection cleanup; also clarify whether the configurer may be reused or configured concurrently.

apps/managedidentity/managedidentity.go
```diff
@@ -200,14 +200,31 @@ func WithClaims(claims string) AcquireTokenOption {
 	}
 }
 
-// WithHTTPClient allows for a custom HTTP client to be set. Service Fabric requires a standard
-// *http.Client with a *http.Transport and does not support custom TLS dialing or verification.
+// WithHTTPClient allows for a custom HTTP client to be set. Flows that must configure the
+// transport underlying the client (for example, Service Fabric certificate pinning) require
+// [WithConfigurableHTTPClient] instead, because a plain ops.HTTPClient exposes no way to
```
Robbie-Microsoft: `WithConfigurableHTTPClient` is not technically required because a `ClientConfigurer` passed through `WithHTTPClient` still matches the type switch below; please describe the new option as a statically checked convenience or make the distinction enforceable.

apps/managedidentity/managedidentity.go
```diff
@@ -200,14 +200,31 @@ func WithClaims(claims string) AcquireTokenOption {
...
 func WithHTTPClient(httpClient ops.HTTPClient) ClientOption {
 	return func(c *Client) {
 		c.httpClient = httpClient
 	}
 }
 
+// WithConfigurableHTTPClient sets an HTTP client that installs a transport MSAL configures.
+// This is required when the implementation of HTTPClient isn't a *http.Client and the credential
+// flow requires customizing the client and its underlying transport.
+func WithConfigurableHTTPClient(client ClientConfigurer) ClientOption {
```
Robbie-Microsoft: Please also add `WithConfigurableHTTPClient` to the documented `New` options so callers can discover the new public API through Go documentation.

apps/managedidentity/servicefabric.go
```diff
@@ -22,49 +22,35 @@ import (
...
-		return nil, "", err
+		return nil, err
 	}
 
-	callerClient, ok := httpClient.(*http.Client)
-	if !ok {
-		return nil, "", errors.New("Service Fabric managed identity requires a standard *http.Client")
-	}
-	derivedClient := *callerClient
+	derivedClient := *httpClient
```
Robbie-Microsoft, line 34: Please reject a nil `httpClient` before dereferencing it, since a `ClientConfigurer` can pass nil to `augment` and currently cause a panic instead of returning a configuration error.

apps/managedidentity/servicefabric_test.go
```diff
@@ -35,6 +35,102 @@ func (f serviceFabricRoundTripperFunc) RoundTrip(request *http.Request) (*http.R
...
+	base           *http.Client
+	configured     *http.Client
+	configureCalls int
+}
+
+func (c *serviceFabricConfigurableClient) Do(request *http.Request) (*http.Response, error) {
+	if c.configured == nil {
+		return nil, errors.New("transport was not configured")
+	}
+	return c.configured.Do(request)
```
Robbie-Microsoft, line 52: This fixture delegates directly to the configured client, so it proves wrapper retention but not middleware preservation; please add observable wrapper middleware, such as injecting a header, and verify that the server receives it through the configured transport.

apps/managedidentity/servicefabric_test.go
```diff
@@ -35,6 +35,102 @@ func (f serviceFabricRoundTripperFunc) RoundTrip(request *http.Request) (*http.R
...
+	}
+	// The configurer, not the derived client, must remain in use so its middleware is preserved.
+	if client.httpClient != configurable {
+		t.Fatalf("expected the configurable client to remain in use, got %T", client.httpClient)
+	}
+	derivedTransport, ok := configurable.configured.Transport.(*http.Transport)
+	if !ok {
+		t.Fatalf("expected derived *http.Transport, got %T", configurable.configured.Transport)
+	}
+	if derivedTransport.TLSClientConfig == nil || derivedTransport.TLSClientConfig.VerifyConnection == nil {
```
Robbie-Microsoft, line 113: Please add a mismatched-thumbprint test through `WithConfigurableHTTPClient` and verify that the destination handler is never invoked, since the existing wrong-pin coverage exercises only the direct `*http.Client` path.

apps/managedidentity/servicefabric_test.go
```diff
@@ -35,6 +35,102 @@ func (f serviceFabricRoundTripperFunc) RoundTrip(request *http.Request) (*http.R
...
+		t.Fatal("expected Service Fabric certificate pinning on the derived transport")
+	}
+	if configurable.configured.CheckRedirect == nil {
```
Robbie-Microsoft, line 116: Please exercise a redirect through `WithConfigurableHTTPClient` and verify that the redirect target receives neither a request nor the `Secret` header, because checking only that `CheckRedirect` is non-nil does not prove the wrapper routes through the augmented client.

apps/managedidentity/servicefabric_test.go
```diff
@@ -35,6 +35,102 @@ func (f serviceFabricRoundTripperFunc) RoundTrip(request *http.Request) (*http.R
 	return f(request)
 }
 
+// serviceFabricConfigurableClient is a TransportConfigurer that supplies a base client to augment
```
Robbie-Microsoft: Please update the stale `TransportConfigurer` name here and the two `ConfigureTransport` failure messages below to `ClientConfigurer` and `ConfigureClient`.

apps/managedidentity/managedidentity.go
```diff
@@ -200,14 +200,31 @@ func WithClaims(claims string) AcquireTokenOption {
...
+	return func(c *Client) {
+		c.httpClient = client
+	}
+}
+
+// ClientConfigurer is an ops.HTTPClient that abstracts a configurable HTTP client.
+type ClientConfigurer interface {
+	ops.HTTPClient
+	ConfigureClient(augment func(*http.Client) (*http.Client, error)) error
```
bgavrilMS, line 242: @jhendrixMSFT - not sure if this callback will be sufficient. For MTLS POP via Managed Identity, MSAL will be creating **short lived certificates**. In MSAL.NET we went with an interface like: 

```csharp
HttpClient GetHttpClient(X509Certificate2 cert)
```

Doesn't MSAL GO **need** to give Azure SDK the tls.certificate when asking for an http client?

apps/managedidentity/managedidentity.go
jhendrixMSFT: fixed

apps/managedidentity/managedidentity.go
jhendrixMSFT: fixed

apps/managedidentity/managedidentity.go
jhendrixMSFT: That's a good point.  I've removed `WithConfigurableHTTPClient` and documented the conditions for a `ClientConfigurer`.

apps/managedidentity/managedidentity.go
jhendrixMSFT: This has been removed.

apps/managedidentity/servicefabric.go
jhendrixMSFT: fixed

apps/managedidentity/servicefabric_test.go
jhendrixMSFT: fixed

apps/managedidentity/servicefabric_test.go
jhendrixMSFT: fixed

apps/managedidentity/servicefabric_test.go
jhendrixMSFT: fixed

apps/managedidentity/servicefabric_test.go
jhendrixMSFT: fixed

apps/managedidentity/managedidentity.go
jhendrixMSFT: With the `ClientConfigurer`, MSAL has complete control over the `*http.Client` and its transport.  So, MSAL should be able to install its own callback on the transport's `TLSClientConfig` that sets the cert as required.  Something like this.
```go
tlsConfig.GetClientCertificate = func(*tls.CertificateRequestInfo) (*tls.Certificate, error) {
    return certStore.Get() // provides latest cert
}
```
Where `certStore` manages the short lived certs.

4gust:
@jhendrixMSFT Thank you for the PR, 
I have a concern about how configuration failures are handled.

```go
err = tt.ConfigureClient(func(c *http.Client) (*http.Client, error) {
    augmentCalls++
    return serviceFabricCertificateVerifiedHTTPClient(c)
})
if err == nil && augmentCalls != 1 {
    return Client{}, fmt.Errorf(
        "ConfigureClient must call augment exactly once to install the Service Fabric client, got %d calls",
        augmentCalls,
    )
}
```

The `augmentCalls` check is useful because it rejects a `ConfigureClient` implementation that returns before calling `augment`.

However, it does not handle a callback failure that is ignored by the client wrapper. For example:

```go
func (c *wrapper) ConfigureClient(
    augment func(*http.Client) (*http.Client, error),
) error {
    _, _ = augment(nil) // configuration fails, but the error is ignored
    return nil          // reports success to MSAL
}
```

In this case, `augmentCalls` is `1` and `err` is `nil`, so `New(...)` succeeds. However, the client was not configured with the Service Fabric certificate validation and redirect protection.

Could we make this fail closed? Ideally, `New(...)` should return an error whenever `augment` fails, even when `ConfigureClient` ignores that error. This prevents returning a client that appears ready but is not correctly configured.

If `New(...)` is intentionally allowed to succeed after configuration fails, then the configuration failure should be saved and returned by every later `AcquireToken(...)` call, without sending a request. It should never be silently ignored.

Please add a test where `ConfigureClient` calls `augment`, ignores its error, and returns `nil`. The test should verify that either:

1. `New(...)` returns the configuration error; or
2. If initialization is intentionally deferred, every `AcquireToken(...)` call returns that error and does not send a request.

Also, `ClientConfigurer` is described as a general interface, but it is currently called only for Service Fabric. Please document that scope clearly, or define and test its behavior for the other managed identity sources.

jhendrixMSFT:
@4gust good catch, this has been updated to fail close on augmentation error.

apps/managedidentity/servicefabric.go
```diff
@@ -22,49 +22,38 @@ import (
...
 	}
 	if callerTransport.TLSClientConfig != nil &&
 		(callerTransport.TLSClientConfig.VerifyPeerCertificate != nil || callerTransport.TLSClientConfig.VerifyConnection != nil) {
-		return nil, "", errors.New("Service Fabric managed identity does not support custom TLS verification")
-	}
-	if callerTransport.TLSNextProto != nil {
-		return nil, "", errors.New("Service Fabric managed identity does not support custom TLS protocol handlers")
+		return nil, errors.New("Service Fabric managed identity does not support custom TLS verification")
 	}
 	derivedTransport := callerTransport.Clone()
```
Robbie-Microsoft, line 58: Go's `Transport.Clone` omits lazily populated standard HTTP/2 state but copies a caller-supplied `TLSNextProto` map, so could we reject a non-nil `derivedTransport.TLSNextProto` after cloning to preserve normal reused HTTP/2 transports while continuing to fail closed for custom alternate-protocol handlers?

apps/managedidentity/servicefabric_test.go
```diff
@@ -35,6 +35,295 @@ func (f serviceFabricRoundTripperFunc) RoundTrip(request *http.Request) (*http.R
...
+	server := httptest.NewTLSServer(http.HandlerFunc(func(response http.ResponseWriter, _ *http.Request) {
+		response.Header().Set("Content-Type", "application/json")
+		_, _ = response.Write(responseBody)
+	}))
+	defer server.Close()
+
+	setServiceFabricEnvironment(t, server.URL, serviceFabricServerThumbprint(server))
+	resetServiceFabricCache(t)
+
+	callerTransport := server.Client().Transport.(*http.Transport).Clone()
```
Robbie-Microsoft, line 300: This manually assigned map represents caller configuration rather than lazily populated standard HTTP/2 state, and its handler is never exercised; please change this case to expect rejection, then initialize a real HTTP/2 transport with a request, reuse it for Service Fabric, and verify matching-pin success and wrong-pin failure.

jhendrixMSFT: Added a few more tests for this.

apps/managedidentity/servicefabric.go
jhendrixMSFT: We can't reliably distinguish a legitimately HTTP/2-configured transport from a custom alternate-protocol handler after cloning — both surface as a non-nil `TLSNextProto` map of opaque funcs. e.g. we have a pre-configured transport in `azcore` that `azidentity` uses (see [here](https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/transport_default_http_client.go#L36-L41)).  It's also not a pinning-bypass vector: the transport completes the pinned TLS handshake before dispatching to the ALPN handler, and custom TLS dialing/verification are already rejected.

Robbie-Microsoft: You’re right that rejecting non-nil `TLSNextProto` would break Azure Core. One path I don’t think the tests cover: `x/net/http2.ConfigureTransports` installs an `h2` callback that retains the original connection pool, and `Transport.Clone` copies that callback. If the pool already has a connection for the authority, it can discard the newly pinned connection and use the old one.

Could we test this by warming an HTTP/2 connection to server A, routing later dials for the same authority to server B whose certificate matches the pin, and checking that the Service Fabric request goes to B rather than A? The wrong-pin test exits before the callback, and the current reuse test doesn’t use the `x/net/http2` callback.

jhendrixMSFT: Yep, good point, and the test confirms it.  In this case we can `nil` out `TLSNextProto` after cloning so the standard library will lazily create a fresh pool.

4gust (approved):

Robbie-Microsoft (approved):

jhendrixMSFT:
@Robbie-Microsoft CI appears to be stuck.  Is there something I need to do in order to trigger the two remaining jobs?

Robbie-Microsoft:
> Robbie-Microsoft CI appears to be stuck. Is there something I need to do in order to trigger the two remaining jobs?

CI won’t run because this PR’s source branch is in your fork rather than the AzureAD repository, and the pipeline excludes fork PRs. Are you able to push the branch to `AzureAD/microsoft-authentication-library-for-go` and open a replacement PR?

jhendrixMSFT:
Superseded by https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/669

PR #667, opened 2026-09-14
Resolve Go lint backlog and enforce formatting in CI

Robbie-Microsoft:
## Summary

This PR resolves the lint backlog created by the `golangci-lint` v2 upgrade and turns the temporary lint configuration into an enforced baseline.

**103 findings across 34 files are reduced to zero.** `errcheck` is enabled, the broad `govet`, `staticcheck`, and `gosec` suppressions are removed, and formatting is enforced in CI.

No public API is added, removed, or re-signatured.

---

## 1. Lint cleanup

| Linter | Findings resolved | Main changes |
|---|---:|---|
| `govet` | 29 | Replace the deprecated `reflect.Ptr` alias with `reflect.Pointer` |
| `staticcheck` | 25 | Simplify conditions, switches, string replacement, formatted writes, and error messages |
| `errcheck` | 27 | Handle file, response, cleanup, encoding, and test-server write errors |
| `gosec` | 22 | Fix actionable findings and replace broad exclusions with narrow justifications |

`errcheck` is now enforced. The temporary `govet` and `staticcheck` exclusions are removed entirely.

## 2. Resource and error handling

File and test-resource errors are now surfaced instead of silently discarded. Test HTTP handlers check response writes and JSON encoding, temporary file cleanup reports failures, and HTTP response bodies are consistently closed.

Region autodetection also closes the first unsuccessful response before retrying and closes the successful retry response after reading it. Errors from closing an already consumed or discarded HTTP response remain explicitly non-fatal so they cannot reject an otherwise valid authentication response.

## 3. Security linting

The previous configuration suppressed `G704`, `G304`, and `G104` repository-wide and disabled `gosec` for all tests. This PR replaces those broad rules with:

- line-level explanations for endpoints supplied by Azure managed identity hosts
- a Service Fabric endpoint explanation tied to HTTPS validation and certificate thumbprint pinning
- explicit handling of test fixture paths, Service Fabric SHA-1 thumbprints, and non-security benchmark randomness
- a Go 1.18-compatible loop-variable aliasing fix
- tighter permissions for generated test directories
- one narrow test-only exclusion for `G101` and `G117`, covering fake credential fixtures and marshaled fake access tokens

## 4. Formatting enforcement

The previously commented formatting check is replaced with an active CI step on the Go 1.22 matrix leg. It runs `gofmt -l -s .`, prints every file requiring formatting, and fails when the list is non-empty.

`.gitattributes` pins Go source files to LF so Windows line-ending conversion does not make otherwise formatted files appear dirty to `gofmt`.

## 5. Behavior and compatibility

There are no public API changes. Most edits are mechanical. Observable changes are limited to Go-standard lowercase error formatting, more complete test cleanup diagnostics, and correct response cleanup during region autodetection.

Managed identity endpoint selection and supported environments are unchanged.

## Tests and validation

- `go build ./apps/...`
- `golangci-lint run` — **0 issues**
- `gofmt -l -s .` — **0 files**
- `go test -short ./apps/cache/... ./apps/confidential/... ./apps/public/... ./apps/internal/... ./apps/managedidentity/...`

The race-enabled suite remains in Linux CI because Go does not support `-race` on Windows ARM64.


apps/internal/oauth/ops/authority/authority.go
```diff
@@ -732,17 +732,26 @@ func detectRegion(ctx context.Context) string {
...
-	if err != nil || resp.StatusCode != http.StatusOK {
-		resp, err = client.Do(req)
-		if err != nil || resp.StatusCode != http.StatusOK {
-			return ""
-		}
+	resp, err = client.Do(req)
+	if err != nil {
+		return ""
 	}
 	response, err := io.ReadAll(resp.Body)
```
4gust, line 759: check whether the retry succeeded before reading its response, keep the clean up fix, and add a test for this situation. 
the success for http call don't necessary mean that the response status is 200

Robbie-Microsoft: Fixed in a8f4498. The retry now checks for 200 before reading, closes non-200 bodies, and has regression coverage for both attempts.

bgavrilMS (approved):

4gust (approved):

PR #668, opened 2026-09-14
Add privacy-safe OpenTelemetry authentication metrics

Robbie-Microsoft:
## Summary

Adds privacy-safe authentication metrics to MSAL Go using the current MSAL.NET OpenTelemetry V2 schema.

The implementation instruments confidential, public, and managed-identity token acquisition while preserving the root module's Go 1.18 compatibility. OpenTelemetry support lives in an optional Go 1.25 nested module, so existing consumers acquire neither an OTel dependency nor a behavior change.

## Architecture

```text
MSAL token acquisition
        ↓
privacy-safe AuthenticationEvent
        ↓
MetricsProvider
        ↓
optional OpenTelemetry adapter
```

The root module remains Go 1.18-compatible and OTel-independent. Without a configured provider, telemetry is a no-op. The optional `apps/integrations/opentelemetry` module requires Go 1.25 and OTel v1.46; applications retain control of their SDK, reader, and exporter.

## Authentication event model

Each caller-facing token acquisition produces one privacy-safe
`AuthenticationEvent`. Cache operations, HTTP calls, retries, and region
detection contribute bounded measurements to the event. MSAL completes it at
most once when the acquisition finishes.

The event is an MSAL Go contract—not an OpenTelemetry span event, log record,
or independently exported event stream. The configured provider receives the
completed event, and the optional OpenTelemetry adapter translates it into the
applicable counters and histograms. For example, one failed acquisition can
increment `MsalFailure` and record both total and HTTP duration.

```text
All supported Go versions
  MSAL Go core module
    → AuthenticationEvent
    → MetricsProvider
        ├─ application-defined provider
        └─ optional OpenTelemetry adapter
             └─ requires Go 1.25+
```

This design:

- keeps OpenTelemetry out of the MSAL Go core module while preserving its Go
  1.18 minimum version
- establishes one bounded privacy boundary for authentication telemetry
- aggregates nested cache and HTTP work into one caller-facing acquisition
- prevents duplicate metrics from internal operations
- permits applications on any supported Go version to implement
  `MetricsProvider` directly
- provides the OpenTelemetry adapter to applications using Go 1.25 or later

## Metrics

| Instrument | Type | Unit |
| --- | --- | --- |
| `MsalSuccess` | Counter | |
| `MsalFailure` | Counter | |
| `MsalTotalDurationV2.1A` | Histogram | `ms` |
| `MsalDurationInL1CacheInUs.1B` | Histogram | `us` |
| `MsalDurationInHttpV2.1A` | Histogram | `ms` |
| `MsalRemainingTokenLifetime.1A` | Histogram | `s` |

Names, API IDs, attributes, and units follow MSAL.NET. `.1A` and `.1B` are retained because they select histogram bucket configurations in Microsoft's MISE Collector. `CanonicalTagsByMetric` exposes the exact per-instrument attribute sets.

## Example output

A failed client-credential acquisition emits data equivalent to:

```text
MsalFailure = 1
  MsalVersion: "1.10.0"
  Platform: "linux"
  ErrorCode: "invalid_client"
  ApiId: 1004
  CallerSdkId: ""
  CacheRefreshReason: 2
  TokenType: 1
  RawStsErrorCode: "7000215"

MsalTotalDurationV2.1A = 125 ms
  MsalVersionPlatform: "1.10.0,linux"
  ApiId: 1004
  TokenSource: ""
  CacheLevel: ""
  CacheRefreshReason: 2
  TokenType: 1
  ErrorCode: "invalid_client"
  Succeeded: false

MsalDurationInHttpV2.1A = 118 ms
  MsalVersionPlatform: "1.10.0,linux"
  ApiId: 1004
  TokenType: 1
  HttpStatusCode: 401
```

OpenTelemetry exporters wrap these measurements in their own resource and scope envelopes. The canonical data contains no client or tenant ID, account, scope, authority or URL, correlation ID, token, secret, response body, or error description. Trace and span IDs are also removed from metric exemplars.

## Public API

Adds `WithMetricsProvider` to confidential, public, and managed-identity clients. The Go 1.18-safe `telemetry.MetricsProvider` receives one typed `AuthenticationEvent` per caller-facing acquisition and can support OTel or other metric backends.

Coverage includes identity-provider, cache, refresh, retry, cancellation, deadline, and service-failure paths. Internal operations do not produce duplicate acquisition events. HTTP duration accumulates across contributing requests and retries.

## Privacy and cardinality

Events exclude client and tenant IDs, accounts, scopes, resources, claims, authorities, URLs, correlation IDs, tokens, secrets, cache contents, raw errors, and service descriptions.

OAuth errors use an explicit allowlist; unknown values become `service_error`. `RawStsErrorCode` contains only the first numeric STS code and remains present with an empty value when unavailable. Token types use a bounded MSAL taxonomy.

## Deliberate exclusions

- Legacy V1 total and HTTP histograms
- L2 duration, because `cache.ExportReplace` cannot prove a hit's source
- Extension duration, because Go has no equivalent operation
- Traces, spans, and caller-defined metric attributes

These are omitted instead of emitting misleading or unbounded data.

## Developer experience

Includes schema and privacy tests, concurrent acquisition and context-cancellation coverage, dedicated Go 1.25 adapter CI with the race detector, and a console-exporter sample demonstrating identity-provider, cache, and sanitized failure metrics.


apps/internal/base/base.go
```diff
@@ -536,11 +598,13 @@ func (b Client) AuthResultFromToken(ctx context.Context, authParams authority.Au
 		defer b.cacheAccessorMu.Unlock()
 		err := b.cacheAccessor.Replace(ctx, m, cache.ReplaceHints{PartitionKey: key})
 		if err != nil {
+			internaltelemetry.ObserveErrorCode(ctx, "cache_error")
 			return AuthResult{}, err
 		}
 	}
 	account, err := m.Write(authParams, token)
 	if err != nil {
+		internaltelemetry.ObserveErrorCode(ctx, "cache_error")
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed in `23b868f`: export failures now record `cache_error` before returning.

apps/internal/telemetry/acquisition.go
```diff
@@ -0,0 +1,242 @@
...
+	if a == nil {
+		return
+	}
+	tokenType := publictelemetry.TokenTypeExtension
+	if strings.EqualFold(accessTokenType, "Bearer") {
+		tokenType = publictelemetry.TokenTypeBearer
+	} else {
+		// AuthenticationScheme is currently exposed only for Azure Arc PoP.
+		tokenType = publictelemetry.TokenTypePoP
+	}
```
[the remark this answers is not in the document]
Robbie-Microsoft: No change: Go currently exposes custom `AuthenticationScheme` values only for Azure Arc PoP, so non-Bearer values are intentionally classified as bounded PoP rather than extension.

apps/internal/telemetry/acquisition.go
```diff
@@ -0,0 +1,242 @@
...
+	a.event.Succeeded = succeeded
+	a.event.TokenSource = tokenSource
+	a.event.ExpiresOn = expiresOn
+	a.event.TotalDuration = time.Since(a.started)
+	if errorCode == "context_canceled" || errorCode == "context_deadline_exceeded" {
+		a.event.ErrorCode = errorCode
+		a.event.RawSTSErrorCode = ""
+	} else if a.event.ErrorCode == "" {
+		a.event.ErrorCode = errorCode
+	}
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed in `23b868f`: successful completion now clears transient polling error and STS-code state, with regression coverage.

neha-bhargava:
What drives adding this support in Go? Are we planning to collect the data? If yes where are we planning to?

Robbie-Microsoft:
> What drives adding this support in Go? Are we planning to collect the data? If yes where are we planning to?

Following up on our Teams discussion: this started as a Hackathon project to bring MSAL Go closer to MSAL.NET's telemetry capabilities, rather than in response to a current customer request. We don't currently have a Microsoft-owned collection destination planned for the Go data.

MSAL.NET consumes these metrics through MISE and private first-party collector integration, which isn't something we can expose or directly reuse here. This PR instead provides opt-in public instrumentation: application developers configure their own OpenTelemetry SDK, exporter, and destination. MSAL defines and emits the privacy-safe measurements but neither collects the data itself nor selects where it is sent.

A Microsoft-owned dashboard or first-party consumer for the Go metrics would require separate follow-up work.

.github/workflows/go.yml
```diff
@@ -50,3 +52,24 @@ jobs:
       #     oboConfidentialClientSecret: ${{ secrets.OBO_CONFIDENTIAL_APP_CLIENT_SECRET }}
       #     oboPublicClientId: ${{ secrets.OBO_PUBLIC_APP_CLIENT_ID }}
       #     CI: ${{secrets.ENABLECI}}
+
+  opentelemetry:
```
bgavrilMS: The actual CI is on ADO, not on GH. Tests run there.

apps/confidential/confidential.go
```diff
@@ -290,6 +293,13 @@ func WithHTTPClient(httpClient ops.HTTPClient) Option {
 	}
 }
 
+// WithMetricsProvider configures a privacy-safe authentication metrics destination.
+func WithMetricsProvider(provider telemetry.MetricsProvider) Option {
```
neha-bhargava, line 297: I would like to understand this more. What does the provider do. Is it like the collector?

4gust (approved):

apps/confidential/confidential.go
Robbie-Microsoft: It's not the collector. It's the object MSAL calls when a token acquisition finishes. MSAL gives it one privacy-safe event, and our OTel implementation turns that event into counters and histograms. The app's exporter can then send those metrics to a collector or whatever backend the app chooses.

Would `MetricsRecorder` be clearer than `MetricsProvider`? I can rename it.

.github/workflows/go.yml
Robbie-Microsoft: Done. I moved the OpenTelemetry adapter coverage into the Azure DevOps pipeline as a separate Go 1.25 stage and removed the GitHub Actions job.

apps/confidential/confidential.go
neha-bhargava: Yes renaming sounds like a better idea

README.md
```diff
@@ -128,6 +128,20 @@ Acquiring tokens with MSAL Go follows this general pattern. There might be some
     accessToken := result.AccessToken
     ```
 
+## OpenTelemetry metrics
+
+MSAL Go can report privacy-safe token acquisition metrics through the
+[`telemetry.MetricsProvider`](https://pkg.go.dev/github.com/AzureAD/microsoft-authentication-library-for-go/apps/telemetry)
+interface. Configure a provider with `confidential.WithMetricsProvider`,
```
neha-bhargava, line 135: Add a sample basic configuration here

apps/confidential/confidential_test.go
```diff
@@ -171,6 +189,47 @@ func TestAcquireTokenByCredential(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	recorder := &metricsRecorder{}
+	client, err := fakeClient(accesstokens.TokenResponse{
+		AccessToken:   token,
+		ExpiresOn:     time.Now().Add(time.Hour),
+		GrantedScopes: accesstokens.Scopes{Slice: tokenScope},
+		TokenType:     authority.AccessTokenTypeBearer,
+	}, cred, fakeAuthority, WithMetricsProvider(recorder))
```
neha-bhargava, line 203: There is a way to end to end test it. I will share it offline

neha-bhargava (approved):
Approve with suggestions

apps/confidential/confidential.go
Robbie-Microsoft: Done. Renamed it to `MetricsRecorder` throughout the API, adapter, examples, and docs.

README.md
Robbie-Microsoft: Done. Added a basic SDK, stdout exporter, recorder, and client configuration example here.

